PR #14, opened 2013-12-02
Add note to ST_Equals that the function does not support geometry collections

nmandery:
I just had a situation where ST_Equals failed to compare GeometryCollections. Looking add the source this is normal behavior, but there is no mention of it in the docs of ST_Equals.

This pull request adds a section to the documentation.

### DIFF
--- a/doc/reference_measure.xml
+++ b/doc/reference_measure.xml
@@ -2501,7 +2501,11 @@ SELECT s.gid, s.school_name
 			points are the same).</para>
 
 		<important>
-		  <para>This function will return false if either geometry is invalid even if they are binary equal.</para>
+	        <para>This function will return false if either geometry is invalid even if they are binary equal.</para>
+		</important>
+
+		<important>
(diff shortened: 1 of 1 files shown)


mwtoews:
The problem is deeper than just ST_Equals. See http://lists.osgeo.org/pipermail/postgis-devel/2015-April/024827.html

If you have a homogeneous geometry collection, e.g. from ST_Polygonize, then use [ST_CollectionExtract(geom, 3)](http://postgis.net/docs/ST_CollectionExtract.html) to get a MultiPolygon, which works with ST_Equals and other relate functions.

strk:
@nmandery are you planning to provide better wording for the manual about this ?
It's basically that geometrycollection with proper topology work while those with ambiguous one would not. A workaround is to pass them through ST_UnaryUnion beforehand.
See https://trac.osgeo.org/geos/ticket/725

pramsey:
OK at https://trac.osgeo.org/postgis/changeset/14693/

PR #16, opened 2013-12-05
Add config flag into docs for sfcgal

smathermather:


robe2:
This looks like it was already addressed a while ago.  I see --with-sfcgal listed and that more importantly it's the sfcgal-config and not the folder.

PR #28, opened 2014-10-20
Fixed county level file download match pattern when stateid = 20

bibanul:
2014 TIGER lines file naming format is tl_YYYY_state_fips*_tablename.zip. The current County level file match download pulls ALL zipfiles for all states when statecode is 20 as it matches /*_ to tl_2014_* when it should in fact match only tl_2014_20*.


robe2:
Confirmed I already fixed for 2014 in 2.2. branch

PR #32, opened 2015-02-10
Add test output file to .gitignore

maurizi:


strk:
The newly ignored file is never generated on my system.
Also, raster/test/regress/loader/*.opts files exist but are all
source (not-generated) files.

maurizi:
@strk I've got this figured this out.  The `BasicOutDB` test was being skipped.  There is a bug with `run_test.pl`, where it will run the `*-pre.pl` files, but not the `*-post.pl` files when a test is skipped.

I'll submit a fix for that shortly.

PR #33, opened 2015-02-11
Fix issue with skipped tests not being cleaned up properly

maurizi:
This was what was causing the untracked file I noticed in #32.

### DIFF
--- a/regress/run_test.pl
+++ b/regress/run_test.pl
@@ -331,7 +331,8 @@ foreach $TEST (@ARGV)
 	{
 		print " skipped (can't read any ${TEST}.{sql,dbf,tif,dmp})\n";
 		$SKIP++;
-		next;
+		# Even though we skipped this test, we should still run the cleanup
+		# scripts
 	}
 
 	if ( -r "${TEST}-post.sql" )


strk:
but, why was the test skipped ? Reading the code it looks like the skip is due to a missing file. Why is there a "pre" without a corresponding file ? There might be another bug hiding here

dustymugs:
Having the *-post.pl run regardless of the test succeeding makes sense. The test itself isn't being skipped (as far as the latest trunk indicates). The *-pre generates the appropriate opt file and the *-post destroys it.

maurizi:
@strk I think you're right, there is some underlying bug if the `BasicOutDb` test is being skipped.  Perhaps it is something specific to my environment that is causing the test to be skipped.

That said, it seems like a good practice to always run the *-post.pl if we run the corresponding *-pre.pl, regardless of whether the actual test is run or not.

strk:
if we figure out what triggers the skip, it would probably make sense to skip the test completely, not running the *-pre.pl either.

pramsey:
Thanks, applied, https://trac.osgeo.org/postgis/changeset/14679/

PR #36, opened 2015-04-11
Enable ST_Transform to use PROJ.4 text

mwtoews:
See [trac#2991](http://trac.osgeo.org/postgis/ticket/2991).

With this PR, the signatures for `ST_Transform` are:
* `geometry ST_Transform(geometry geom, integer to_srid);` (this is the existing one)
* `geometry ST_Transform(geometry geom, text to_proj);`
* `geometry ST_Transform(geometry geom, text from_proj, text to_proj);`
* `geometry ST_Transform(geometry geom, text from_proj, integer to_srid);`

I'm not sure if more examples are needed. I was considering this one, which draws a buffer 2,200 km from Iceland using a standard `geography` technique, and using a dynamic azimuthal equidistant projection:
```
WITH data AS (
  SELECT 'SRID=4326;POINT(18 65)'::geometry geom, 2200000.0 dist
)
SELECT 'using geography' AS description,
  ST_Buffer(geom::geography, dist)::geometry AS geom
FROM data
UNION SELECT 'using a dynamic azimuthal equidistant projection',
  ST_Transform(ST_Buffer(ST_MakePoint(0, 0), dist, 32),
    concat('+proj=aeqd +lat_0=', ST_Y(geom), ' +lon_0=', ST_X(geom),
           ' +x_0=0 +y_0=0 +ellps=WGS84 +no_defs '), 4326)
FROM data;
```
Shown on EPSG:3857, where the better looking green polygon is the aeqd result, and the red is the geography buffer.
map
however, I don't think an example that highlights known problems with _ST_BestSRID would be effective. So I went with an [example by Paul](http://boundlessgeo.com/2012/02/flattening-the-peel/) instead, but without the temporary SRID=55000.

Also, I've removed `postgis_transform_geometry` (`transform_geom`), since it was never documented (see [trac#2207](http://trac.osgeo.org/postgis/ticket/2207)), and the functionality would be replaced by these functions. This was [on the chopping block before](http://lists.osgeo.org/pipermail/postgis-devel/2012-February/018150.html).


strk:
How's this different from postgis_transform_geometry ?

mwtoews:
The old function translated from/to using PROJ.4 text, and set a user-specified SRID to the geometry result. These new functions allow the to/from projection to be defined by an SRID, in combination with a PROJ.4. (There are four combinations using `from`/`to` and `srid`/`proj`, thus four `ST_Transform` functions, where the equivalent of `from_srid` is a geometry with a valid SRID set).

With the new set of functions, the undocumented `postgis_transform_geometry` would effectively be:
```
CREATE OR REPLACE FUNCTION postgis_transform_geometry(geometry, text, text, integer)
  RETURNS geometry AS
'SELECT ST_SetSRID(ST_Transform($1, $2, $3), $4);' LANGUAGE sql IMMUTABLE STRICT;
```
One aspect of `postgis_transform_geometry` that I disagree with is that it assigns a seemingly arbitrary SRID to the output geometry.  A geometry transformed to an arbitrary PROJ.4 text should have an undefined SRID (0).

strk:
The "arbitrary" SRID is actually specified by the user, right ?

By comparison, your ST_Transform(geometry, text, text) leaves the SRID
untouched, is that correct ?

Is there anything that your function allows that isn't currently possible
with postgis_transform_geometry ?

mwtoews:
The functions with `to_proj` return a geometry with undefined SRID. `postgis_transform_geometry` cannot use or return geometry with an undefined SRID, e.g. shows "ERROR:  tranform: destination SRID = 0".

The equivalent functions could be re-assembled using SQL functions:
```
CREATE OR REPLACE FUNCTION ST_Transform(geom geometry, to_proj text)
  RETURNS geometry AS
'SELECT ST_SetSRID(postgis_transform_geometry($1, proj4text, $2, 99), 0)
FROM spatial_ref_sys WHERE srid=ST_SRID($1);'
  LANGUAGE sql IMMUTABLE STRICT;

CREATE OR REPLACE FUNCTION ST_Transform(geom geometry, from_proj text, to_proj text)
  RETURNS geometry AS
'SELECT ST_SetSRID(postgis_transform_geometry(ST_SetSRID($1, 99), $2, $3, 99), 0)'
  LANGUAGE sql IMMUTABLE STRICT;

CREATE OR REPLACE FUNCTION ST_Transform(geom geometry, from_proj text, to_srid integer)
  RETURNS geometry AS
'SELECT postgis_transform_geometry(ST_SetSRID($1, 99), $2, proj4text, $3)
FROM spatial_ref_sys WHERE srid=$3;'
  LANGUAGE sql IMMUTABLE STRICT;
```
where there are the arbitrary SRID=99 is set to avoid the "tranform" [sic] error from `postgis_transform_geometry`, then unset where it should be undefined.

strk:
To me it sounds like a bug that "postgis_transform_geometry" does not
accept an unknown SRID as the destination.

A quick look at the code also tells me that the function does not accept
in input a geometry with unknown srid which is also meaningless to me.

Would fixing those bugs make you all set-up for your needs ?

mwtoews:
But `postgis_transform_geometry` still has no documentation, regression tests, and strangely requires an SRID for the resulting geometry. This PR fixes these, and additionally shows error messages for invalid `to`/`from` and `srid`/`proj` parameters, which the sql function workarounds in my previous comment cannot always do.

dbaston:
@mwtoews and @strk , I started a branch that pulls in Mike's SQL functions (above), test cases, and documentation edits.  Given that we can't just delete `postgis_transform_geometry`, I think this gets most of the usefulness while adding less code.  Any objections to pushing this in?

https://github.com/dbaston/postgis/tree/trac-2991b

strk:
I'm ok to add these but please make it clear in the documentation
what the SRID of the output geometry is when "to_proj" is given
( I don't think it's evidenced right now )

dbaston:
Committed this functionality at r14698

pramsey:
Committed, so, closing.

PR #37, opened 2015-05-04
Add installcheck step to travis build

strk:
testing travis recipe to run online tests


strk:
works, pushed as r13481

PR #38, opened 2015-05-17
fixed confused help message

ivolf:
* configure.ac: --with-pgconfig parameter vs. --with-pg_config help message of configure (fixed confused help message to --with-pgconfig).


strk:
Thanks, committed as r13516 in trunk and r13517 in 2.1 branch

PR #39, opened 2015-06-17
Add ST_ApproximateMedialAxis (#3169)

strk:
Requires SFCGAL (any version).
Uses SFCGAL specific API if available (1.2+),
post-processes StraightSkeleton output otherwise.

Includes regression test and documentation.


strk:
This is just hear to gather some comments.
\cc @vmora @robe2 

I'm thinking maybe it should just be an additional parameter to ST_StraightSkeleton, instead.

robe2:
I like not having yet another function.  So yes nd extra arg would be better.

strk:
Uhm, on the other hand SFCGAL 1.2.0 will have an ``sfcgal_geometry_approximate_medial_axis`` function which we would be calling. Maybe it'll be more confusion to bury a different call into an existing one.

Also, if we want a plpgsql implementation of "approximate medial axis" when SFGAL built is < 1.2 it still takes a different signature. Well I guess I'll go with what I have now, less problems overall.

strk:
pushed in commit 533361f15d4393fe8bd535cfebb4be96f65e25a1 (I guess) -- or r13678

PR #40, opened 2015-07-01
resolve ST_Equals TODO

dbaston:


strk:
Beware of rounded boxes.
Basically a bounding box may be slightly larger than needed when
it is created by reading the version cached in a serialized geometry.

This is because serialization forces use of floats by ensuring the
float version of the box fully covers the double version.
When reading things back the "fitting" box info is discarded.

Depending on where the values passed to ST_Equals originate
(read from disk or output from another function) the exact equality
might fail.

It would be fun to try at reproducing this scenario in a regression
test to proof what I'm saying and ensure it's not broken in the future.

dbaston:
If the inputs to `ST_Equals` are both of type `GSERIALIZED`, won't both bounding boxes have forced to float somewhere upstream?

strk:
There are cases in which the bounding box is not included in the GSERIALIZED.
In those cases, it is computed with full precision when needed.

Your patch seems to be actually only using gbox_same_2d when both
inputs do have the cache, so should be safe.

Only I thought to mention the possibly hiding problem
(may be worth a comment in the code)

dbaston:
I was also thinking about a `gbox_same_float` using the same rounding logic as `gserialized_from_gbox`, or maybe a more loose `gbox_same_within_tol`.  Do you think one of those would be preferable?

strk:
I like the "gbox_same_float" idea (stricter)

--strk;

pramsey:
Applied to SVN at r13788, thanks

PR #41, opened 2015-07-01
Fix missed cunit asserts

dbaston:
Looked like maybe the 13 should have been 23...


pramsey:
Closed out in SVN at r13787

PR #44, opened 2015-07-16
add memcmp short-circuit to ST_Equals

dbaston:
This can be a huge benefit for certain types of queries, for example looking for duplicate geometries that have different IDs:

    SELECT count(*)
    FROM inputs a, inputs b
    WHERE a.id < b.id AND ST_Equals(a.geom, b.geom);

This patch reduces the runtime of the above from 13 seconds (trunk) to 500 ms.

(patch at https://github.com/postgis/postgis/pull/44.patch)

### DIFF
--- a/postgis/lwgeom_geos.c
+++ b/postgis/lwgeom_geos.c
@@ -3097,6 +3097,14 @@ Datum ST_Equals(PG_FUNCTION_ARGS)
 		}
 	}
 
+	/*
+	 * short-circuit: if geom1 and geom2 are binary-equivalent, we can return
+	 * TRUE.  This is much faster than doing the comparison using GEOS.
+	 */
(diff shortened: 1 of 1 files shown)


strk:
Sounds good to me. @pramsey  ?

strk:
How many rows do you have in "inputs" ?
I guess it's the many negative results that usually drive the performance.

pramsey:
+1

dbaston:
In this case there are 6,708 polygons in the table, and `count(*)` returns 1,077 duplicate pairs.  Since the function has already passed a bounding-box equality test to get to this point, it seems there is a reasonable chance that the `memcmp` test will pass.

strk:
a ticket for this would also help not forgetting it

dbaston:
Here you go: https://trac.osgeo.org/postgis/ticket/3223#ticket

strk:
And here you go: https://trac.osgeo.org/postgis/changeset/13864 - Thanks!

PR #45, opened 2015-07-16
fix ST_Equals compiler warning

dbaston:
Looks like I introduced this one earlier when switching from `gserialized_get_gbox_p` to `gserialized_read_gbox_p`.  Alternative is to revert `ST_Equals` to use `gserialized_read_gbox_p`.  It seemed, though, that the intent of a short-circuit test was to use the box if it's available, not to force one to be constructed as the `get` variety does.  If this logic is sound, might want to change the other uses of `get` to `read`.

### DIFF
--- a/liblwgeom/liblwgeom.h.in
+++ b/liblwgeom/liblwgeom.h.in
@@ -1849,6 +1849,12 @@ extern GSERIALIZED* gserialized_from_lwgeom(LWGEOM *geom, int is_geodetic, size_
 extern LWGEOM* lwgeom_from_gserialized(const GSERIALIZED *g);
 
 /**
+ * Pull a #GBOX from the header of a #GSERIALIZED, if one is available.  If
+ * it is not, return LW_FAILURE.
+ */
+extern int gserialized_read_gbox_p(const GSERIALIZED *g, GBOX *gbox);
+
+/**
 * Pull a #GBOX from the header of a #GSERIALIZED, if one is available. If
 * it is not, calculate it from the geometry. If that doesn't work (null
 * or empty) return LW_FAILURE.


dbaston:
@pramsey this compiler warning has since been resolved with 992143a16cd0d3628e4286f6dfd83edab110c510 so I can close this, but I'm wondering what you think about the logic of `gserialized_get_gbox_p` vs `gserialized_read_gbox_p` for all of these short-circuit scenarios.

pramsey:
Actually I have strong misgivings, because of our use of float32 bounds on disk storage and frequent up-popping of float64 bounds during in-memory work (Ie, the boxes might be float64 until you serialize them and re-read them) which can lead to the same geometry having different boxes. Now, we may have plugged all the leaks in that abstraction, but then again we may not… I know we certainly had some issues with trying to circuit on boxes for precisely this reason.

dbaston:
Both `gserialized_get_gbox_p` and `gserialized_read_gbox_p` are going to be reading a float32 box though, right?  The difference is just that the `read` variety is going to construct a box if there isn't one?  That doesn't seem to line up with the logic expressed in the comments, for example in `ST_Contains` :

	/*
	 * short-circuit 1: if geom2 bounding box does not overlap
	 * geom1 bounding box we can prematurely return FALSE.
	 * Do the test IFF BOUNDING BOX AVAILABLE.
	 */

But the bounding box is always available, because `gserialized_read_gbox_p` is going to construct one, right?

pramsey:
Right, yes, since you’re reading from gserialized they’ve necessarily been float-rounded and serialized for you to even see them. So, OK, yes it would work. _get_ is also careful too float-round everything before it returns it, so that shouldn’t be a problem either.

strk:
The comment "IFF BOUNDING BOX AVAILABLE" is indeed confusing. Might refer to it being EMPTY, although often the check for EMPTY is performed before that.

dbaston:
I don't think it's worth making any code changes here --- a geometry is either going to be small, missing a box, and fast to deserialize, or it's going to be large enough to have a box and calling `gserialized_read_gbox_p` is not going to deserialize it.  Any disagreement with closing this PR and instead tweaking the comments to more clearly reflect what's happening?

pramsey:
Concur

PR #52, opened 2015-08-04
Stub implementation of lwt_GetFaceEdges

strk:
for travis eye...


strk:
Last bit to take care of seems to be re-ordering ring edges so they start with the edge with smallest identifier, I'll be looking at this tomorrow.

PR #54, opened 2015-08-06
Migrate Travis CI builds to container-based architecture

mloskot:
The `.travis.yml` update is sound.

## Status
Build is failing at the `installcheck` step due to issue:
* build generates `postgis-2.2`
* installcheck loads `postgis-2.1`
```
ERROR:  could not access file "$libdir/postgis-2.1": No such file or directory
```
* See detailed log at https://travis-ci.org/mloskot/postgis/builds/74320589

## Problem
[Pre-installed PostGIS](http://docs.travis-ci.com/user/using-postgresql/#Using-PostGIS) version conflicts with the build:
> All available versions of PostgreSQL come with PostGIS 2.1 packages preinstalled.

## Solution
* Install PostgreSQL from sources into `--prefix=$HOME/postgresql`, which does not require `sudo`.
* Alternatively, it may be possible to use sudo as per this message from the top of the build log:
> If you require sudo, add 'sudo: required' to your .travis.yml
> See http://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details.

However, `sudo: required` means switch back to old non-container architecture.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
   - ./configure
   - make
   - make check
-  - sudo make install
+  - make install
   - make installcheck
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,3 +46,6 @@ notifications:
     on_success: change
     on_failure: always
     use_notice: false
+
+after_failure:
+  - cat /tmp/pgis_reg/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+
 env:
   global:
     - RUNTESTFLAGS=-v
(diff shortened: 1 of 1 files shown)


strk:
The problem is basically with testing EXTENSION (that's what installcheck does).
Extensions need to be installed on the PostgreSQL extension dir.
Since container-based travis architecture does not allow us to write in there (does it?) we cannot install.
I'm not sure why the "sudo make install" step did not fail though...

mloskot:
@strk according to the [build 1 log](https://travis-ci.org/mloskot/postgis/builds/74320589) it failed:
```
The command "make check" exited with 0.
$ sudo make install
sudo: must be setuid root
The command "sudo make install" exited with 1.
```

strk:
Ok, great.
But it didn't interrupt the build ?
Can it be made fatal, to reduce noise ?

mloskot:
6 sie 2015 17:57 "Sandro Santilli" <notifications@github.com> napisał(a):
>
> But it didn't interrupt the build ?

"When one of the build commands returns a non-zero exit code, the Travis CI
build runs the subsequent commands as well, and accumulates the build
result."
http://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step

> Can it be made fatal, to reduce noise ?

Yes. Move all the `script` commands into shell script and check return
codes at each step, then report single code from the script as the build
status.

strk:
@molskot I've recently used container-based builds on gitlab-ci, and found out you don't need sudo because you're already root in there ! Maybe it is the same here, did you try just removing the sudo usage ? (and an whoami call among the steps would help too...).

Seen the poor man's waterfall display, btw ? :)
https://trac.osgeo.org/postgis/

pramsey:
@mloskot, @strk, is this still ongoing, or superceded?

strk:
I'm not moving any step here. This wasn't merged because failing.

pramsey:
If it's not going anywhere, close it out.

strk:
Up to @mloskot, it's his PR ...

mloskot:
Since the new container-based builders require no use of `sudo` whereas, AFAIK, we require it here, then I see no solution. Building PostgreSQL from sources makes little sense: it increases maintenance complexity and build startup performance gain would be cleared by the extended build time.

strk:
I get travis is different from github-ci in that reguard ?
https://github.com/postgis/postgis/pull/54#issuecomment-158306443

My understanding is that "sudo" is not allowed because you're already root (at least on github-ci shared builders).

Actually, now that I'm learning more about dockers it makes sense to be root inside a docker, as it's a jailed root.

mloskot:
@strk  I have no experience with github-ci, in fact, you've just made me aware of it.

PR #55, opened 2015-08-06
Implement lwt_RemoveIsoNode and stub lwt_MoveIsoNode

strk:
Adds deleteNodesById callback


strk:
pushed as fade90f099e0470d82bd7b3cd78eb44812711dbf (and this is really about MoveIsoNode)

PR #63, opened 2015-09-02
attempt to fix #3220

dbaston:
Posting here for Travis checking


robe2:
Merged on svn on r14035

PR #64, opened 2015-09-09
Have ./configure fail if LEX or YACC are missing

strk:
After writing the patch I realized we're shipping generated parser files with tarballs, so ./configure should NOT fail if lex/yacc are not found. So here's my patch, probably to remain here forever (ops).

Alternatively, someone might tweak the patch to check for the existance of the generated lexer files and avoid failing if the files are already there ?

\cc @pramsey @robe2

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,16 @@ AC_PROG_YACC
 AC_SUBST([LEX])
 AC_SUBST([YACC])
 
+if test "x$YACC" = "x"; then
+	dnl YACC was not found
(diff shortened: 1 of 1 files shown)


robe2:
Speaking of which, can we turn off the releases option on GitHub.  I don't want people getting confused thinking those are official releases, and those don't have generated anything.

strk:
It seems you can't (that's the price to pay for using external services).

pramsey:
No, don't fail.

PR #66, opened 2015-09-21
Do not have postgis module export library symbols (#3281)

strk:
Allows other modules to use a dynamically linked liblwgeom


strk:
Closing as a configure mediated version landed in trunk already

PR #70, opened 2015-10-22
remove/replace broken links in docs

dbaston:


strk:
Committed as r14299, thanks (sorry, http reference to PR in commit log did not work automatically)

PR #71, opened 2015-11-13
Trac #3366, read-write point iterator

dbaston:
https://trac.osgeo.org/postgis/ticket/3366

Posting for Travis testing/comments.

I'm still a bit pained every time I write out "LWPOINTITERATOR"....


dbaston:
Applied to trunk at r14400

PR #75, opened 2015-11-28
Trac #75 - Enhancement to PIP short-circuit (WIP)

dbaston:
https://trac.osgeo.org/postgis/ticket/75

Thought I'd test this out for `geos_intersects` to start.  This gives a nice performance increase (about 40% for my test data) with single-point multipoints, which for some reason are fairly common in data I work with.

The test data I used:
-2015 TIGER faces for FIPS 25001 (`tl_2015_25001_faces`)
-A fixed set of 100,000 random points within the extent of the faces above (`codpoints`)

I did testing using the following queries:
`select count(*) from codpoints p inner join tl_2015_25001_faces f on st_intersects(p.geom, f.geom);`
and
`select count(*) from codpoints p inner join tl_2015_25001_faces f on st_intersects(p.geom_multi, f.geom);`
where the contents of `geom` and `geom_multi` are identical, except that `geom_multi` is a multipoint.

Timings on svn-trunk:
single-point: ~13.5s
multi-point: ~30.5s

Timings with this PR:
single-point: ~13.5s
multi-point: ~17.3s

If there is support for the general approach I took here, I can extend it to other relevant predicates such as `ST_Contains` and `ST_Within`.


dbaston:
Applied to trunk at r14613

PR #76, opened 2015-11-29
Trac #3382 optimize index operations for small geometries

dbaston:
Full description at https://trac.osgeo.org/postgis/ticket/3382


dbaston:
Committed to trunk at r14478

PR #77, opened 2015-11-30
Trac #3365 remove code for unsupported versions of GEOS

dbaston:
https://trac.osgeo.org/postgis/ticket/3365


dbaston:
Committed at r14466

PR #81, opened 2015-12-23
window version of clustering functions [WIP]

dbaston:
Proof-of-concept.  To be determined:

- What do we name the window version of these functions?  Unless we get rid of the aggregate version, the window version will need a different name.
- What are the conventions, to be consistent across different clustering implementations for:
   * ids. Is a sequence starting at zero OK?  Can the sequence have gaps?
   * How are null inputs handled?  Can they just get a null cluster_id?
   * How are empty inputs handled?  Should they get a null cluster_id, or does each empty input go into its own cluster?
   * How do we handle inputs that are not considered to be in *any* cluster?  (This doesn't happen for currently-implemented algorithms, but it would happen for DBSCAN)


pnorman:
> Unless we get rid of the aggregate version, the window version will need a different name

Are you sure? There is both an avg() window function  and avg() aggregate function.

I was going to ask, do we need to have the aggregate version? I think the window function versions are always superior, and the aggregate functions haven't been present in a release.

dbaston:
Yes, I'm sure.  There's only one avg() function.  You can apply any aggregate function as a window function, which is what happens with avg.  That behavior wouldn't be very useful for ST_ClusterIntersecting or ST_ClusterWithin, because of their semantics (you'd get every cluster returned for each row).

I personally think the window implementation is superior, but the aggregate versions were already released in 2.2.

pramsey:
> How are null inputs handled? Can they just get a null cluster_id?

Yes, that is, not a cluster_id for NULL, but a cluster_id **of** NULL.

> How are empty inputs handled? Should they get a null cluster_id, or does each empty input go into its own cluster?

I think again a cluster_id **of** NULL would be fine. We can't drop them because it's a window, but they don't really semantically participate in the cluster, just like a NULL Input.

> How do we handle inputs that are not considered to be in any cluster? (This doesn't happen for currently-implemented algorithms, but it would happen for DBSCAN)

I think we have to either give them all individual unique cluster ids or one final id. ("The set of things not in sets is..."). Probably the latter would be more useful.

dbaston:
Closing this out.  In the short term, implement only `ST_ClusterDBSCAN` as a window function; anything that can be performed with `ST_ClusterIntersecting` and `ST_ClusterWithin` can be performed with this function.

PR #83, opened 2015-12-29
Use recursive snapping to improve predictability

strk:
Fixes geometry-intersects-edge exception when snapping twice
to the same pointset. See #2412.

### DIFF
--- a/liblwgeom/lwgeom_topo.c
+++ b/liblwgeom/lwgeom_topo.c
@@ -420,6 +420,43 @@ lwt_be_ExistsEdgeIntersectingPoint(LWT_TOPOLOGY* topo, LWPOINT* pt)
  *
  ************************************************************************/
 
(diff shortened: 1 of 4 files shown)


strk:
The new code seemed to enter an infinite loop on travis (3.3.8-CAPI-1.7.8) -- I guess this is fixed in 3.3.9, with the new snapping algorithm, so code in PostGIS should guard against that.

strk:
It would not be bad for Travis to test against GEOS-3.3.9, btw, as it's out since 2013-09-04

strk:
This needs some care as GEOS up to 3.3.8 had an unstable snapping algorithm that could keep adding duplicated vertices over and over and over ... (which results in an infinite loop as shown by travis builds)

strk:
committed upstream

PR #85, opened 2016-01-15
Trac #3428

dbaston:
https://trac.osgeo.org/postgis/ticket/3428


liblwgeom/cunit/cu_misc.c
```diff
@@ -156,6 +156,18 @@ static void test_clone(void)
...
+static void test_lwmpoint_from_lwgeom(void)
+{
+	/* This cast is so ugly, we only want to do it once.  And not even that. */
+	LWGEOM* (*transfn)(LWGEOM*) = (LWGEOM* (*)(LWGEOM*)) &lwmpoint_from_lwgeom;
+
+	do_transformation_test("MULTIPOLYGON (EMPTY)", "MULTIPOINT EMPTY", transfn);
+	do_transformation_test("POINT (30 10)", "MULTIPOINT ((30 10))", transfn);
+	do_transformation_test("LINESTRING Z (30 10 4,10 30 5,40 40 6)", "MULTIPOINT Z (30 10 4,10 30 5,40 40 6)", transfn);
+	do_transformation_test("POLYGON((35 10,45 45,15 40,10 20,35 10),(20 30,35 35,30 20,20 30))", "MULTIPOINT(35 10,45 45,15 40,10 20,35 10,20 30,35 35,30 20,20 30)", transfn);
+	do_transformation_test("MULTIPOINT M (10 40 1,40 30 2,20 20 3,30 10 4)", "MULTIPOINT M (10 40 1,40 30 2,20 20 3,30 10 4)", transfn);
```
strk: worth adding some curves here too, and TIN...

postgis/lwgeom_functions_basic.c
```diff
@@ -2806,3 +2806,26 @@ Datum ST_Scale(PG_FUNCTION_ARGS)
...
   PG_RETURN_POINTER(ret);
 }
+
+Datum ST_Points(PG_FUNCTION_ARGS);
+PG_FUNCTION_INFO_V1(ST_Points);
+Datum ST_Points(PG_FUNCTION_ARGS)
+{
+	if (PG_ARGISNULL(0))
+	{
+		PG_RETURN_NULL();
```
strk: just a note: you don't need to do that when the function is defined as STRICT, it's automatic.

dbaston: What happens if a function is called with `PGISDirectFunctionCall1`?  Does that bypass postgres' `STRICT` protection?

liblwgeom/cunit/cu_misc.c
dbaston: The function being tested is a pretty thin wrapper around the `LWPOINTITERATOR`, which has its own tests for all of these types.  I can add a couple additional cases here, though.

I think a larger issue is that it's currently a bit of a pain to add tests for all of the different types, especially when you think of all the different ways to get an EMPTY.  I think it would be worthwhile either to have some kind of random geometry generator (which you can start with a specified seed) or some standard lists of test geometries that can be pulled in to any individual test as desired.

postgis/lwgeom_functions_basic.c
strk: On Fri, Jan 15, 2016 at 04:17:03AM -0800, Dan Baston wrote:
> What happens if a function is called with `PGISDirectFunctionCall1`?  Does that bypass postgres' `STRICT` protection?

good question, better check

liblwgeom/cunit/cu_misc.c
strk: On Fri, Jan 15, 2016 at 04:20:20AM -0800, Dan Baston wrote:

> I think a larger issue is that it's currently a bit of a pain to add tests for all of the different types, especially when you think of all the different ways to get an EMPTY.  I think it would be worthwhile either to have some kind of random geometry generator (which you can start with a specified seed) or some standard lists of test geometries that can be pulled in to any individual test as desired.

+1 for sharing a common set of geoms for testing.
Random would be too much, for when we want to check against expected
behavior. You should find a complete coverage of types in
regress/typmod.

The random idea (fuzzing, in general) is good for the garden test.

--strk;

dbaston:
Committed at r14604

PR #87, opened 2016-02-03
Trac #3437, ST_Intersects incorrect for MultiPoints

dbaston:
https://trac.osgeo.org/postgis/ticket/3437


dbaston:
Interim (or final?) fix applied to trunk at r14686, 2.2 at r14687

strk:
I'd leave a POSTGIS_GEOS_VERSION macro, we know 3.6.0 will work.

dbaston:
Do we know that it's faster though?

strk:
Not w/out testing but if the comments and commit log refer
to a bug in GEOS it is useful to have immediate info about
which version of GEOS fixes that.

Then, if it's choosen not to use the prepared version for
points (or maybe small multipoints), it can still be done,
but not documenting the choice as driven by a bug...

PR #88, opened 2016-02-05
ST_GeometricMedian (trac #3364)

dbaston:
https://trac.osgeo.org/postgis/ticket/3364

I've been sitting on this one for a while with some mixed feelings.  It's a nice function that's more useful than a centroid in many situations.  But it's an approximate iterative algorithm that, while generally performing very well with real data, can be stumped with a contrived case like this:

```
postgres=# select st_astext(st_geometricmedian('MULTIPOINT (0 0, 1 1, 0 1, 2 2)'));
ERROR:  Median failed to converge within 1e-08 after 10000 iterations.

```

The convergence criteria can be adjusted in terms of the necessary tolerance, maximum iterations, and whether you even care if it converges or not, but I'm curious what others think about this.  Seems like most (all?) PostGIS functions are exact solutions.

### DIFF
--- a/doc/html/image_src/Makefile.in
+++ b/doc/html/image_src/Makefile.in
@@ -67,6 +67,7 @@ IMAGES= \
 	../images/st_extrude03.png \
 	../images/st_generatepoints01.png \
 	../images/st_generatepoints02.png \
+	../images/st_geometricmedian01.png \
 	../images/st_issimple01.png \
 	../images/st_issimple02.png \
 	../images/st_issimple03.png \
--- a/doc/html/image_src/st_geometricmedian01.wkt
+++ b/doc/html/image_src/st_geometricmedian01.wkt
@@ -1,0 +1,4 @@
+Style6;POINT(25 25)
+Style5;POINT(65 65)
+Style7-smallpoint;MULTIPOINT((20 20), (20 30), (30 20), (190 190))
+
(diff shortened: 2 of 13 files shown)


strk:
We also have ST_ConcaveHull returning a non-exact result.
I've no problem with this, but maybe add an optional parameter
to specify a custom limit on the number of iteration ?

Could it be worth for the function to return whatever approximation
was reached after the given number of iterations ?

dbaston:
Yes, the signature is:
`ST_GeometricMedian(g geometry, tolerance float8 DEFAULT 1e-8, max_iter int DEFAULT 10000, fail_if_not_converged boolean DEFAULT true)`

I'm wondering if this is too conservative though, for example if `fail_if_not_converged` should default to false.

strk:
I'm ok with `fail_if_not_converted` defaulting to false.
What happens if max_iter is given as 0 or negative ?
Is any NULL accepted ?
Should the default tolerance be a function of geometry extent
rather than a fixed number ?

pramsey:
Do you prefer to fail on non-convergence or just return what you happen to have after N iterations? (Admission: I prefer the latter)

dbaston:
I prefer to return what we happen to have (by default).  Before pushing
this in, I'd like to make the default tolerance a function of geometry
extent, as @strk suggests.

dbaston:
Committed at r14746

PR #90, opened 2016-02-18
Fix segmentize on geography

mhugo:
Segments were split on the straight line going from one vertex to the other and then "projected back" on the surface of the sphere. It resulted in segments with wrong distances.
Segments need to be computed based on equal angles between two adjacent vertices.


regress/geography_expected
```diff
@@ -27,3 +27,4 @@ geog_precision_pazafir|0|0
 #2422|1|1609|t|t|1400.230|1396.816|1400.230|1400.230
 #2422|1|1600|t|t|1400.230|1396.816|1400.230|1400.230
 #2422|1|1068|f|f|1400.230|1396.816|1400.230|1400.230
+segmentize_geography|t
```
strk: I think it would be better to give a rounded number than a boolean value.
Otherwise a max distance of 0 would still not be caught as a failure

mhugo:
@strk code updated with a rounded number in test

mhugo:
updated code that should now fix the broken regression test

pramsey:
Thanks, committed at https://trac.osgeo.org/postgis/changeset/14706

robe2:
As noted here - https://trac.osgeo.org/postgis/ticket/3539  this change broke ST_Segmentize functionality pretty badly.  Please fix or we'll have to revert.

mhugo:
@robe2 @pramsey Here is the new patch that fixes both segment distances and curvatures, using sphere_direction and sphere_project functions.

pramsey:
yeah, better to use that one. I have an approach that I think would be more efficient but I don't think I'll get it done in a short period of time...

robe2:
Thanks.  I'll test later today and then commit

robe2:
Test fails on sfcgal

` 
 regress_sfcgal .. ok
 sfcgal/empty .. ok
 sfcgal/geography .. ok
 sfcgal/legacy .. ok
 sfcgal/measures .. ok
 sfcgal/regress_ogc_prep .. ok
 sfcgal/regress_ogc .. ok
 sfcgal/regress .. ok
 sfcgal/tickets .. failed (diff expected obtained: /projects/postgis/tmp/2.3_pg9.6w64/test_126_diff)
 sfcgal/concave_hull .. ok
 sfcgal/wmsservers .. ok
 sfcgal/approximatemedialaxis .. ok
 uninstall .. ok (4578)

`

No worries, looks like you just forgot to make same change there.  I'll fix on my end.

mhugo:
@strk thanks !

strk:
Robe did all, only mirroring pushes are all in my name :)

PR #92, opened 2016-02-24
Fix cast from box3d to geometry to keep Z axis. (trac #3466)

rjuju:


dbaston:
Relevant (unresolved) postgis-devel discussion on the issue begins here: https://lists.osgeo.org/pipermail/postgis-devel/2016-February/025618.html

rjuju:
I've fixed the issue you raised during the code sprint (we should return a POLYHEDRAL instead of POLYGON when the box3d has all dimension non zero). I'm not familiar with the lwgeom API, so there may be some better way to return a POLYHEDRAL geometry. Please tell me if I should do things differently. I also added the lw*_free() calls that were already missing, based on the BOX2D_to_LWGEOM() function.

Results are now:

```
# select st_astext(box3d(st_geomfromtext(
        'LINESTRING(0 0 0, 0 0 0)')));
    st_astext    
═════════════════
 POINT Z (0 0 0)

# select st_astext(box3d(st_geomfromtext(
        'LINESTRING(0 0 0, 1 0 0)')));
         st_astext          
════════════════════════════
 LINESTRING Z (0 0 0,1 0 0)

# select st_astext(box3d(st_geomfromtext(
        'LINESTRING(0 0 0, 1 1 0)')));
                  st_astext                  
═════════════════════════════════════════════
 POLYGON Z ((0 0 0,0 1 0,1 1 0,1 0 0,0 0 0))

# select st_astext(box3d(st_geomfromtext(
        'LINESTRING(0 0 0, 1 1 1)')));
                                                                                                             st_astext                                                                                                             
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
 POLYHEDRALSURFACE Z (((0 0 0,0 1 0,1 1 0,1 0 0,0 0 0)),((0 0 1,0 1 1,1 1 1,1 0 1,0 0 1)),((0 0 0,0 0 1,0 1 1,0 1 0,0 0 0)),((1 0 0,1 0 1,1 1 1,1 1 0,1 0 0)),((0 0 0,0 0 1,1 0 1,1 0 0,0 0 0)),((0 0 0,0 0 1,1 0 1,1 0 0,0 0 0)))

```

dbaston:
Yikes, building a polyhedral surface seems to be quite involved.  What do you think about the following as a way to cut down the code volume?

First, reate a fixed array of points, populate it with the 8 points of the cube:
```
POINT3D points[10];
points[0] = { box->xmin, box->ymin, box->zmin };
points[1] = { box->xmin, box->ymax, box->zmin };
etc.
```
Then make a helper function to create a 3d polygon given four points.

rjuju:
I wasn't quite happy with current implementation, so I like this approach.

The downside is that such a function would require 7 arguments (srid, gbox, nrings and the 4 points). If it's ok, I'll update current code as you suggested.

dbaston:
I think you can get by with just the four points as arguments, no?  Within the function body, when you call `lwpoly_construct`, you can use `srid = SRID_UNKNOWN, gbox = NULL, nrings = 1`.

rjuju:
I thought about this, but it'd make this wrapper function very specific. I'm not familiar with api, but if it's it possible to change these parameters on each element of the collection I suppose it's the best tradeoff.

rjuju:
Last commit implement this refactoring, and it's realler cleaner. Thanks for the design :)

I also fixed a bug I didn't notice with previous implementation (polygon 5 and 6 was the same, stupid copy/paste error).

dbaston:
Nice.  Why not use `lwpoly_construct_rectangle` in all the cases where you're building a rectangle?  Also, is there a reason to use `POINT4D` instead of `POINT3D`?

rjuju:
I didn't use `lwpoly_construct_rectangle` in all cases to keep some consistency between the 2d and 3d cast functions and keep the patch minimal. I can update both  to use this new `lwpoly_construct_rectangle` function is you prefer.

The ptarray functions (ptarray_insert_point and ptarray_append_point) uses `POINT4D` parameters. A pointarray has z and m flags, so I assumed internal functions would handle less dimensions if needed.

dbaston:
I'd go with `lwpoly_construct_rectangle` wherever you can - I think it make it easier to read.  And your point about the `POINT4D` makes sense, thanks.

postgis/lwgeom_box3d.c
```diff
@@ -206,62 +206,194 @@ Datum BOX3D_to_LWGEOM(PG_FUNCTION_ARGS)
...
+		points[0] = (POINT4D) { box->xmin, box->ymin, box->zmin };
+		points[1] = (POINT4D) { box->xmin, box->ymax, box->zmin };
+		points[2] = (POINT4D) { box->xmax, box->ymax, box->zmin };
+		points[3] = (POINT4D) { box->xmax, box->ymin, box->zmin };
+		points[4] = (POINT4D) { box->xmin, box->ymin, box->zmax };
+		points[5] = (POINT4D) { box->xmin, box->ymax, box->zmax };
+		points[6] = (POINT4D) { box->xmax, box->ymax, box->zmax };
+		points[7] = (POINT4D) { box->xmax, box->ymin, box->zmax };
+
+		geoms = (LWGEOM **) lwalloc(sizeof(LWGEOM *) * ngeoms);
```
dbaston: Maybe worth exploring if these geometries can be constructed and serialized without heap allocations?  I'm happy to look into it later unless you want to.

regress/regress_expected
```diff
@@ -87,7 +87,7 @@ ERROR:  parse error - invalid geometry at character 23
 76|f
 76a|f
 106|BOX3D(0 0 0,7 7 0)
-107|POLYGON((0 0,0 7,7 7,7 0,0 0))
+107|POLYHEDRALSURFACE(((0 0 0,0 7 0,7 7 0,7 0 0,0 0 0)),((0 0 7,0 7 7,7 7 7,7 0 7,0 0 7)),((0 0 0,0 0 7,0 7 7,0 7 0,0 0 0)),((7 0 0,7 0 7,7 7 7,7 7 0,7 0 0)),((0 0 0,0 0 7,7 0 7,7 0 0,0 0 0)),((0 7 0,0 7 7,7 7 7,7 7 0,0 7 0)))
```
dbaston: We should also have tests to catch the potential 3D polygon outputs from this cast.

postgis/lwgeom_box3d.c
rjuju: The points var is already stack memory

geoms can be made stack easily too (switch to LWGEOM *geoms[6], and some fixes to the destructor).

The lwcollection (geom here) can be entirely build in stack memory, but I think it's too much knowledge of the datatype.

I don't really see how to avoid heap allocations in lwpoly_construct_rectangle without redefining a pointarray and LWPOLY from scratch like the lwcollection. I think it's the most expensive part of this code path though.

A quick example with geoms and the lwcollection in stack:
```C
		POINT4D points[8];
		LWGEOM *geoms[6];
		LWCOLLECTION collection;
		int ngeoms = 6, i;

		/* Initialize the 8 vertices of the box */
		points[0] = (POINT4D) { box->xmin, box->ymin, box->zmin };
		points[1] = (POINT4D) { box->xmin, box->ymax, box->zmin };
		points[2] = (POINT4D) { box->xmax, box->ymax, box->zmin };
		points[3] = (POINT4D) { box->xmax, box->ymin, box->zmin };
		points[4] = (POINT4D) { box->xmin, box->ymin, box->zmax };
		points[5] = (POINT4D) { box->xmin, box->ymax, box->zmax };
		points[6] = (POINT4D) { box->xmax, box->ymax, box->zmax };
		points[7] = (POINT4D) { box->xmax, box->ymin, box->zmax };

		/* add bottom polygon */
		geoms[0] = (lwpoly_as_lwgeom(lwpoly_construct_rectangle(&points[0],
						&points[1], &points[2], &points[3])));
		/* add top polygon */
		geoms[1] = (lwpoly_as_lwgeom(lwpoly_construct_rectangle(&points[4],
						&points[5], &points[6], &points[7])));
		/* add left polygon */
		geoms[2] = (lwpoly_as_lwgeom(lwpoly_construct_rectangle(&points[0],
						&points[4], &points[5], &points[1])));
		/* add right polygon */
		geoms[3] = (lwpoly_as_lwgeom(lwpoly_construct_rectangle(&points[3],
						&points[7], &points[6], &points[2])));
		/* add back polygon */
		geoms[4] = (lwpoly_as_lwgeom(lwpoly_construct_rectangle(&points[0],
						&points[4], &points[7], &points[3])));
		/* add front polygon */
		geoms[5] = (lwpoly_as_lwgeom(lwpoly_construct_rectangle(&points[1],
						&points[5], &points[6], &points[2])));

		collection.type = POLYHEDRALSURFACETYPE;
		collection.flags = gflags(1, 0, 0);
		FLAGS_SET_BBOX(collection.flags, 0);
		collection.bbox = NULL;
		collection.srid = SRID_UNKNOWN;
		collection.ngeoms = ngeoms;
		collection.maxgeoms = ngeoms;
		collection.geoms = geoms;
		FLAGS_SET_SOLID(collection.flags, 1);

		result = geometry_serialize((LWGEOM *) &collection);

		for (i=0; i<ngeoms; i++)
			lwpoly_free((LWPOLY *) geoms[i]);
```

Was is something like that you had in mind?

regress/regress_expected
rjuju: Do you mean to add tests for all the different geometries the cast function can return? (point, line and rectangle)

postgis/lwgeom_box3d.c
dbaston: Agreed -- too much knowledge of the `LWCOLLECTION` data type.  Let's leave as-is.

regress/regress_expected
dbaston: Point and line would be good too, thought I was thinking of just the three different code paths for generating a polygon (flat in z plane, flat in y plane, flat in x plane)

rjuju:
Last commits add regress test for all box3d::geometry code pathes and consistent use of `lwpoly_construct_rectangle`, for box3d and box2d cast functions.

I checked, the box2d cast generating code path is covered by multiple test (in_geohash.sql tests geomfromgeohash_01 to geomfromgeohash_04 and ticket.sql #3300 for instance).

I had to reintroduce hasz and hazm parameters to the new wrapper function, but it indeed seems a much better idea to have them than to blindly construct 3D polygon.

dbaston:
Very nice, thank you!  I made just a few changes -- mostly, using distinct ordinate values in the test cases so that we can check point order, and then reordering some of the points so that the generated polygons (and faces of the polyhedral surface) follow the right-hand rule.  Applied to trunk at r14786.

rjuju:
Thank you!

Sorry, I wasn't aware of the right-hand rule, I'll check the commit.

PR #94, opened 2016-02-24
ST_AsText : adding optional argument 'precision' 

marcu:
The optional argument 'precision' is used for specifying the number of significant digits in the output doubles

### DIFF
--- a/postgis/lwgeom_ogc.c
+++ b/postgis/lwgeom_ogc.c
@@ -848,14 +848,20 @@ Datum LWGEOM_asText(PG_FUNCTION_ARGS)
 	char *wkt;
 	size_t wkt_size;
 	text *result;
+	int dbl_dig_for_wkt = DBL_DIG;
 
 	POSTGIS_DEBUG(2, "Called.");
 
 	geom = PG_GETARG_GSERIALIZED_P(0);
 	lwgeom = lwgeom_from_gserialized(geom);
 
+	if (PG_NARGS() > 1)
+	{
+		dbl_dig_for_wkt = PG_GETARG_INT32(1);
+	}
+
 	/* Write to WKT and free the geometry */
-	wkt = lwgeom_to_wkt(lwgeom, WKT_ISO, DBL_DIG, &wkt_size);
+	wkt = lwgeom_to_wkt(lwgeom, WKT_ISO, dbl_dig_for_wkt, &wkt_size);
 	lwgeom_free(lwgeom);
 	POSTGIS_DEBUGF(3, "WKT size = %u, WKT length = %u", (unsigned int)wkt_size, (unsigned int)strlen(wkt));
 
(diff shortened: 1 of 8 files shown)


mwtoews:
This is related to http://trac.osgeo.org/postgis/ticket/407 enhancement idea, which woud use `extra_float_digits` for all ST_As* fuctions. I wonder if a default should be obtained from that global.

postgis/lwgeom_ogc.c
```diff
@@ -848,14 +848,20 @@ Datum LWGEOM_asText(PG_FUNCTION_ARGS)
...
 	POSTGIS_DEBUG(2, "Called.");
 
 	geom = PG_GETARG_GSERIALIZED_P(0);
 	lwgeom = lwgeom_from_gserialized(geom);
 
+	if (PG_NARGS() > 1)
+	{
+		dbl_dig_for_wkt = PG_GETARG_INT32(1);
+	}
+
```
mwtoews: Continuing on with the idea from https://trac.osgeo.org/postgis/ticket/407, a default precision could be placed in the "else" part here by retrieving the `extra_float_digits` global, and something linke `dbl_dig_for_wkt = DBL_DIG + extra_float_digits;` or `dbl_dig_for_wkt += extra_float_digits`

This is based on the [PostgreSQL docs](http://www.postgresql.org/docs/current/static/runtime-config-client.html)
> The parameter value [extra_float_digits] is added to the standard number of digits (`FLT_DIG` or `DBL_DIG` as appropriate).

dbaston:
Ideally `ST_AsText` would behave similarly to other text-producing functions....so the current output of `POINT(1.1e+02 1.1)` in the doc example should be `111.11` to be consistent with functions like `ST_AsKML` and `ST_AsGeoJSON`.

dbaston:
> I wonder if a default should be obtained from that global.

Whether we need/want to change the defaults is probably outside the scope of this improvement.

mwtoews:
Right, setting defaults is not directly related to this enhancement, and should be ignored here.

vesnikos:
At some cases, the scientific format might cause some issues. See 
http://gis.stackexchange.com/questions/190732/postgis-floating-point-representation-problem-with-st-astext

robe2:
Are we done with discussion of this?  Is this ready for commit?

regress/out_geography_expected
```diff
@@ -66,3 +66,5 @@ geojson_options_13|{"type":"LineString","crs":{"type":"name","properties":{"name
 geojson_options_14|{"type":"LineString","crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:EPSG::4326"}},"coordinates":[[1,1],[2,2],[3,3],[4,4]]}
 geojson_options_15|{"type":"LineString","crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:EPSG::4326"}},"bbox":[1,1,4,4],"coordinates":[[1,1],[2,2],[3,3],[4,4]]}
 geojson_options_16|{"type":"LineString","crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:EPSG::4326"}},"bbox":[1,1,4,4],"coordinates":[[1,1],[2,2],[3,3],[4,4]]}
+text_precision_01|POINT(111.1111111 1.1111111)
+text_precision_02|POINT(1.1e+2 1.1)
```
Komzpa: is 1.1e+2 a desired result in this case?

Komzpa:
To get this merged we need:

- [ ] ticket in trac
- [ ] rebase on trunk
- [ ] change precision to mean "number of signs after the point" similar to ST_AsGeoJSON
- [ ] get rid of scientific format, test on values too large (100000000000) and too small (0.000000001)

@marcu do you need any help with that? :)

Komzpa:
ticketed: https://trac.osgeo.org/postgis/ticket/3986

Komzpa:
@marcu I did the switch from "number of significant digits" to "number of digits after comma" in https://github.com/postgis/postgis/pull/190. Are you available for a rebase of this PR on top of that one? :)

Komzpa:
Let's go to #194 which has the branch rebased. (wow, exactly 100 pull requests later!)

PR #95, opened 2016-02-24
Allowing ST_Centroid to work on Circularstring and coumpoundgeometry

Remi-C:
With test and doc


pramsey:
Thanks, in https://trac.osgeo.org/postgis/changeset/14696/

Remi-C:
Hey,
can't do a angle(line1,line2) because in postgis.sql.in st_startPoint is defined after st_angle.
Cheers,
Rémi-C

dbaston:
This breaks the build, since any changes made to `regress_ogc_expected` need to be duplicated in the SFCGAL version of `regress_ogc_expected`.  We need to fix this situation.....

strk:
yeah, getting travis to buidl with SFCGAL would be an excellent start. As you have Osllandia in Paris, you might raise the point. See https://trac.osgeo.org/postgis/ticket/2778

PR #97, opened 2016-02-25
NEw ST_Angle function, with test and doc

Remi-C:
Hey,
new function with 2 behaviours,
ST_Angle(P1,P2,P3) returns the angle of P1P2P3
ST_ANgle(P1,P2,P3,P4) returns the angle of 2 vectors P1P2,P3P4

For an easy visual test, use the query in the doc to generate a few dozen angles, and display in qgis.
Cheers

### DIFF
--- a/doc/reference_measure.xml
+++ b/doc/reference_measure.xml
@@ -825,7 +825,7 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 		<refnamediv>
 		  <refname>ST_Angle</refname>
 
-		  <refpurpose>Returns the angle in radian measured clokwise between 3 points, or the angle measured clockwise between 2 vectors. Vectors can be described with 4 points or 2 lines.</refpurpose>
+		  <refpurpose>Returns the angle between 3 points, or between 2 vectors (4 points or 2 lines).</refpurpose>
 		</refnamediv>
 		<refsynopsisdiv>
 		  <funcsynopsis>
@@ -834,7 +834,7 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 			  <paramdef><type>geometry </type><parameter>point1</parameter></paramdef>
 			  <paramdef><type>geometry </type><parameter>point2</parameter></paramdef>
 			  <paramdef><type>geometry </type><parameter>point3</parameter></paramdef>
-			  <paramdef><type>geometry </type><parameter>point4</parameter></paramdef>
+			  <paramdef choice="opt"><type>geometry </type><parameter>point4</parameter></paramdef>
 			</funcprototype> 
 			<funcprototype>
 			  <funcdef>float <function>ST_Angle</function></funcdef>
@@ -851,7 +851,7 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 			For 4 points,compute the angle measured clockwise of P1P2,P3P4.
 			Results are always positive, between 0 and 2*Pi radians.
 			
-			Uses ST_Azimuth.
+			Uses azimuth of pairs or points.
 			</para>
 			<para>ST_Angle(P1,P2,P3) = ST_Angle(P2,P1,P2,P3)</para> 
 			<para>Result is in radian and can be converted to degrees using a built-in PostgreSQL function degrees(), as shown in the example.</para> 
--- a/doc/reference_measure.xml
+++ b/doc/reference_measure.xml
@@ -825,7 +825,7 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 		<refnamediv>
 		  <refname>ST_Angle</refname>
 
-		  <refpurpose>Returns the angle in radian measured clokwise between 3 points, or the angle measured clockwise between 2 vectors.</refpurpose>
+		  <refpurpose>Returns the angle in radian measured clokwise between 3 points, or the angle measured clockwise between 2 vectors. Vectors can be described with 4 points or 2 lines.</refpurpose>
 		</refnamediv>
 		<refsynopsisdiv>
 		  <funcsynopsis>
@@ -836,14 +836,21 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 			  <paramdef><type>geometry </type><parameter>point3</parameter></paramdef>
 			  <paramdef><type>geometry </type><parameter>point4</parameter></paramdef>
 			</funcprototype> 
+			<funcprototype>
+			  <funcdef>float <function>ST_Angle</function></funcdef>
+			  <paramdef><type>geometry </type><parameter>line1</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>line2</parameter></paramdef>
+			</funcprototype>
 		  </funcsynopsis>
 		</refsynopsisdiv>
 		<refsection>
 			<title>Description</title>
 
 			<para> For 3 points, computes the angle measured clockwise of P1P2P3.
+			If input are 2 lines, get first and last point of the lines as 4 points.
 			For 4 points,compute the angle measured clockwise of P1P2,P3P4.
 			Results are always positive, between 0 and 2*Pi radians.
+			
 			Uses ST_Azimuth.
 			</para>
 			<para>ST_Angle(P1,P2,P3) = ST_Angle(P2,P1,P2,P3)</para> 
--- a/doc/reference_measure.xml
+++ b/doc/reference_measure.xml
@@ -820,6 +820,64 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 		</refsection>
 
   </refentry>
+  
+  	<refentry id="ST_Angle">
+		<refnamediv>
+		  <refname>ST_Angle</refname>
+
+		  <refpurpose>Returns the angle in radian measured clokwise between 3 points, or the angle measured clockwise between 2 vectors.</refpurpose>
+		</refnamediv>
+		<refsynopsisdiv>
+		  <funcsynopsis>
+			<funcprototype>
+			  <funcdef>float <function>ST_Angle</function></funcdef>
+			  <paramdef><type>geometry </type><parameter>point1</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>point2</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>point3</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>point4</parameter></paramdef>
+			</funcprototype> 
+		  </funcsynopsis>
+		</refsynopsisdiv>
+		<refsection>
+			<title>Description</title>
+
+			<para> For 3 points, computes the angle measured clockwise of P1P2P3.
+			For 4 points,compute the angle measured clockwise of P1P2,P2P3.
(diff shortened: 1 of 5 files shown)


postgis/lwgeom_functions_basic.c
```diff
@@ -2399,6 +2400,88 @@ Datum LWGEOM_azimuth(PG_FUNCTION_ARGS)
...
+	for(i=0; i<n_args; i++)
+	{
+		geom = PG_GETARG_GSERIALIZED_P(i);
+		geom_unser = lwgeom_from_gserialized(geom) ; 
+		if (lwgeom_is_empty(geom_unser) == true)
+		{/*4th arg is not set, don't use it*/
+			n_args = i ; 
+			if (n_args < 3)
+				lwpgerror("Empty geometry"); 
+			break;
```
strk: Should this be a PG_RETURN_NULL() instead ?

doc/reference_measure.xml
```diff
@@ -820,6 +820,64 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
 		</refsection>
 
   </refentry>
+  
+  	<refentry id="ST_Angle">
+		<refnamediv>
+		  <refname>ST_Angle</refname>
+
+		  <refpurpose>Returns the angle in radian measured clokwise between 3 points, or the angle measured clockwise between 2 vectors.</refpurpose>
```
strk: "clokwise" -> "clockwise" (first occurrence)

If I may suggest also keeping the "refpurpose" text small (may omit radian and clockwise, leaving it specified in the "description")

doc/reference_measure.xml
```diff
@@ -820,6 +820,64 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
...
+		</refnamediv>
+		<refsynopsisdiv>
+		  <funcsynopsis>
+			<funcprototype>
+			  <funcdef>float <function>ST_Angle</function></funcdef>
+			  <paramdef><type>geometry </type><parameter>point1</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>point2</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>point3</parameter></paramdef>
+			  <paramdef><type>geometry </type><parameter>point4</parameter></paramdef>
```
strk: Isn't the 4th point optional ? There should be a tag or attribute to specify that in the docs.

doc/reference_measure.xml
```diff
@@ -820,6 +820,64 @@ SELECT degrees(ST_Azimuth(ST_Point(25, 45), ST_Point(75, 100))) AS degA_B,
...
+			</funcprototype> 
+		  </funcsynopsis>
+		</refsynopsisdiv>
+		<refsection>
+			<title>Description</title>
+
+			<para> For 3 points, computes the angle measured clockwise of P1P2P3.
+			For 4 points,compute the angle measured clockwise of P1P2,P3P4.
+			Results are always positive, between 0 and 2*Pi radians.
+			Uses ST_Azimuth.
```
strk: "Uses ST_Azimuth" isn't really true, it's using liblwgeom internal functions (on which ST_Azimuth is also based).

postgis/lwgeom_functions_basic.c
strk: Sorry, I meant within the conditional block, not outside

postgis/lwgeom_functions_basic.c
```diff
@@ -2399,6 +2400,88 @@ Datum LWGEOM_azimuth(PG_FUNCTION_ARGS)
...
+			n_args = i ; 
+			if (n_args < 3)
+				lwpgerror("Empty geometry"); 
+			break; 
+		}
+		
+		lwpoint = lwgeom_as_lwpoint(geom_unser);
+		if ( ! lwpoint )
+		{
+			PG_FREE_IF_COPY(geom, i);
```
strk: These memory releases should really affect all the previous geometry inputs, or only the last one would be released.

postgis/lwgeom_functions_basic.c
```diff
@@ -2399,6 +2400,88 @@ Datum LWGEOM_azimuth(PG_FUNCTION_ARGS)
...
+			lwpgerror("Argument must be POINT geometries");
+			PG_RETURN_NULL();
+		}
+		srids[i] = lwpoint->srid;
+		/* Extract first point */
+		if ( ! getPoint2d_p(lwpoint->point, 0, &points[i]) )
+		{
+			PG_FREE_IF_COPY(geom, i);
+			lwpgerror("Error extracting point");
```
strk: This should really only happen when the point is empty (hint hint [ for reducing complexity ])

postgis/lwgeom_functions_basic.c
```diff
@@ -2399,6 +2400,88 @@ Datum LWGEOM_azimuth(PG_FUNCTION_ARGS)
...
+				PG_RETURN_NULL();
+			}
+		}
+	}
+	/* compute azimuth for the 2 pairs of points
+	 * note that angle is not defined identically for 3 points or 4 points*/ 
+	if (n_args == 3)
+	{
+		if ( ! azimuth_pt_pt(&points[0], &points[1], &az1) )
+			PG_RETURN_NULL();
```
strk: I guess the code here expects azimuth_pt_pt to call lwerror itself, better specify it with a comment, or a reader would be wondering why no error message is printed.

An lwpgerror could even be left to be even ready shall liblwgeom stop reporting errors in such a primitive way :)

regress/lwgeom_regress_expected
```diff
@@ -21,3 +21,11 @@ BoundingDiagonal3|SRID=4326;LINESTRING(999999986991104 999999986991104,999999986
...
 BoundingDiagonal5|SRID=3857;LINESTRINGM(4 4 0,5 4 1)
 BoundingDiagonal6|SRID=3857;LINESTRINGM EMPTY
+ST_Azimuth_regular|2.3562
+ST_Azimuth_same_point|
+ERROR:  Operation on mixed SRID geometries
+ERROR:  Argument must be POINT geometries
+ST_Angle_4_pts|4.71238898038469
+ST_Angle_4_pts|0.785398163397448
+ST_Angle_3_pts|1.5707963267949
+ERROR:  Operation on mixed SRID geometries
```
strk: The EMPTY case is missing a test

Remi-C:
Hey @strk , thanks for the review!
I don't get the PG_RETURN_NULL stuff,
what do you mean exactly?
I corrected the other things you mention, know working on the memory stuff.
Cheers

regress/lwgeom_regress_expected
Remi-C: Hey,
I removed the empty test case because ST_Azimuth is failing it.
(it returns another error than the one expected)
it returns `ERROR:  getPoint2d_p: point offset out of range`

Remi-C: hm never mind its ok I added the missing test, thanks

Remi-C: hm never mind its ok I added the missing test, thanks

strk: > I removed the empty test case because ST_Azimuth is failing it.
> (it returns another error than the one expected)
> it returns `ERROR:  getPoint2d_p: point offset out of range` 

Evidently your is_empty calls are not working.
Better fix that first.

strk:
I just meant that any "lwpgerror" call should be followed by
a PG_RETURN_NULL(), which wasn't done in the n_args < 3
check (btw, on re-reading that check is probably bogus
and the source of your problems with empty ?)

Remi-C:
@strk I don't get the  memory issue you mention.
`POINT2D points[4]; ` doesn't need to be freed manually I guess.
Other variables are freed in each loop, so no need to free those in case of error some loop after?
Do you mean I should free `geom_unser` and / or `lwpoint`
before each `PG_RETURN_NULL` ? 
Cheers

strk:
I mean the PG_FREE_IF_COPY calls you add.
They are inconsistent.
Ideally you'd have a PG_FREE_IF_COPY call for each
and every parameter, before exiting or raising an error.

Remi-C:
Hey @strk, here is a rewrite,
with better separation of exception and actual computing.
Exception are tried on serialized geometry.
About  `PG_FREE_IF_COPY`,
from what I understand there is no need to use it on a parameter if you havn't used `PG_GETARG_GSERIALIZED_P`.
If this is true, it should be ok as it is now (only one arg is in memory at a time, being cleaned at the end of the loop anyway)
Cheers,
Remi-C

strk:
Ok so I see you're detoasting/free-if-copying once for checking and again for actual using.
Could you avoid the duplication ?

postgis/lwgeom_functions_basic.c
```diff
@@ -2399,6 +2400,124 @@ Datum LWGEOM_azimuth(PG_FUNCTION_ARGS)
...
+	/* extract points */
+	for(i=0; i<n_args; i++)
+	{
+		geom = PG_GETARG_GSERIALIZED_P(i);
+		geom_unser = lwgeom_from_gserialized(geom) ; 
+		lwpoint = lwgeom_as_lwpoint(geom_unser); 
+		if (!lwpoint)
+		{ 
+			lwpgerror("Error unserializing geometry"); 
+			PG_RETURN_NULL() ;
```
strk: Also this misses a PG_FREE_IF_COPY...

Remi-C: @strk Ok as you suggested I stored the serialized pointers to keep getting
those twice.
Now `PG_FREE_IF_COPY` is always executed if exceptions or in regular
operations.

2016-02-26 14:47 GMT+01:00 Sandro Santilli <notifications@github.com>:

Remi-C:
Hey @strk , here the pointers are stored to avoid calling twice the functions.
Cheers

postgis/lwgeom_functions_basic.c
```diff
@@ -2399,6 +2400,129 @@ Datum LWGEOM_azimuth(PG_FUNCTION_ARGS)
...
+			for (j=0;j<n_args;j++)
+				PG_FREE_IF_COPY(seri_geoms[j], j);
+			lwpgerror("Error extracting point");
+			PG_RETURN_NULL();
+		}
+		lwfree(geom_unser);
+		/* dont do lwpoint_free(lwpoint); , this memory is needed ! */ 
+	}
+	for (j=0;j<n_args;j++)
+		PG_FREE_IF_COPY(seri_geoms[j], j);
```
strk: It is not safe to release the serialized format after deserializing, because LWGEOM can still refer to the serialized POINTLIST.

Remi-C:
@strk : thanks to help me learn all of this.
Cheers,
RemiC

robe2:
@Remi-C  looks like this isn't ready yet as @strk had one last note doesn't look like you addressed.  Also this needs to be rebased if to be committed for PostGIS 2.4.

robe2:
@Remi-C  can you rebase and I'll test and  commit for PostGIS 2.4

Remi-C:
Hey Regina, @strk,
I removed the serialized memory freeing.

About my rebasing,
I think I messed something,
I rebased so my branch st_angle is updated and contains all the latest
postgis commits.
What I don't understand is why now the pull request contains also all the
postgis commits?

Shall we drop this pull request and make a new one?
Cheers,
Rémi-C

2017-08-10 12:09 GMT+02:00 Regina Obe <notifications@github.com>:

strk:
Try this (assuming "origin" points to the official mirror):

  git rebase origin
  git push -f

robe2:
keep in mind it's too late for this to go into 2.4.  Has to wait for 2.5

Remi-C:
Hey @robe2, no problem, it'll be for 2.5 with any luck.
Thanks @strk, this fixed it (after some awkward manual edits)

Minor typos

PR #99, opened 2016-02-26
Geometry validity flag

mhugo:
This adds a flag to mark a geometry as valid in order to shortcut validity tests (initially an SFCGAL need).

New functions introduced:
- ST_Validate: calls isValid() on the geometry and returns it with the validity flag set if it is valid. Returns null otherwise
- ST_HasValidityFlag: tests the presence of the validity flag
- ST_IsValid and ST_IsValidDetail in SFCGAL backend


strk:
If you're changing the SERIALIZED_FORMAT I'd prefer to read a formal RFC or at least a thread on the development mailing list for it.

dbaston:
@strk see https://github.com/postgis/postgis/blob/svn-trunk/liblwgeom/liblwgeom.h.in#L137

strk:
Uhm, so merged already ? Or what did I miss ?

dbaston:
The flag was defined earlier this week, but this PR hasn't been merged yet.

Komzpa:
Are other functions, like ST_SetPoint, resetting the validity flag? This may become a large pain if they don't...

robe2:
Are we all set with this or more work needed here?  In which case we should probably punt til 2.4

robe2:
Where are we with this?

robe2:
I ask again what's going on with this @mhugo , @strk, @dbaston

Komzpa:
Close/open to restart CI tests on current codebase.

mhugo:
Hi, we should (finally) have some time to go on on this and hopefully finish in the coming days / weeks :)

From what I see, apart from checking that the code still compiles against current trunk, what's missing:
- list functions that touch geometries "in place", as @Komzpa said. They should be modified to make sure they reset the validity flag
- probably a function to force the validity flag. But it has to be stated very clearly somewhere that it is at user's own risk.

Let me know if you see other points

mhugo:
I forgot we had a discussion about flags in a GSERIALIZED header http://postgis.17.x6.nabble.com/Validity-flag-td5009775.html

So we should come first with a proposal for an "extended flags" flag

PR #100, opened 2016-02-26
Allow fixed PrecisionModel for overlay operations [WIP]

dbaston:
Implemented this for `ST_Intersection`, but should be easy to extend to other applicable functions.  Can be used like:

```
pg23=# SELECT ST_AsText(ST_Intersection('LINESTRING (0 0, 0.9 1)', 'LINESTRING(0 1, 1 0)'));
                 st_astext                  
--------------------------------------------
 POINT(0.473684210526316 0.526315789473684)
(1 row)

Time: 17.599 ms
pg23=# SELECT ST_AsText(ST_Intersection('LINESTRING (0 0, 0.9 1)', 'LINESTRING(0 1, 1 0)', op_precision := 1e-04));
      st_astext       
----------------------
 POINT(0.4737 0.5263)
(1 row)

```


strk:
Any chance to encode the precision into the serialized format, like it was done for validity ?

dbaston:
Maybe a future item, but there's no way to sneak it into the existing format.

mwtoews:
Note that this functionality is similar to ST_SnapToGrid:
```
SELECT ST_AsText(ST_SnapToGrid(ST_Intersection('LINESTRING (0 0, 0.9 1)', 'LINESTRING(0 1, 1 0)'), 1e-04));
      st_astext
----------------------
 POINT(0.4737 0.5263)
(1 row)
```

strk:
Snapping to a grid may break the topology of a geometry.
Instead, requesting a fixed precision to GEOS should never
return a topologically-invalid result.

Worth finding a similar case to include it as a testcase.

robe2:
Dan based on what you said, I guess we should push this to 2.3.  I'm also not confident we'll release GEOS 3.6 before PostGIS 2.3.  @strk has a habit of dragging his feet on those things and I don't want to hold up release for that.  I know we have ST_MinimumClearance and ST_MinimumClearanceLine  that require GEOS 3.6, which would be cool to have in place (but people can get that in 2.3.1) ., Not worth the extra risk having to say we officially support PostgreSQL 9.6 on 2.2 if we can make 9.6 release time otherwise.

dbaston:
@robe2 agreed, `ST_MinimumClearance` is far from critic

dbaston:
@robe2, Agreed, having ST_MinimumClearance in PostGIS probably isn't enough reason to force  GEOS 3.6 out.

strk:
GEOS-3.6 is out now, with setPrecision support. Time to rebase this PR and merge for PostGIS 2.4 ?

robe2:
cough cough - are we ready to rebase and all that good stuff.

robe2:
cough cough @dbaston  are you going to do anything with this in 2.4.  If so clean this up and commit.

I want to release a beta1 after FOSS4G conference or you can work on it at code sprint.

dbaston:
@robe2 we can look at this at the code sprint, but it might not happen for 2.4.

robe2:
@dbaston  so still a lot of work to do on it.  Just wanted to get a sense of likeliness.

PostgreSQL 10 beta3 just came out.  I expect next they will release will be an rc1 in another 2 weeks or so.  At that point I'd like to be able to jump to release a little after their rc1 comes out.

Komzpa:
What work should be done? Is there something to help with? :)

dbaston:
@Komzpa beyond rebasing it, I'd want to be more confident that the design is sound.  In particular, I'd want to make sure it offers some kind of robustness advantage over manually snapping as @mwtoews demonstrated.  At the moment, I don't have a test project where I can look into that.

strk:
For testing difference between ST_SnapToGrid and this approach, compute this:
```
with inp as ( SELECT 
   'LINESTRING(0 0,10 10)'::geometry a,
   'LINESTRING(0 10,10 0)'::geometry b
)
SELECT 
  ST_AsText(ST_Intersection(a,b,10)) prec,
   ST_AsText(ST_Intersection(ST_SnapToGrid(a,10), ST_SnapToGrid(b,10))) prec
FROM inp
;
```

strk:
See https://lists.osgeo.org/pipermail/postgis-devel/2017-September/026456.html for discussion about upcoming new fully-robust algorithm in JTS which may be triggered by passing FixedPrecision geometries to GEOS algorithms.

liblwgeom/lwgeom_geos.c
```diff
@@ -549,18 +586,18 @@ lwgeom_intersection(const LWGEOM *geom1, const LWGEOM *geom2)
 
 	LWDEBUG(3, "intersection() START");
 
-	g1 = LWGEOM2GEOS(geom1, 0);
+	g1 = LWGEOM2GEOS_PREC(geom1, LW_FALSE, op_precision);
```
strk: We still want to be able to work with floating grid, how can it be done ?

dbaston: @strk By passing `NULL` for the `op_precision` argument

liblwgeom/lwgeom_geos.c
```diff
@@ -681,18 +718,26 @@ lwgeom_unaryunion(const LWGEOM *geom1)
...
 	if ( 0 == g1 )   /* exception thrown at construction */
 	{
 		lwerror("First argument geometry could not be converted to GEOS: %s", lwgeom_geos_errmsg);
 		return NULL ;
 	}
 
 	g3 = GEOSUnaryUnion(g1);
+	GEOSGeom_destroy(g1);
+
+	/* As of GEOS 3.6 r4118, GEOSUnaryUnion doesn't snap generated
```
strk: Was this filed upstream ? Worth having a link to the ticket, so to know when it is time to remove this check

liblwgeom/lwgeom_geos.h
```diff
@@ -38,6 +38,7 @@
 */
 LWGEOM *GEOS2LWGEOM(const GEOSGeometry *geom, char want3d);
 GEOSGeometry * LWGEOM2GEOS(const LWGEOM *g, int autofix);
+GEOSGeometry * LWGEOM2GEOS_PREC(const LWGEOM *g, int autofix, const double* precision);
```
strk: please document the semantic of the new parameter above the signature in header file (here)

strk: specify (after double-checking) that the geometry coordinates would possibly _move_ due to snapping to the precision grid.

strk:
Daniel do you still have the branch of this work ? I'd be nice to revive

strk:
It does not look like I can reopen this PR either, not sure why

dbaston:
Still accessible at https://github.com/dbaston/postgis/tree/geos-precision 
But I assume the plan now would be to use the new overlay engine in GEOS (when it is developed)

strk:
The new overlay engine in GEOS will automatically kick in when a fixed precision model is used, so having a way from PostGIS to request fixed precision would still be useful.

PR #101, opened 2016-03-01
Add dump-restore testing to travis

strk:
See https://trac.osgeo.org/postgis/ticket/3444

This is to check if it passes


strk:
r14726 = 54e0ef203fe78b5dc595c33adf6a9bbea2e07ef9 (refs/remotes/trunk)

PR #102, opened 2016-03-07
Add IRC badge

strk:
Just an idea, mostly in response to recent @mloskot mediated (unintended) advertisement of a proprietary chat service :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,14 +11,21 @@ GitLab-CI:
  [![Gitlab-CI](https://gitlab.com/postgis/postgis/badges/svn-trunk/build.svg)]
  (https://gitlab.com/ci/projects/3944?ref=svn-trunk)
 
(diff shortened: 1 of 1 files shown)


mloskot:
:+1:

strk:
Would be even better to drop the "kiwiirc.com" hosted client dependency, see https://trac.osgeo.org/osgeo/ticket/1639

(I guess I could just change the link to use `irc://` schema but I suspect most users going to Slack or Gitter won't have the browser correctly configured for that - correct me if I'm wrong)

mloskot:
I think most IRC users are tech savvy enough to get the `irc://` working. Those who are not, they may go to Gitter.

strk:
Pushed as r15024 = 307414769559ca936ed20f6af5b5d2c51455d88d (refs/remotes/trunk)

PR #105, opened 2016-06-15
[sketch] add skip for empty collection elements in LWGEOM2GEOS

Komzpa:
we can easily crash postgis with this query, this pullrequest sketches the solution:

```
with
        params as (
        select
            11 :: float as sidewalk_offset,
            1 :: float  as epsilon
    ),
        road as (
-- L-shaped road, 10 m
        select 'SRID=3857;LINESTRING(10 0, 0 0, 0 10)' :: geometry as geom
    ),
        sidewalks as (
        select ST_Collect(
                   ST_OffsetCurve(geom, sidewalk_offset),
                   ST_OffsetCurve(geom, -sidewalk_offset)
               ) geom
        from road, params
    )
select
    ST_Intersects(road.geom, sidewalks.geom),
-- should be false
    ST_Intersects(ST_Buffer(road.geom, sidewalk_offset + epsilon), sidewalks.geom) -- should be true
from road, sidewalks, params;
```

### DIFF
--- a/liblwgeom/lwgeom_geos.c
+++ b/liblwgeom/lwgeom_geos.c
@@ -332,7 +332,7 @@ LWGEOM2GEOS(const LWGEOM *lwgeom, int autofix)
 	/*
 	LWGEOM *tmp;
 	*/
-	uint32_t ngeoms, i;
+	uint32_t ngeoms, i, j;
 	int geostype;
 #if LWDEBUG_LEVEL >= 4
 	char *wkt;
@@ -443,9 +443,11 @@ LWGEOM2GEOS(const LWGEOM *lwgeom, int autofix)
 		ngeoms = lwc->ngeoms;
 		if ( ngeoms > 0 )
 			geoms = malloc(sizeof(GEOSGeom)*ngeoms);
-
+		j = 0;
+		// i is index of lwgeom geometry, j is index of geos geometry - skip empty collection entries
 		for (i=0; i<ngeoms; ++i)
 		{
+			if (lwgeom_is_empty(lwgeom)) continue; // skip empty
 			GEOSGeometry* g = LWGEOM2GEOS(lwc->geoms[i], 0);
 			if ( ! g )
 			{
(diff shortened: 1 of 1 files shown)


liblwgeom/lwgeom_geos.c
```diff
@@ -443,19 +443,21 @@ LWGEOM2GEOS(const LWGEOM *lwgeom, int autofix)
 		ngeoms = lwc->ngeoms;
 		if ( ngeoms > 0 )
 			geoms = malloc(sizeof(GEOSGeom)*ngeoms);
-
+		j = 0;
+		// i is index of lwgeom geometry, j is index of geos geometry - skip empty collection entries
 		for (i=0; i<ngeoms; ++i)
 		{
+			if (lwgeom_is_empty(lwgeom)) continue; // skip empty
```
pramsey: To skip empty sub-geometries, shouldn't this be
```c
if(lwgeom_is_empty(lwc->geoms[i]))
```
right now this is just skipping only in the case where the whole collection is empty, which I'm surprised isn't handled further up the chain

liblwgeom/lwgeom_geos.c
```diff
@@ -443,19 +443,21 @@ LWGEOM2GEOS(const LWGEOM *lwgeom, int autofix)
...
 			GEOSGeometry* g = LWGEOM2GEOS(lwc->geoms[i], 0);
 			if ( ! g )
 			{
 				while (i) GEOSGeom_destroy(geoms[--i]);
 				free(geoms);
 				return NULL;
 			}
-			geoms[i] = g;
+			geoms[++j] = g;
```
pramsey: Won't this leave `geoms[0]` null? You want to read-then-increment (`j++`) not increment-then-read (`++j`).

pramsey:
Have opened an osgeo ticket [3579](https://trac.osgeo.org/postgis/ticket/3579) for this report.

pramsey:
Ticket is now closed out on osgeo.

PR #108, opened 2016-07-18
ST_AsGeobuf implementation (wip)

bjornharrtell:
This is initial code for ST_AsGeobuf. See https://github.com/mapbox/geobuf for more information about the format.

ST_AsGeobuf takes two parameters, a query string and geometry column name and will produce a full FeatureCollection Geobuf with other columns as properties.

External library protobuf-c along with generated code from it's compiler is used to encode and serialize the binary Geobuf data. 

This is my first attempt at PostgreSQL/PostGIS C code and ages since I did anything in C or C++ so please go easy on me. :)

### TODO

* [ ] Translate column types into supported Geobuf value types
* [x] Handle multigeometry types
* [ ] Handle other than two dimensions
* [ ] Handle other than default precision
* [ ] Add protobuf-c as a proper dependency
* [ ] Make this and dependency on protobuf-c optional via configure
* [ ] Don't overallocate
* [ ] Deallocations where appropriate
* [ ] Use an internal structure to pass encoding context


bjornharrtell:
Superseded by https://git.osgeo.org/gogs/postgis/postgis/pulls/4.

danny-wang:
ST_asgeobuf function doesn’t have “combine” function defined, so it cannot be pushed down to worker when i use cluster postgres. Do u have plan to support combine function?

bjornharrtell:
@danny-wang I have no such plans.

pramsey:
Same, no plans at this time.

danny-wang:
now ST_asgeobuf is a cpu-busy job, it only use one cpu, if you define combine function, i think it can use multi-cpu, the can reduce the time-cost?

PR #110, opened 2016-07-31
EncodedPolylines: precision is second parameter... not third !

cquest:
precision parameter had no effect on ST_AsEncodedPolyline and st_linefromencodedpolyline as code was looking for it as the third parameter instead of the second.

regress tests need to be updated... (not enough familiar with that, sorry)


dbaston:
I made a ticket for this on trac: https://trac.osgeo.org/postgis/ticket/3627

In the commit diff (27c11b26dc14225b2c9ef46ef97f5aae07362a08) you can see the changes I made to fix the tests.  There's not a whole lot to it - one file with SQL statements, and one file with expected results of those statements.

cquest:
Thanks for doing the link with trac...

robe2:
I'm closing this since the related ticket is closed and looks patched to me.

PR #112, opened 2016-08-05
Fix #3607 bug. Inconsistency with multilinestring

select-artur:
Hello!

This patch fixes #3607 ticket. After this fix the query give the following result:

```sql
=> SELECT ST_AsText(ST_LocateBetweenElevations('MULTILINESTRING Z((0 0 0, 10 10 10))'::geometry, 11, 11));
        st_astext        
-------------------------
 MULTILINESTRING Z EMPTY
(1 row)
```

I suppose ST_LocateBetweenElevations() becomes consistent after it. After viewing code I don't see any drawbacks.


strk:
Committed in trunk and backported, see https://trac.osgeo.org/postgis/ticket/3607 for details. Thanks!

PR #115, opened 2016-08-25
changes needed for gitattributes

crunchyjohn:
Hello all,

I see something weird in PostGIS at the moment.
When I do:
mkdir ~/git
cd ~/git
git clone git@github.com:postgis/postgis.git
cd postgis
git status

I notice the following:
	modified:   extras/tiger_geocoder/create_geocode.bat
	modified:   extras/tiger_geocoder/upgrade_geocoder.bat

Although, I didn't modify these files.  I just did a regular clone.  

I suspect the .gitattributes file has something to do with it.
The 2 files impacted are Windows files, so probably have the wrong carriage-return type.
And, in the .gitattributes file, I see this rule:
*.bat       eol=crlf
So, I think this forces a change after clone.

I suspect this problem was introduced in the latest commit from 8/17 (it seems the endline changes were committed there).

I've submitted the patch with the 2 files it wants to change.


strk:
See d13e59745a0d9f903732ed9211f5345d9b904d2f

PR #117, opened 2016-09-29
create aggregate parallel safe

gsmol:
Hello!
For parallel aggregation to work, you need to explicitly define Parallel Mode in aggregate definition.
Unfortunately, parallel mode parameter cannot be used with old-style aggregate definition that uses base_type parameter.
Based on:
https://www.postgresql.org/message-id/b1e247ca-23ad-1d24-cecc-28600550f7bb@postgrespro.ru

### DIFF
--- a/postgis/postgis.sql.in
+++ b/postgis/postgis.sql.in
@@ -3742,23 +3742,23 @@ CREATE OR REPLACE FUNCTION ST_Combine_BBox(box2d,geometry)
 
 
 -- Availability: 1.2.2
-CREATE AGGREGATE ST_Extent(
+CREATE AGGREGATE ST_Extent(geometry) (
 	sfunc = ST_CombineBBox,
 #if POSTGIS_PGSQL_VERSION >= 96
   combinefunc = ST_CombineBBox,
+  parallel = safe,
 #endif
 	finalfunc = box2d,
-	basetype = geometry,
 	stype = box3d
 	);
 
 -- Availability: 2.0.0
-CREATE AGGREGATE ST_3DExtent(
+CREATE AGGREGATE ST_3DExtent(geometry) (
 	sfunc = ST_CombineBBox,
 #if POSTGIS_PGSQL_VERSION >= 96
   combinefunc = ST_CombineBBox,
+  parallel = safe,
 #endif
-	basetype = geometry,
 	stype = box3d
 	);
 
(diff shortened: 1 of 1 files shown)


pramsey:
Yeah, no, this isn't going to work for us. Most of our aggregates accumulate to an array in a side memory context, and then do all the work in the finalfunc, so the standard aggregate parallelism in PgSQL 9.6 doesn't nothing for us. The only real exception is MemGeomUnion, which does to work in each transfer call. I don't think this patch helps.

Komzpa:
@pramsey right we hit two problems:
 - a single non-parallel-safe function in query disables parallelism for all the query, even if there's a parallelizable subquery - the functions that are parallel safe should be marked as such, even if they are parallel useless;
 - we're analyzing a large "log", grouping geometries by some scattered non-indexable ID. combining them in 8 workers seems to gain benefit when we're using array_agg, because of parallel disk reads that can utilize their performance - would be much nicer to be able to do this with parallelizable ST_Collect (or even ST_Extent, for the purpuose) rather than parallelizable array_agg.

strk:
As you're changing aggregate signatures (to add a parameter type) I guess you'll have to add those aggregates to the postgis_drop_before.sql, for proper upgrade

strk:
Note that travis doesn't test upgrades, so the green there isn't useful for this (long time on the todo list).
You should check upgrades from older versions (2.0, 2.1, 2.2, 2.3).

gsmol:
True, I didn`t taken upgrade issue into account.
I will look into it, thanks

strk:
It might be enough to add a "Changed" commend right below
the "Availability" one, like:

-- Changed: 2.4.0 changed to use new aggregate syntax and be parallel-safe when supported by PostgreSQL version

strk:
Consider pushing your code to a fork on the OSGeo git service:
https://git.osgeo.org/gogs/postgis/postgis/

Doing so you could enable your fork to be guarded by Drone:
https://drone.osgeo.kbt.io
and in turn doing that would automatically check the branch for upgrades behavior (it would still need you to provide some test expected to fail with a non-upgraded aggregate)

strk:
Another thing: parameter for aggregates are only supported from PostgreSQL 9.4 upward, so this cannot be accepted w/out a conditional

Komzpa:
@strk 9.4 adds support for named parameters, not parameters (they are in place since 8.2)

robe2:
I only committed the patch part for the non-transition functions.  The transition ones I think we can reconsider in 2.4.  Since we can't ever parallelize them at this point, they if marked should be marked something else like RESTRICTED.

https://www.postgresql.org/docs/9.6/static/parallel-safety.html

But things may change in PostgreSQL 10, so I'm leaving that open for discussion (we may at least be able to mark them RESTRICTED in 2.3.2) and conditional in 2.4.  For now, I'm going to close this out.  

refer to https://trac.osgeo.org/postgis/ticket/3650  for details

Komzpa:
@robe2 having the functions not marked as parallel safe means that parallel execution gets completely disabled for them. Thus, with your trimmed patch using ST_Extent will disable parallelism for all the query, even though ST_Extent may be used in context that does not require parallelism for itself.

robe2:
Yah I was also concerned with those since they use GEOS so wasn't sure how the whole threading thing would work.  So I really didn't want to push it into a micro without more testing.  ST_Extent was okay because it's a simple aggregate and it has no extra dependencies beyond PostGIS.

Komzpa:
Parallel execution is implemented via multiprocessing but not threading, thus thread safety does not change anything for it.
We're running with above patch since September and did not run into issues.

PR #119, opened 2016-11-09
Fix sphere_direction refs #3667

mhugo:


robe2:
Thanks for the quick turn around.  Committed to both trunk and 2.3 branch

PR #120, opened 2016-12-07
Add discrete frechet distance

ssugiyama:
I made a patch. Please review it. 

See https://trac.osgeo.org/postgis/ticket/3677


doc/reference_measure.xml
```diff
@@ -2598,6 +2598,91 @@ The current implementation supports only vertices as the discrete locations. Thi
...
+                <para><xref linkend="ST_FrechetDistance" /></para>
+          </refsection>
+	</refentry>
+
+	<refentry id="ST_FrechetDistance">
+	  <refnamediv>
+		<refname>ST_FrechetDistance</refname>
+
+		<refpurpose>Returns the Fréchet distance between two geometries. This is a measure of similarity between curves that takes into account the location
+		and ordering of the points along the curves.  Units are in the units of the spatial reference system of the geometries.</refpurpose>
```
strk: Can you add a reference of the algorithm specification, and an explanation of how this is different from Hausdorff ?

strk: Sorry, found the reference below, that's ok.

strk:
How about a single SQL definition, with a default 3rd argument ?

ssugiyama:
Well, that seems to be practical. I did so. 

What should I do next?

strk:
Push your change (current "Changed Files" tab only shows a single signature but with no "densify" optional argument). Then make sure Travis build succeeds.

robe2:
This does not appear to be well guarded for lower versions.

If I compiled against GEOS 3.6.1 I get a compile error:

``

lwgeom_geos.o:lwgeom_geos.c:(.text+0x6a3): undefined reference to `GEOSFrechetDistance'

``

Also not clear why you removed the function with the 3rd proto (densfrac), What @strk meant is you keep that one but mark the last argument as optional.

Take a look at ST_Subdivide as an example that does this:

http://postgis.net/docs/manual-dev/ST_Subdivide.html


Looking at our docu it seems people are not correctly flagging optional args in documentation so took me sometime to find a good example you can follow.

robe2:
More suggestions

1) rename PG_FUNCTION from frechetdistance to  ST_FrechetDistance.  This is our new convention.  You were probably copying something like HausdorffDistance. which follows our old convention.

2) As noted earlier, your function is not guarded, you need to wrap your datum func into something like

> #if POSTGIS_GEOS_VERSION < 37
	lwpgerror("The GEOS version this PostGIS binary "
	        "was compiled against (%d) doesn't support "
	        "'ST_FrechetDistance' function (3.7.0+ required)",
	        POSTGIS_GEOS_VERSION);
	PG_RETURN_NULL();
#else /* POSTGIS_GEOS_VERSION >= 37 */


 #if 

3) Bring back your densifrac version, but have them handled by a single ST_FrechetDistance datum and SQL signature.  I see they are handled by different GEOS functions, so I guess you could maybe set optional arg to -1 and have the code check if -1 is passed in, go thru the regular GEOSFrechetDistance  and if some other value is passed in, use the GEOSFrechetDistanceDensify.

I was thinking of NULL as a value, but I think it would be cleaner to use a non NULL value for 3rd arg, so you can keep the function marked as STRICT to save cycles.

ssugiyama:
Thank you for your suggestions. They help me a lot. I think I followed all of them. Please check them.

robe2:
@ssugiyama  looks good.  I'm making a minor adjustment to the regress test, cause you double-included regress_buffer_params test when it doesn't need to be in that loop.  Aside from that looks good.  I'll commit shortly.

ssugiyama:
@robe2 Oh, thank you so much for your kindness!

PR #122, opened 2017-01-22
Add orientation checking and forcing functions

dbaston:
This PR adds four functions: `ST_ForceCW`, `ST_ForceCCW`, `ST_IsCW`, and `ST_IsCCW`, following  discussions in Trac tickets [#3689](https://trac.osgeo.org/postgis/ticket/3689) and  [#3651](https://trac.osgeo.org/postgis/ticket/3651).

The `ST_IsCW` and `ST_IsCCW` functions are proposed as an alternative to porting `ST_Orientation` to liblwgeom.


strk:
I don't like returning `true` for non-polygonal components (as written in the document).
What if, for example, the input is a closed linestring ?
Any unsupported input should return either NULL or raise an exception, IMHO.

Komzpa:
please don't do exceptions, it's pain later on and makes one write many wrappers.

dbaston:
I was also thinking about the closed LineString case.  But I thought it was more important to have the results of the checking functions be consistent with the forcing functions.  `ST_ForceCCW` will pass a point straight through, so `ST_IsCCW` should return true for points.  Similarly, `ST_ForceCCW` will pass a closed LineString straight through without regard to its orientation, so it seemed confusing to implement different behavior for `ST_IsCCW`.

strk:
How about `ST_ForceCW` ? Will it also pass a closed LineString untouched
regardless of orientation ?

And then we'd allow for `ST_IsCW(x) AND ST_IsCCW(x)` to be true ?

Should probably be discussed on the dev mailing list, rather than here...

dbaston:
Also worth thinking about the case of a GeometryCollection with a Polygon and a Point.  What should `ST_IsCCW` do here if it cannot process points?

dbaston:
> And then we'd allow for `ST_IsCW(x) AND ST_IsCCW(x)` to be true ?

Yes.  Similarly, for a polygon with mixed orientation, both `ST_IsCW(x)` and `ST_IsCCW(x)` would return false.

PR #123, opened 2017-02-07
Add check for pkg-config

bjornharrtell:
Since fee914f1174c1ab6869fbc9e23634fcd68a8455c pkg-config became required for configuration, with this it should become optional.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -897,6 +897,15 @@ AC_SUBST([HAVE_JSON])
 fi dnl }
 
 
+dnl ===========================================================================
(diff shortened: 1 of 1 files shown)


strk:
I'd avoid to abort/error out on missing pkg-config, maybe just AC_WARN

bjornharrtell:
Ah yes that was a mistake. I'll close this one to return on Gogs.

robe2:
related ticket - https://trac.osgeo.org/postgis/ticket/3707

PR #125, opened 2017-02-13
Fix mishandling of spaces in "configure --with-pgconfig" path

laurenz:
It is quite common to have spaces in installation path names
on Windows, but this will cause trouble with the --with-pgconfig
configure option.
Fix by adding double quotes throughout.

This was brought to my attention by [this stackoverflow question](http://stackoverflow.com/q/42093865/6464308).


strk:
Thanks, committed to trunk:

r15313 = 26a9cf74e09521476ae272ced8970018e439eb8f (refs/remotes/trunk)

Please close this yourself if you can (I'd have to leave the
terminal to do that)

laurenz:
Thank you.

PR #126, opened 2017-03-28
Fix documentation of centroid examples.

jackieleng:
I've noticed that in the ST_centroid example images (http://postgis.net/docs/ST_Centroid.html) the centroid dot is green instead of blue. I've fixed the documentation so it now says green. I hope this helps.


strk:
Pushed as r15343 = d8db168ce9817b51ff76b3d230377265b05f246f
po updates done in a second commit with `make update-pot` (r15344 = b0a3a90ce8be2cfadaa306be8a34b0ef9e0edce1)

PR #127, opened 2017-03-31
Recursive subdivide fixes

Komzpa:
- don't lose geometry on deep subdivision;
 - don't increment depth twice if we need to subdivide both left and right sides.

### DIFF
--- a/liblwgeom/lwgeom.c
+++ b/liblwgeom/lwgeom.c
@@ -1979,11 +1979,11 @@ lwgeom_subdivide_recursive(const LWGEOM *geom, int maxvertices, int depth, LWCOL
 		return n;
 	}
 	
-	/* But don't go too far. 2^25 = 33M, that's enough subdivision */
-	/* Signal callers above that we depth'ed out with a negative */
-	/* return value */
+	/* But don't go too far. 2^50 ~= 10^15, that's enough subdivision */
+	/* Just add what's left */
 	if ( depth > maxdepth )
 	{
+		lwcollection_add_lwgeom(col, lwgeom_clone_deep(geom));
 		return 0;
 	}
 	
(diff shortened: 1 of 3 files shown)


strk:
Please add a testcase showing why this is needed

Komzpa:
Here's example of asymetricity in data loss:
```
with holed_geom as (
    select ST_Difference(
               ST_Expand('SRID=3857;POINT(0 0)' :: geometry, 2 ^ 30),
               ST_Buffer(
                   'SRID=3857;POINT(0 0)' :: geometry,
                   1,
                   1000)
           ) as geom
)
select ST_AsText(
    ST_SymDifference(
        ST_Union(geom),
        (
            select geom
            from holed_geom
        )
    )
)
from (
         select ST_Subdivide(geom, 10) geom
         from holed_geom
     ) z;
```

Komzpa:
Here's how the lost part of geometry looks like:
image

Komzpa:
This case is specifically visible when you're trying to get 'inverted' geometry by subtracting a tableful of data from infinity bounding box, and then subdivide to get data you can manage.

strk:
I'm talking about automated testing, see regress/subdivide.sql
and regress/subdivide_expected

strk:
You need to add the new regression test file in regress/Makefile.in

Komzpa:
@strk this file is there alredy, just for newer GEOS only - it's not running on Travis.

robe2:
I'm testing now.  I'll commit to 2.4 and then 2.3 if it passes my 9.6 testing.

robe2:
Failed my windows testing for a really stupid reason. I get answer back in scientific notation.

I'm changing the last example slightly so hopefully it's consistent across all platforms.

`https://github.com/postgis/postgis/pull/127|1.6e+15`

strk:
Please also drop the github pull reqeust referenc and substitute
with a trac ticket number reference (to be opened if not done
so yet).

Regina: also before committing to SVN please squash-rebase
and include the author name in the commit log (I didn't know
you were all setup to merge from git to SVN..)

robe2:
Okay committed with minor change to the ticket test so doesn't give scientific notation under windows.

robe2:
@strk sorry I missed your last note before I committed.  Guess it's too late for all of that.

PR #128, opened 2017-04-25
Make regular quotes within programlisting tags for PDF manual

mwtoews:
Fixes [#3726](https://trac.osgeo.org/postgis/ticket/3726)


robe2:
Mike,

Thanks for the patch.  PDF is much easier to cut and paste from.

Committed change to both 2.3 and trunk

PR #131, opened 2017-05-15
Add note to ST_Equals about not supporting geometry collections

lb1c:
Closing the loop on an old fork of postgis. We never did the PR about our docs change.

Since the current docs https://postgis.net/docs/ST_Equals.html don't mention this limitation i open a PR about it. If this has been addressed otherwise without my knowledge, please feel free to reject and close.


robe2:
Committed the change at r15395 - https://trac.osgeo.org/postgis/changeset/15395

PR #134, opened 2017-08-10
Troubleshooting issue with travis

robe2:


strk:
There's your problem:
```
drwxr-x--- 33 travis travis 4096 Aug 10 08:27 travis
```

Basically, the PostgreSQL server user (postgres?) cannot access travis's home.
Try adding this step:
```
chmod 755 ~
```

I would not upgrade PostgreSQL in the same PR, better keep focused :)

strk:
> ERROR:  could not access file "/home/travis/build/postgis/postgis/regress/00-regress-install/lib/postgis-2.3": Permission denied

You changed permissions too late, they are needed *before* `make check` (which is the pre-install check, thus needing to access travis directory).

Also, just changing /home/travis permission should be fine, no need to descend (as you're trying..)

robe2:
@strk is it going to cause a problem when we mirror that I just inline edited and the patch I created is a branch of this repo.  I guess we can test if you are not sure or I can delete this patch branch and pull request before I commit.

robe2:
Fixed at svn r3803.

strk:
it looks like it didn't create a problem, but please delete the branch now that you don't need it anymore

PR #136, opened 2017-08-14
Use GEOSNode instead of UnaryUnion for ST_Node; fixes #3647

Wassasin:
As per https://trac.osgeo.org/postgis/ticket/3647.
Probably the same solution as tested by @strk. Changes the results of the regression tests for ST_Node, but seems not to conflict with the documentation.


strk:
Make sure to only use GEOSNode when really available. It depends on GEOS version.  Also you saw how results change so all tests should also be switched on GEOS version and documentation should explain the difference

dbaston:
`GEOSNode` is available in GEOS 3.4+, which has been available for over 4 years.  Is GEOS 3.4 not a requirement for current versions of PostGIS?

robe2:
Right now 3.3.0 or higher is required for PostGIS 2.4 and PostGIS 2.3.

I'd like to bump the requirement for PostGIS 2.4 to at least GEOS 3.4.  It's too late to do that for 2.3.

robe2:
oh I forgot we also agreed on sprint to bump to 3.5

https://trac.osgeo.org/postgis/ticket/3810

But I asked on mailing list just in case anyone has a serious objection:

https://lists.osgeo.org/pipermail/postgis-devel/2017-August/026302.html

robe2:
@pramsey  are you around?  I wanted to accept this pull request, but I'd have to change the configure.ac to do so to make 3.5 the minimum and I think you are still working on it.  If you haven't started yet, I can do that, but didn't want to disrupt your surgery.

pramsey:
Go ahead

liblwgeom/lwgeom_geos_node.c
```diff
@@ -146,106 +109,28 @@ lwgeom_node(const LWGEOM* lwgeom_in)
...
+	if ( ! gn ) {
+		lwerror("GEOSNode: %s", lwgeom_geos_errmsg);
 		return NULL;
 	}
 
 	/* Linemerge (in case of overlaps) */
-	gm = GEOSLineMerge(gu);
-	GEOSGeom_destroy(gu);
+	gm = GEOSLineMerge(gn);
+	GEOSGeom_destroy(gn);
```
strk: please check if this is still needed, as GEOSNode might be taking care of line-merging already

regress/node_expected
```diff
@@ -1,3 +1,3 @@
 t1|SRID=10;MULTILINESTRING((0 0,5 0),(5 -5,5 0),(5 0,10 0),(5 0,5 5))
-t2|SRID=10;MULTILINESTRING((0 0,8 0),(8 0,10 0,15 0,20 0),(20 0,25 0))
+t2|SRID=10;LINESTRING(0 0,8 0,10 0,15 0,20 0,25 0)
```
strk: the expectance is that original nodes are retained, please don't change that (maybe it should be documented)

robe2:
@Wassasin can you make the changes requested by @strk and then I think we are good to committ.

If you can do so before Monday that would be great.

strk:
Had you tried just dropping the LineMerge step ? (I did  not)

Wassasin:
It seems that GEOSNode does not perform LineMerge:

```
< MULTILINESTRING((0 0,2.5 2.5),(0 5,2.5 2.5),(2.5 2.5,5 5,10 0),(10 0,11 0,12 0,20 0),(20 0,22 0),(2.5 2.5,5 0))
---
> MULTILINESTRING((0 0,2.5 2.5),(2.5 2.5,5 5,10 0),(10 0,11 0),(11 0,12 0),(12 0,20 0),(22 0,20 0),(0 5,2.5 2.5),(2.5 2.5,5 0))
```
My changes have been reverted, and now the only change is that GEOSNode is applied in stead of GEOSUnaryUnion. The preservation of nodes has been restored. In this case the tests need not be changed, except for the one cu_node.c case where only the lines change order and direction.

I was thinking of adding a note to the docs mentioning the changed behavior in the next major Postgis version. The docs indeed already mention that all input nodes are preserved.

strk:
thanks for the care

robe2:
So I'll commit today.  Probably in a couple of hours.  I assume that is okay.

strk:
Please don't mix github and trac ticket references. Github ones need always be full URLs

robe2:
I did put in a full url for github.  That's why the commit reference shows on this ticket.  But seems I think it expects the first close line to be for it otherwise it doesn't automatically close.

strk:
I see a vare reference in https://github.com/postgis/postgis/commit/b0ecd884558142002c6c395edd40bcae514a6444 - is github omitting the full URL from rendering ?

robe2:
Yes it is and converting it to a hyperlink -- here is my commit on svn - https://trac.osgeo.org/postgis/changeset/15571

strk:
Stupid GitHub. We cannot even fix it !

PR #139, opened 2017-08-29
return 1 in ST_Subdivide

Komzpa:
We have a large geometry in which ST_Subdivide loses a box.
Re-reading the code found this issue: we add into collection but return length as 0, but not 1.


strk:
Can you please also provide a regression test ?

Komzpa:
After another re-reading, `n` is unused later, thus this patch does not fix or break anything.

The case I have is big, I'll try to chop it down.

pramsey:
I'm OK w/ the change, it doesn't affect anything, and wouldn't have any effect on a missing box either. Am I right that this is an issue you flagged not as a fix for your problem but just something you noticed while doing a code review?

Komzpa:
Yes, this is not a fix for my actual problem, and I don't have a reproducer that I can share, for now.

pramsey:
OK, it's tiny, I've applied it to trunk. Thanks and I'm closing!

PR #144, opened 2017-10-02
Unions to avoid memcpy in sortable hash generation

Komzpa:


pramsey:
Hokay, I was a little worried that this leans on compiler behaviour, but it seems to be part of the C standard, so ... cool. Nice work.

strk:
looks good to me

pramsey:
Super, committed via https://trac.osgeo.org/postgis/ticket/3871

PR #146, opened 2017-10-03
Hot path for sorting planar points in btree operator class

Komzpa:
Removing everything that slows down btree sorting to get to maximum speed.

### DIFF
--- a/liblwgeom/g_serialized.c
+++ b/liblwgeom/g_serialized.c
@@ -19,6 +19,7 @@
  **********************************************************************
  *
  * Copyright 2009 Paul Ramsey <pramsey@cleverelephant.ca>
+ * Copyright 2017 Darafei Praliaskouski <me@komzpa.net>
  *
  **********************************************************************/
 
@@ -109,7 +110,7 @@ int32_t gserialized_get_srid(const GSERIALIZED *s)
 	if ( srid == 0 )
 		return SRID_UNKNOWN;
 	else
-		return clamp_srid(srid);
+		return srid;
 }
 
 void gserialized_set_srid(GSERIALIZED *s, int32_t srid)
@@ -126,6 +127,15 @@ void gserialized_set_srid(GSERIALIZED *s, int32_t srid)
 	s->srid[0] = (srid & 0x001F0000) >> 16;
 	s->srid[1] = (srid & 0x0000FF00) >> 8;
 	s->srid[2] = (srid & 0x000000FF);
+}
+
+inline static int gserialized_cmp_srid(const GSERIALIZED *s1, const GSERIALIZED *s2)
+{
+	return (
+		s1->srid[0] == s2->srid[0] &&
+		s1->srid[1] == s2->srid[1] &&
+		s1->srid[2] == s2->srid[2]
+	) ? 0 : 1;
 }
 
 GSERIALIZED* gserialized_copy(const GSERIALIZED *g)
@@ -240,14 +250,14 @@ static uint64_t uint32_interleave_2(uint32_t u1, uint32_t u2)
 }
 #endif
 
+union floatuint {
+	uint32_t u;
+	float f;
+};
 
 uint64_t gbox_get_sortable_hash(const GBOX *g)
 {
-
-	union floatuint {
-		uint32_t u;
-		float f;
-	} x, y;
+	union floatuint x, y;
 
 	/*
 	* Since in theory the bitwise representation of an IEEE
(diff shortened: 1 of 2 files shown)


pramsey:
Bridge too far....

Komzpa:
This doesn't break SQL API.
https://trac.osgeo.org/postgis/ticket/3864

Further major improvements are possible, but require using sortsupport infrastructure.

pramsey:
Can you point to sortsupport examples? I'm interested to learn more.

Komzpa:
In docs it's called "Abbreviated keys".
Here is a structure to fill: https://github.com/postgres/postgres/blob/master/src/include/utils/sortsupport.h
Examples:
 - https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/uuid.c#L25
 - https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/varlena.c#L1382 - a big chunk of code that explains why Collate C text ordering is much faster than usual - thus why geohash sorting is faster.

pramsey:
Not seeing any references to this in the SQL guides of PgSQL, I assume we'd have to hand enter our hooks into pg_amproc, which would be a return to old form back in the early 2000s when we had to hand enter our GIST index bindings, before the advent of CREATE OPERATOR CLASS, and friends.

Komzpa:
@pramsey according to @funny-falcon it's doable in SQL:

function 2: https://www.postgresql.org/docs/devel/static/xindex.html#xindex-btree-support-table
example code: https://www.postgresql.org/docs/devel/static/xindex.html#xindex-opfamily

pramsey:
Yep, just another line in the opclass, nice.

strk:
> Yep, just another line in the opclass, nice.

Let's see if we can make it upgradeable.
IIRC I added some support for per-opclass-line upgrades
("-- Availability:" comment supported on each line)

Komzpa:
@pramsey in IRC @robe2 said that you're the one to review this. 
I propose the following:
 - for 2.4.x, commit this;
 - for 2.5.x, do a bigger change. Add sortsupport, and discuss whether the sort needs to be on box center, or just the first point of geometry is fine too. For latter case, we can skip box construction for simple geometries that don't have it serialized, without affecting the order too much.

liblwgeom/g_serialized.c
```diff
@@ -109,7 +110,7 @@ int32_t gserialized_get_srid(const GSERIALIZED *s)
 	if ( srid == 0 )
 		return SRID_UNKNOWN;
 	else
-		return clamp_srid(srid);
+		return srid;
```
strk: This is unrelated to "clean up sorting code path", isn't it ?
Could you please rebase and make sure only related changes are seen in the "Files changed" tab on github ?

Komzpa: gserialized_get_srid is one of topmost not-inlined functions in sort:
image

and it can't be inlined since it calls clamp_srid:
image

Komzpa: This optimization was really walked around later, as I made srid cmp function that doesn't construct SRID. But it also doesn't clamp it, which makes `cmp_srid(a,b)` and `get_srid(a)==get_srid(b)` diverge in opinions in edge cases. I think if we can't commit to non-clamping get_srid, we can't commit to non-clamping cmp_srid, so these both should go in pair.

Since we are clamping on write, without data corruption you shouldn't get non-clamped values on disk.

strk: I think the clamp-on-read was done for the restore case, when restoring a dump from pre-clamp dates (postgis_restore should take care of that, but pg_upgrade would not). Anyway this PR is about "clean up", not about "performance improvement", am I wrong ?

Komzpa: Sorry for misleading wording. It is about performance, and addresses https://trac.osgeo.org/postgis/ticket/3864.

According to @robe2 pre-2.0 dumps come in HEXWKB, and for those clamping is active: https://github.com/postgis/postgis/blob/a5c7920a059c7273864760539b6845d74b34289c/liblwgeom/lwin_wkb.c#L701

pramsey:
Committed. Further use of sortsupport fanciness should go in a fresh ticket/PR

liblwgeom/g_serialized.c
```diff
@@ -284,8 +294,43 @@ int gserialized_cmp(const GSERIALIZED *g1, const GSERIALIZED *g2)
...
 	size_t sz2 = SIZE_GET(g2->size);
+	union floatuint x, y;
+
+	/*
+	* For two non-same planar points, we can skip a lot of machinery.
+	*/
+	if (
+		sz1 > 16 && // 16 is size of EMPTY, if it's larger - it has coordinates
+		sz2 > 16 &&
+		*(uint32_t*)(g1->data) == POINTTYPE &&
```
notorca: The C Standard, 6.3.2.3, paragraph 7 [ISO/IEC 9899:2011], states
A pointer to an object or incomplete type may be converted to a pointer to a different object or incomplete type. If the resulting pointer is not correctly aligned for the referenced type, the behavior is undefined.

liblwgeom/g_serialized.c
```diff
@@ -284,8 +294,43 @@ int gserialized_cmp(const GSERIALIZED *g1, const GSERIALIZED *g2)
...
+		*(uint32_t*)(g2->data) == POINTTYPE &&
```
notorca: Same ussue here

liblwgeom/g_serialized.c
```diff
@@ -284,8 +294,43 @@ int gserialized_cmp(const GSERIALIZED *g1, const GSERIALIZED *g2)
...
+		!FLAGS_GET_BBOX(g1->flags) &&
+		!FLAGS_GET_GEODETIC(g1->flags) &&
+		!FLAGS_GET_BBOX(g2->flags) &&
+		!FLAGS_GET_GEODETIC(g2->flags)
+	)
+	{
+		double *dptr = (double*)(g1->data);
```
notorca: and here

liblwgeom/g_serialized.c
```diff
@@ -284,8 +294,43 @@ int gserialized_cmp(const GSERIALIZED *g1, const GSERIALIZED *g2)
...
+		x.f = 2.0 * dptr[1];
+		y.f = 2.0 * dptr[2];
+		hash1 = uint32_interleave_2(x.u, y.u);
+
+		dptr = (double*)(g2->data);
```
notorca: one more

notorca:
https://www.securecoding.cert.org/confluence/display/c/EXP36-C.+Do+not+cast+pointers+into+more+strictly+aligned+pointer+types

liblwgeom/g_serialized.c
pramsey: The data buffer is constructed carefully so that aligned access is expected. Doubles exist on double alignment boundaries, and ints on their alignment boundary. https://github.com/postgis/postgis/blob/svn-trunk/liblwgeom/g_serialized.txt
Also, the storage type in pgSQL is defined as double aligned, so the segments we're getting back are on double-aligned boundares.
I'm not sure whether "it works" is sufficient for this objection.
One possible issue I guess is for geometries built on top of memory they allocated themselves, whether palloc() (when running within pgsql) or malloc() are returning double-aligned memory.

notorca: There is no guarantee, than result of malloc have same alignment as double.

notorca: And there is one more problem here, *(uint32_t*)(g1->data) and double *dptr = (double*)(g1->data); creates 2 aliases for same memory address with incompatible types.

liblwgeom/g_serialized.c
notorca: Should be:
```
dptr = (double*)(g2->data+sizeof(double));
x.f = 2.0 * dptr[0];
y.f = 2.0 * dptr[1];
```

PR #147, opened 2017-10-04
Fix division by zero in lw_dist2d_pt_arc

Algunenano:
In `lw_dist2d_pt_arc`, if the point is in the center of the arc a division by zero happens which leads to undefined behaviour.

The PR solves this by retuning the radius as distance an setting the first point (A1) as the closest (all the points in the arc are at the same distance from P).


Algunenano:
Trac issue in https://trac.osgeo.org/postgis/ticket/3874

I don't really mind about the authorship thing in here, but I'll do it in any other PR

pramsey:
Committed

PR #148, opened 2017-10-04
geohash_point_as_int: Fix undefined behaviour in shift operation

Algunenano:
Fixing undefined behaviour in test_geohash_point_as_int:
```
Running test 'test_geohash_point_as_int' in suite 'computational_geometry'.
lwalgorithm.c:669:18: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
```

0x0001 is a signed int so shifting it 31 bits is undefined. 0x00001u is unsigned so the bit shift is ok.

### DIFF
--- a/liblwgeom/lwalgorithm.c
+++ b/liblwgeom/lwalgorithm.c
@@ -680,7 +680,7 @@ unsigned int geohash_point_as_int(POINT2D *pt)
 			mid = (lon[0] + lon[1]) / 2;
 			if (longitude > mid)
 			{
-				ch |= 0x0001 << bit;
+				ch |= 0x0001u << bit;
 				lon[0] = mid;
 			}
 			else


strk:
Looks good, but may I ask on which sistem you get the runtime error ?
I'd love to see that runtime error in one of the bots we're using.
Maybe CARTO would be interested in donating a Jenkins client to add
another test environment ?

Algunenano:
I'm running on x86_64 with sanitizer flags: ```-fsanitize=undefined``` in this case.

It should be possible to add some flags and check them (depending on the configuration it will, or not, stop the program). Not sure if these flags are available in gcc 4.8.4 though, but I can confirm it works in gcc 7.2.0 and clang 5.0.0.

pramsey:
Looks like we could do that trick on travis w/ the right toolchain https://docs.travis-ci.com/user/languages/c/

strk:
Yep, and/or on Jenkins,GitLab-CI,Drone

Drone builds are based on a Docker image so it should be very simple
to install what it takes to enable that. Please see
https://git.osgeo.org/gogs/postgis/postgis-docker
if you want to give it a try

strk:
Trac ticket corresponding to this PR is ?

Algunenano:
Sorry, it's here: https://trac.osgeo.org/postgis/ticket/3875

PR #150, opened 2017-10-05
Unify signum declarations and make it branchless

Algunenano:
There are 2 different definitions of signum (on as a macro and another one as a function). I've removed them and add the macro in `liblwgeom_internal.h`.

### DIFF
--- a/liblwgeom/liblwgeom_internal.h
+++ b/liblwgeom/liblwgeom_internal.h
@@ -122,6 +122,14 @@
 #define SIZE_SET(varsize, size) (((varsize) & 0x00000003)|(((size) & 0x3FFFFFFF) << 2 ))
 
 /**
+* Macro that returns:
+* -1 if n < 0,
+*  1 if n > 0,
+*  0 if n == 0
+*/
+#define SIGNUM(n) ((n > 0) - (n < 0))
+
+/**
 * Tolerance used to determine equality.
 */
 #define EPSILON_SQLMM 1e-8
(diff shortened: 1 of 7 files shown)


liblwgeom/liblwgeom_internal.h
```diff
@@ -122,6 +122,14 @@
 #define SIZE_SET(varsize, size) (((varsize) & 0x00000003)|(((size) & 0x3FFFFFFF) << 2 ))
 
 /**
+* Macro that returns:
+* -1 if n < 0,
+*  1 if n > 0,
+*  0 if n == 0
+*/
+#define SIGNUM(n) (n > 0) - (n < 0)
```
Komzpa: `SIGNUM(10) * 5` will become `(10 > 0) - (10 < 0) * 5`. Maybe another parentheses?

Algunenano: You are totally right, I'll update it.

Komzpa: it bothers me that no test failed, although there are some maths in this PR affected by this.

Algunenano: It was never used as part of an math operation. I've added a couple of extra tests.

strk:
As annoying as it is, a link back to the associated trac ticket is important.
Here it is, manually placed: https://trac.osgeo.org/postgis/ticket/3878

@Komzpa this is another thing the "github" team could do: automate these cross-linking :)

liblwgeom/liblwgeom_internal.h
```diff
@@ -122,6 +122,14 @@
 #define SIZE_SET(varsize, size) (((varsize) & 0x00000003)|(((size) & 0x3FFFFFFF) << 2 ))
 
 /**
+* Macro that returns:
+* -1 if n < 0,
+*  1 if n > 0,
+*  0 if n == 0
+*/
+#define SIGNUM(n) ((n > 0) - (n < 0))
```
pramsey: the traditional thing (ha ha, like I'm an expert) for macro defined functions is to put parens around every use of the argument value, so that if the value is itself an expression, the expanded macro won't evaluate in surprising ways, like
```
#define SIGNUM(n) (((n) > 0) - ((n) < 0))
```

strk:
On Thu, Oct 05, 2017 at 05:50:54AM -0700, Paul Ramsey wrote:

> the traditional thing (ha ha, like I'm an expert) for macro defined functions is to put parens around every use of the argument value, so that if the value is itself an expression, the expanded macro won't evaluate in surprising ways, like
> ```
> #define SIGNUM(n) (((n) > 0) - ((n) < 0))

+1

PR #152, opened 2017-10-06
Fix undefined behaviour with negative input in TYPMOD_GET_SRID

Algunenano:
Fixes an undefined behaviour when the SRID is negative.

Some output example using undefined behaviour sanitizer:
```
Running test 'test_typmod_macros' in suite 'serialization/deserialization'.
cu_libgeom.c:35:7: runtime error: left shift of 535589632 by 3 places cannot be represented in type 'int'
cu_libgeom.c:45:14: runtime error: left shift of 280871168 by 3 places cannot be represented in type 'int'
```


Algunenano:
Associated trac ticket: https://trac.osgeo.org/postgis/ticket/3880

PR #153, opened 2017-10-06
Avoid undefined behaviour in zigzag with negative inputs

Algunenano:
Output from sanitizer before changes:
```
Running test 'test_zigzag' in suite 'varint'.                                                                                                                                                                                    
varint.c:170:14: runtime error: left shift of negative value -1
varint.c:175:14: runtime error: left shift of negative value -1
varint.c:180:14: runtime error: left shift of negative value -127
```

Changes in unzigzag weren't really necessary but I wanted to keep the branchless check as in zigzag.
Also, please let me know if you would rather have the (X ? Y : Z) in the same line


Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/3882

PR #156, opened 2017-10-10
Revert "Move remove_repeated_points to an in_place basis"

Komzpa:
This reverts commit 3fa892e8e0d32f8dde92964639fa4f71063993fc.

This commit broke regress testing.

### DIFF
--- a/liblwgeom/cunit/cu_algorithm.c
+++ b/liblwgeom/cunit/cu_algorithm.c
@@ -1003,28 +1003,6 @@ static void test_geohash_point_as_int(void)
 	CU_ASSERT_EQUAL(gh, rs);
 }
 
-static void test_lwgeom_remote_repeated_points(void)
-{
-	LWGEOM *g;
-	char *ewkt;
-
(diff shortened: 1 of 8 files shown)


Komzpa:
References https://trac.osgeo.org/postgis/ticket/3877

pramsey:
I see one broken test in topology that I'm working on now. Any more?

strk:
@pramsey it's hard to know if anything else breaks until we get failures on every single commit.
If you need more than a day or two it may be good to work in a branch (both github and gitlab mirrors support checking PRs automatically) and revert trunk meanwhile (I'm also working in topology so it helps to have a green light: https://trac.osgeo.org/postgis/ticket/3884)

Komzpa:
@pramsey
Travis is red on trunk, and that means two latest PRs on github aren't properly tested.
As far as I see it's just that one.
I'd be happy if there is a proper fix coming.
I'd like to put some work on -Werror and more strict UB sanitizing, and for that I need initially green CI.

pramsey:
can comment out the one test from regression suite for now and then put it back when I've found the fix

Komzpa:
Commenting out test will leave false impression of that ticket being still resolved, although there's regression on it. 
You can revert commit and put it back when you've found the fix. It's a clean revert anyway, just `git revert 3fa892e8e0d32f8dde92964639fa4f71063993fc` and it's done (and revert a revert when it's fixed).

pramsey:
OK, will bring them back when I've figured otu the topology issue

PR #157, opened 2017-10-10
Fix GEOSNode error message

jazzzz:
The log mentioned GEOSUnaryUnion instead of GEOSNode,


pramsey:
Thanks, committed at r15952

PR #161, opened 2017-10-11
BTree sorting logic fix

Komzpa:
Patch by Darafei Praliaskouski <me@komzpa.net>

 - allow fast path for geography points;
 - fix gcc warnings;
 - compare EMPTY geometries as normal memcmp path.

PR #162, opened 2017-10-11
Fix GCC warnings

Komzpa:
Patch by Darafei Praliaskouski <me@komzpa.net>

Obey const pointer in autofix logic.

### DIFF
--- a/liblwgeom/lwgeom_geos.c
+++ b/liblwgeom/lwgeom_geos.c
@@ -210,20 +210,40 @@ GEOS2LWGEOM(const GEOSGeometry *geom, char want3d)
 
 
 
-GEOSCoordSeq ptarray_to_GEOSCoordSeq(const POINTARRAY *);
-
+GEOSCoordSeq ptarray_to_GEOSCoordSeq(const POINTARRAY *, int fix_ring);
 
 GEOSCoordSeq
-ptarray_to_GEOSCoordSeq(const POINTARRAY *pa)
+ptarray_to_GEOSCoordSeq(const POINTARRAY *pa, int fix_ring)
 {
 	uint32_t dims = 2;
 	uint32_t i;
+	int append_points = 0;	
 	const POINT3DZ *p3d;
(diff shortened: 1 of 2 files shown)


liblwgeom/lwgeom_geos.c
```diff
@@ -273,18 +274,24 @@ ptarray_to_GEOSLinearRing(const POINTARRAY *pa, int autofix)
...
-			pa = ptarray_addPoint(pa, getPoint_internal(pa, 0), FLAGS_NDIMS(pa->flags), pa->npoints);
-			npa = pa;
+			npa = ptarray_addPoint(pa, getPoint_internal(pa, 0), FLAGS_NDIMS(pa->flags), pa->npoints);
 		}
 		/* Check ring for having at least 4 vertices */
-		while ( pa->npoints < 4 )
+		while ( npa->npoints < 4 )
 		{
-			ptarray_append_point(pa, getPoint_internal(pa, 0), LW_TRUE);
+			ptarray_append_point(npa, (POINT4D*)getPoint_internal(npa, 0), LW_TRUE);
```
pramsey: Would you use getPoint4d_p instead? Getting addicted to pointing right into the array is a good way to start reading off the end of it.

pramsey:
Very not good, in `make check`
```
  Test: test_lwgeom_clip_by_rect ...make[1]: *** [check] Segmentation fault: 11
```

pramsey:
Why all this mucking around screwing w/ const inputs in ptarray_to_GEOSLinearRing when in ptarray_to_GEOSCoordSeq you can create and return a coordseq of the desired size without mutating the input (and, btw, breaking the const contract!)

Komzpa:
@pramsey thanks! I did as you suggested. The only warning left is fixed in separate PR, https://github.com/postgis/postgis/pull/161

strk:
Trac ticket: https://trac.osgeo.org/postgis/ticket/3900

Komzpa:
Committed

PR #170, opened 2017-11-06
Documentation: extra details for st_voronoi parameters and return values

anneb:
Added some details for special cases, such as zero, one or more points


pramsey:
Doesn't seem greatly improved, at least in the patch view... https://github.com/postgis/postgis/pull/170/files

pramsey:
Committed to trunk at r16093

strk:
is our .editorconfig correctly set for .xml ?

PR #172, opened 2017-11-20
postgis/: Change TRUE and FALSE for true and false

Algunenano:
Required after changes in PostgreSQL:
From http://www.postgresql-archive.org/PostgreSQL-Weekly-News-November-19-2017-td5992809.html
    - Remove TRUE and FALSE.  Code should be using true and false.  Existing code
    can be changed to those in a backward compatible way.  The definitions in the
    ecpg header files are left around to avoid upsetting those users
    unnecessarily.  Reviewed-by: Michael Paquier <[hidden email]>
    https://git.postgresql.org/pg/commitdiff/6337865f36da34e9c89aaa292f976bde6df0b065

Patch by Raul Marin (Algunenano), Carto

### DIFF
--- a/libpgcommon/lwgeom_transform.c
+++ b/libpgcommon/lwgeom_transform.c
@@ -167,7 +167,7 @@ PROJ4SRSCacheIsEmpty(MemoryContext context)
 	 * Always return false since this call is mandatory according to tgl
 	 * (see postgis-devel archives July 2007)
 	 */
-	return FALSE;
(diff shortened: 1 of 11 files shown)


Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/3931

pramsey:
Commited, gracias https://trac.osgeo.org/postgis/ticket/3931#comment:1

PR #174, opened 2017-12-15
Fix parallel build directory creation in Makefile

Komzpa:


Komzpa:
merged in https://github.com/postgis/postgis/commit/e9dc0ba155dfdf711bd351d9437711dfcf4fc756

PR #175, opened 2017-12-22
Tree-based distance calculation

pramsey:
Builds an O(n) tree on geometry edges, using existing spatial autocorrelation to get "OK" spatial structure in the tree. Goes from there... handles all types, including crazy curves.


Komzpa:
Trac ticket is https://trac.osgeo.org/postgis/ticket/3949

Komzpa:
I've got in mind another algorithm that utilizes a tree and distance calculation:
https://github.com/mapbox/polylabel

Is this implementation of tree helping to implement that Pole of Inaccessibility search algorithm? If yes, would be great to replace ST_PointOnSurface with it for polygons :)

Komzpa:
This was merged at https://github.com/postgis/postgis/commit/5ad6e06a617eaa48a91631ef74e5e42caed75b3a

PR #176, opened 2017-12-22
Weighted median

Komzpa:
Todo:
- [x] test with all negative weights;
- [x] `Enhanced:` documentation entry;

### DIFF
--- a/liblwgeom/lwgeom_median.c
+++ b/liblwgeom/lwgeom_median.c
@@ -32,7 +32,7 @@
 #include "lwgeom_log.h"
 
 static void
-calc_weighted_distances_3d(const POINT3D* curr, const POINT4D* points, size_t npoints, double* distances)
+calc_weighted_distances_3d(const POINT3D* curr, const POINT4D* points, uint32_t npoints, double* distances)
 {
 	size_t i;
 	for (i = 0; i < npoints; i++)
@@ -42,9 +42,9 @@ calc_weighted_distances_3d(const POINT3D* curr, const POINT4D* points, size_t np
 }
 
 static double
-iterate_4d(POINT4D* curr, const POINT4D* points, size_t npoints, double* distances)
+iterate_4d(POINT4D* curr, const POINT4D* points, uint32_t npoints, double* distances)
 {
-	size_t i;
+	uint32_t i;
 	POINT4D next = { 0, 0, 0, 1};
 	double delta = 0;
 	double denom = 0;
@@ -125,11 +125,11 @@ iterate_4d(POINT4D* curr, const POINT4D* points, size_t npoints, double* distanc
 }
 
 static POINT4D
-init_guess(const POINT4D* points, size_t npoints)
+init_guess(const POINT4D* points, uint32_t npoints)
 {
 	assert(npoints > 0);
 	POINT4D guess = { 0, 0, 0, 0 };
-	size_t i;
+	uint32_t i;
 	for (i = 0; i < npoints; i++)
 	{
 		guess.x += points[i].x * points[i].m;
@@ -144,14 +144,14 @@ init_guess(const POINT4D* points, size_t npoints)
 }
 
 static POINT4D*
-lwmpoint_extract_points_4d(const LWMPOINT* g, size_t* npoints, int* input_ok)
+lwmpoint_extract_points_4d(const LWMPOINT* g, uint32_t* npoints, int* input_ok)
 {
 	assert(npoints != NULL);
 	assert(input_ok != NULL);
 	assert(*input_ok == LW_TRUE);
-	size_t i;
-	size_t n = 0;
-	int is_3d = lwgeom_has_z((LWGEOM*) g);
+	uint32_t i;
+	uint32_t n = 0;
+	int has_z = lwgeom_has_z((LWGEOM*) g);
 	int has_m = lwgeom_has_m((LWGEOM*) g);
 
 	POINT4D* points = lwalloc(g->ngeoms * sizeof(POINT4D));
@@ -194,7 +194,7 @@ lwmpoint_extract_points_4d(const LWMPOINT* g, size_t* npoints, int* input_ok)
 	}
 	if (input_ok)
 	{
-		if (!is_3d) 
+		if (!has_z) 
 		{
 			for (i = 0; i < n; i++)
 			{
@@ -219,14 +219,14 @@ lwmpoint_extract_points_4d(const LWMPOINT* g, size_t* npoints, int* input_ok)
 LWPOINT*
 lwmpoint_median(const LWMPOINT* g, double tol, uint32_t max_iter, char fail_if_not_converged)
 {
-	size_t npoints = 0; /* we need to count this ourselves so we can exclude empties and weightless points */
-	size_t i;
+	uint32_t npoints = 0; /* we need to count this ourselves so we can exclude empties and weightless points */
+	uint32_t i;
 	int input_ok = LW_TRUE;
 	double delta = DBL_MAX;
+	POINT4D median;
 	double* distances;
 	/* m ordinate is considered weight, if defined */
-	POINT4D* points = lwmpoint_extract_points_4d(g, &npoints, &input_ok);
-	POINT4D median;
+	POINT4D* points = lwmpoint_extract_points_4d(g, &npoints, &input_ok);	
 
 	if (!input_ok)
 	{
@@ -257,8 +257,8 @@ lwmpoint_median(const LWMPOINT* g, double tol, uint32_t max_iter, char fail_if_n
 		delta = iterate_4d(&median, points, npoints, distances);
 	}
 
-	lwfree(points);
 	lwfree(distances);
+	lwfree(points);
 
 	if (fail_if_not_converged && delta > tol)
 	{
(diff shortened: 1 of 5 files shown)


liblwgeom/lwgeom_median.c
```diff
@@ -28,43 +28,50 @@
 #include "lwgeom_log.h"
 
 static void
-calc_distances_3d(const POINT3D* curr, const POINT3D* points, uint32_t npoints, double* distances)
+calc_weighted_distances_3d(const POINT4D* curr, const POINT4D* points, size_t npoints, double* distances)
 {
-	uint32_t i;
```
dbaston: PostGIS point array size is limited to 32-bit. (I'm not sure why all counters were changed to `size_t` ?)

dbaston: Add comment to clarify that `m` ordinate is taken to be a weight

liblwgeom/lwgeom_median.c
```diff
@@ -28,43 +28,50 @@
...
 	}
+	if (!FP_IS_ZERO(denom))
+	{
+		next.x /= denom;
+		next.y /= denom;
+		next.z /= denom;
+	}
+	else
+	{
+		hit = LW_TRUE;
```
dbaston: Can this statement have an effect?

liblwgeom/lwgeom_median.c
```diff
@@ -28,43 +28,50 @@
...
 static void
-calc_distances_3d(const POINT3D* curr, const POINT3D* points, uint32_t npoints, double* distances)
+calc_weighted_distances_3d(const POINT4D* curr, const POINT4D* points, size_t npoints, double* distances)
 {
-	uint32_t i;
+	size_t i;
 	for (i = 0; i < npoints; i++)
 	{
-		distances[i] = distance3d_pt_pt(curr, &points[i]);
+		distances[i] = distance3d_pt_pt((POINT3D*)curr, (POINT3D*)&points[i]) * points[i].m;
```
dbaston: I'd suggest using a weighted distance function that accepts two `POINT4D` pointers, instead of casting to `POINT3D` every time this is called.

liblwgeom/lwgeom_median.c
```diff
@@ -115,37 +119,58 @@ iterate_3d(POINT3D* curr, const POINT3D* points, uint32_t npoints, double* dista
...
+			if (!has_m)
+			{
+				points[n].m = 1.0;
+				n++;
+			}
+			else
+			{
+				if (!FP_IS_ZERO(points[n].m))
+				{
+					n++;
```
dbaston: Add comment that points with zero weight are ignored

dbaston: What happens when weights are negative?

liblwgeom/lwgeom_median.c
```diff
@@ -173,11 +198,17 @@ lwmpoint_median(const LWMPOINT* g, double tol, uint32_t max_iter, char fail_if_n
 
 	median = init_guess(points, npoints);
 
+	if (FP_IS_ZERO(median.m))
```
dbaston: Under what condition does our initial guess have a zero m value, and why do we want to return empty in this case?

liblwgeom/lwgeom_median.c
```diff
@@ -173,11 +198,17 @@ lwmpoint_median(const LWMPOINT* g, double tol, uint32_t max_iter, char fail_if_n
...
+	{
+		lwfree(points);
+		return lwpoint_construct_empty(g->srid, 0, 0);
+	}
+
 	distances = lwalloc(npoints * sizeof(double));
 
 	for (i = 0; i < max_iter && delta > tol; i++)
```
dbaston: Does weighting impact the significance of `tol` ?

liblwgeom/lwgeom_median.c
```diff
@@ -28,43 +28,50 @@
...
 }
 
 static double
-iterate_3d(POINT3D* curr, const POINT3D* points, uint32_t npoints, double* distances)
+iterate_4d(POINT4D* curr, const POINT4D* points, size_t npoints, double* distances)
 {
-	uint32_t i;
-	POINT3D next = { 0, 0, 0 };
+	size_t i;
+	POINT4D next = { 0, 0, 0, 1 };
```
dbaston: Does `next` need to be `POINT4D` ? Is the `m` ordinate ever accessed?

liblwgeom/lwgeom_median.c
```diff
@@ -158,12 +183,12 @@ lwmpoint_extract_points_3d(const LWMPOINT* g, uint32_t* ngeoms)
...
-	uint32_t npoints; /* we need to count this ourselves so we can exclude empties */
-	uint32_t i;
+	size_t npoints; /* we need to count this ourselves so we can exclude empties and weightless points */
+	size_t i;
 	double delta = DBL_MAX;
 	double* distances;
-	POINT3D* points = lwmpoint_extract_points_3d(g, &npoints);
-	POINT3D median;
+	POINT4D* points = lwmpoint_extract_points_4d(g, &npoints);
+	POINT4D median;
```
dbaston: Does this need to be 4D? The returned values are either 2D or 3D.

liblwgeom/lwgeom_median.c
Komzpa: Accessing 32-bit variables in 64-bit environments is slower in my experience, especially in loops. To get that right across architectures maps.me uses internal convention to use `size_t` for counters.

Documentation change should be there, at least `Enhanced:` entry.

liblwgeom/lwgeom_median.c
Komzpa: I'd even prefer to hand-inline the logic as it's a single line :)
Tricky part is to get naming right: the weight is stored just in one of the points, and if you pass two `POINT4D`'s you're passing two of them.

liblwgeom/lwgeom_median.c
Komzpa: A point with negative weight will cancel out a point with positive weight. When total sum of weights is negative, the point will go to infinity. Need to add a check for that.

dbaston:
Seems reasonable, but I'd like to review and test before this is committed to trunk.

liblwgeom/lwgeom_median.c
Komzpa: Initial guess can have zero m value if sum of negative values is equal to sum of positive values. 

Let's visualize sum of distances for all cases: 
larger than zero weight, 
`create table median_0_debug as (select ST_SetSRID( ST_MakePoint(x, y), 3857) as geom,  5*sqrt(x^2+y^2) -  sqrt((x-10)^2+y^2)    from generate_series(-100, 100) x, generate_series(-100,100) y);`
image
 - we have a nice minimum.

Weight of zero,
`create table median_0_debug as (select ST_SetSRID( ST_MakePoint(x, y), 3857) as geom,  sqrt(x^2+y^2) -  sqrt((x-10)^2+y^2)    from generate_series(-100, 100) x, generate_series(-100,100) y);`
image
- minimum is actually a line in this case. If we add more points, it shows it may be possible to find median:

image

Unfortunately, this causes division by zero in centroid code now. A common solution to issues in median now seems to be "fail if asked to fail, empty othewise".

Less than zero:
`create table median_0_debug as (select ST_SetSRID( ST_MakePoint(x, y), 3857) as geom, -1* sqrt(x^2+y^2) -  sqrt((x-10)^2+y^2)    from generate_series(-100, 100) x, generate_series(-100,100) y);`
image
 - we have a peak in center, and minimum is in infinity somewhere. It's basically inverse of positive case. We really don't want to iterate to infinity, and unlikely want `POINT(Infinity, Infinity)` in database.

Komzpa:
@dbaston 
I've done some more research and found this:
http://onlinelibrary.wiley.com/doi/10.1111/j.1467-9787.1992.tb00200.x/abstract

Basically Weiszfeld algorithm may find just a local optimum in case of negative weights, as problem is not convex anymore. It's going to be one closest to the centroid, though.

Fermat-Weber problem is constrained by positive only weights. 

There are different ways I see to resolve:

 - add warning to documentation that in case of negative weights it may be just local optimum.

 - fail immediately on negative weights input. we can fail not always, but only on fail_if_not_converged is true - local optimum is a kind of non-converged global optimum :)

 - replace Weiszfeld algorithm with another tweaked version, proposed say in this paper: https://www.researchgate.net/publication/265359176_The_Weber_Problem_On_The_Plane_With_Some_Negative_Weights

dbaston:
I think the simplest thing would be to fail on negative weights. Do you have a particular use case for negative weights?

dbaston:
By "fail on negative weights," I mean remove support for them entirely. It seems like they add a lot of code complexity. Is there a use case that benefits from the way they're partially handled in this PR?

dbaston:
I'm going to mute this PR; please notify me back in if there's something to look at.

liblwgeom/lwgeom_median.c
strk: Ansi C does not allow declaring variables after code (my compiler complains about this)

Komzpa:
@dbaston please take another look. it's now fixed.

dbaston:
@Komzpa thanks. Do you a know a way to get GitHub to compare this PR with the repository state before the weighted changed were committed?

Komzpa:
@dbaston set up one for you by trimming top of svn-trunk history.
It's a bit crowded but has all median as single diff.
 https://github.com/Komzpa/postgis/pull/1/files

PR #178, opened 2017-12-25
Delete box2d.c

Komzpa:
Coverage suggests that box2d.c isn't really used. Check if removing doesn't break build.


Komzpa:
ticket: https://trac.osgeo.org/postgis/ticket/3955

PR #186, opened 2018-01-09
Simplify KMeans code

Komzpa:
https://trac.osgeo.org/postgis/ticket/3977


liblwgeom/lwkmeans.c
```diff
@@ -1,77 +1,195 @@
...
-	POINT2D sum;
-	POINT2D **pts = (POINT2D**)objs;
-	POINT2D *center = (POINT2D*)centroid;
-
-	sum.x = sum.y = 0.0;
+	unsigned int i;
+	unsigned int* weights;
 
-	if (num_objs <= 0) return;
+	weights = lwalloc(sizeof(int) * config->k);
```
pramsey: Consider moving this to config, so you don't have to alloc/free it on each iteration.

Komzpa: Moved out of function. Got rid of config altogether, -50 lines.

Komzpa:
@pramsey @tomvantilburg please take a look and wave an OK if it's fine to commit.

tomvantilburg:
I'm happy with the new implementation. It is certainly faster, and much better than what we had :wave:
Just for the looks of it, here's the result of the clustering on a regular grid (bluer is larger maxdistance, from 1.4 to 4.5):
image

tomvantilburg:
And the situation as it was previously (curiously, dropping 2 clusters out of 1000):
image

pramsey:
:wave:

PR #188, opened 2018-01-10
clang-format

Komzpa:
What if we used clang-format to format all our c files?


Algunenano:
I think this is a pretty good idea and I would add it also as part of the PR checks.

Once that is said, I see that the column limit is set to 120 but most of the code uses 80. Since the great thing about clang-format is making it automatic I'd rather set it lower and let it handle it.

Komzpa:
@Algunenano I changed it to 80.
For now we can only use clang-format as recommendation, not requirement.

We can't require clang-format-5.0, it hangs on uthash.h:
https://bugs.llvm.org/show_bug.cgi?id=35900

We can't require clang-format-6.0 yet, it's not released yet and it breaks code in lwgeom_api.c:
https://bugs.llvm.org/show_bug.cgi?id=35936

Algunenano:
Maybe it could be applied over the PR patch, not the whole project. In my opinion, It's more about making PR reviews simpler and cleaner more than having an uniform style across every file.

Komzpa:
It looks like after git-clang-format can do it:
```
git-clang-format-6.0 --style=file
```
formats all the staged files only in changed lines.

Komzpa:
lwgeom_api.c issue went away with code update.

new blocker is `extensions/address_standardizer/pagc_api.h`:
https://bugs.llvm.org/show_bug.cgi?id=36002

PR #190, opened 2018-01-13
make lwprint_double shared beteween output functions

Komzpa:


Komzpa:
Ticket: https://trac.osgeo.org/postgis/ticket/3987

PR #191, opened 2018-01-14
Geography overlays

Komzpa:
https://trac.osgeo.org/postgis/ticket/3973


pramsey:
Slow down, this isn't a gimmee https://lists.osgeo.org/pipermail/postgis-devel/2018-January/026800.html

Komzpa:
@pramsey let's have another look after it compiles and has decent test coverage.

robe2:
FWIW I did soften my blow to a -0.5

https://lists.osgeo.org/pipermail/postgis-devel/2018-January/026801.html

But I did change the ticket to `PostGIS Fund Me`

I'm feeling this won't be ready for 2.5 unless @dangerground has some time to put into it from now til  August 2018 which it sounded like he might not.  Sounds like it needs a lot of work to be ready.

PR #197, opened 2018-01-24
Add ST_MinimumRotatedRectangle

dbaston:


Komzpa:
I've tried to summarize already used naming in PostGIS:

image

For more words:
http://www.thesaurus.com/browse/boundary?s=t
http://www.thesaurus.com/browse/box?s=t
http://www.thesaurus.com/browse/envelope?s=t
http://www.thesaurus.com/browse/extent?s=t

Komzpa:
@dbaston I think function should be renamed.

"minumum" should say what's exactly being minimized. It's `Radius` in `MinimumBoundingRadius`. For box, is it perimeter, area or are they guaranteed to always go smaller together?

This wiki page suggests that minimum area and minimum perimeter are diferent, but the papers in the prooflink don't actually seem to mention minimum-perimeter boxes:
https://en.wikipedia.org/wiki/Minimum_bounding_box_algorithms

At least for degenerate line case there are infinitely many minimal area boxes that are all just the line continued. 

Googling around suggests word "rotated" isn't used much to describe the thing.
JTS says it's minimumRectangle, https://locationtech.github.io/jts/javadoc/org/locationtech/jts/algorithm/MinimumDiameter.html#getMinimumRectangle() - but minimizes the input's Diameter, which is one of the sides in .

"rectangle" means this function is stuck in 2D forever.

I think it should just go into `ST_Envelope`. We've got enough other axis aligned box functions already, and nothing in its name suggests it must be axis aligned.

@robe2 @pramsey thoughts?

dbaston:
@Komzpa what are you suggesting the function be renamed to?

There was some past discussion here: https://github.com/OSGeo/geos/pull/54

dbaston:
Lively IRC discussion suggests this should be `ST_OrientedEnvelope` instead.

dbaston:
Can we turn the stupid covecov bot off?

dbaston:
Or are we worried that this patch decreases coverage by 0.02%?

strk:
> Can we turn the stupid covecov bot off?

+1 please shot it down

Komzpa:
Hi, 

can you please elaborate on "stupidity" of the bot? what exactly about it you don't like?

It's exposing imperfections in code and commits, that is true.

It really helps with big rewrites, such as KMeans and Median for me - it quickly spotlights lines of code that need more attention, that for instance led to fix of #3965 (lack of testing for the "find a replacement id" loop stood out).

dbaston:
From my point of view, it worsens the signal/noise ratio of communications
from GitHub, in two ways:

1) It sends out an email notification after every single commit, with
information that I'm generally not interested in (most code coverage %
changes are negligible)
2) It is being lumped in with the unit tests to determine pass/failure of a
PR. The PR pass/fail already has enough noise with Travis/Buildbot
malfunctions. Failing a PR because it decreases questionably-measured code
coverage by hundredths of a percent just makes the information value of the
pass/fail status even lower. I say "questionably measured" because we have
(a) code running in multiple langauages, (b) code running in and out of
PostgreSQL, (c) complex sections of library-version-dependent code that are
turned on/off by the preprocessor, (d) autogenerated code, and (e) code
that executes differently depending on geometry types, number of inputs,
etc. (ie if your PR has 100% coverage, but your test cases are all single
points, then probably aren't exercising it much.).

I do think code coverage reporting can be valuable, and it would be helpful
to see reports generated by Jenkins or something else (especially the type
where you can see all of the source, with lines color coded by whether or
not they're hit in testing.). But we don't need to send out noisy emails
and autofail PRs based on it. If you are doing big rewrites, I would
recommend running this kind of stuff locally. Maybe we could add some wiki
instructions for it.

Dan

strk:
Me I'm only annoyed at the mail notification, if it was a silent
web-only report wouldn't disturb me too much.

pramsey:
I'd be happy to see it gone. It doesn't add anything to my work flow personally.

Komzpa:
I've switched off the comment posting. It should now only build a web report and post a tickmark as CI integration.

dbaston:
Can some meaningful threshold value be set for the tickmark?

PR #198, opened 2018-01-26
lwgeom_median: Avoid division by zero

Algunenano:
Related to https://trac.osgeo.org/postgis/ticket/3997

### DIFF
--- a/liblwgeom/cunit/cu_algorithm.c
+++ b/liblwgeom/cunit/cu_algorithm.c
@@ -1230,6 +1230,19 @@ static void test_median_robustness(void)
 	/* Unsupported geometry type */
 	do_median_test("POLYGON((1 0,0 1,1 2,2 1,1 0))", NULL, LW_TRUE, 1000);
 	do_median_test("POLYGON((1 0,0 1,1 2,2 1,1 0))", NULL, LW_FALSE, 1000);
+
+	/* Intermediate point included in the set */
+	do_median_test("MULTIPOINT ZM ("
+			"(0 0 20000 0.5),"
+			"(0 0 59000 0.5),"
+			"(0 48000 -20000 1.3),"
+			"(0 -48000 -20000 1.3),"
+			"(0 -3000 -3472.22222222222262644208967685699462890625 1),"
+			"(0 3000 3472.22222222222262644208967685699462890625 1),"
+			"(0 0 -1644.736842105263121993630193173885345458984375 1),"
+			"(0 0 1644.736842105263121993630193173885345458984375 1)"
+			")",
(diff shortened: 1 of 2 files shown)


liblwgeom/cunit/cu_algorithm.c
```diff
@@ -1229,6 +1230,19 @@ static void test_median_robustness(void)
...
+
+	/* Intermediate point included in the set */
+	do_median_test("MULTIPOINT ZM ("
+			"(0 0 20000 0.5),"
+			"(0 0 59000 0.5),"
+			"(0 48000 -20000 1.3),"
+			"(0 -48000 -20000 1.3),"
+			"(0 -3000 -3472.22222222222262644208967685699462890625 1),"
+			"(0 3000 3472.22222222222262644208967685699462890625 1),"
+			"(0 0 -1644.736842105263121993630193173885345458984375 1),"
```
Komzpa: I love this one. Really great, @Algunenano :)

Algunenano: Sad news. For some reason (I'm guessing different float operation order) this hits in clang (5.0) but not in gcc (7.2.1, 6.4.1). I'm investigating the issue to try to find a common case

Algunenano: Good news :smile: The issue was due to invalid rounding (-0.00000000000029917588874109483636135671651489707215 vs 0.0) in `init_guess` with the  `-ffloat-store` flag  that I guess clang just ignores.
I'll have a look to see if the flag is really useful.

Komzpa:
Hi @pramsey and @Algunenano,

since change https://github.com/postgis/postgis/commit/37bb96a2fc94719517102bef562d0149b2834d30 the tests are red and the change was introduced in https://github.com/postgis/postgis/commit/3bb8b930ea49e71dd84f91c0085b143744dafd0c#diff-245be84dc8debe637ec06a7838cccb96L1146

can something please be done about it?

Algunenano:
I'm investigating the issue. It appears that (0 * 0.5 + 0 * 0.5 + 48000  * 1.3 - 48000 * 1.3 - 3000 * 1 + 3000 * 1 + 0 * 1 + 0 * 1) doesn't result in 0 if the flag `-ffloat-store` is activated, but `-0.0000000000002991758...` instead so it takes a lot more iterations to end. Clang doesn't appear to support it so it was deactivated in all my local tests.

We can either remove the tests if the flag is active or deactivate the flag by default (and activate it only under some `configure` option) but I need to make some more tests before proposing a fix. I'll send a PR either today or tomorrow in any case.

PR #199, opened 2018-01-26
lwpoly_construct_circle: Avoid division by zero

Algunenano:
Delay calculating theta until after the number of segments is checked. Since the division by zero is undefined the compiler could optimize away the `if (segments_per_quarter < 1)` check.

### DIFF
--- a/liblwgeom/lwpoly.c
+++ b/liblwgeom/lwpoly.c
@@ -120,13 +120,13 @@ LWPOLY*
 lwpoly_construct_circle(int srid, double x, double y, double radius, uint32_t segments_per_quarter, char exterior)
 {
 	const uint32_t segments = 4*segments_per_quarter;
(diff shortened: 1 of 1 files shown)


Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/4003

Algunenano:
A couple of builds failed due to unrelated issues:
```
The command "sudo apt-get install -y --allow-unauthenticated --no-install-recommends --no-install-suggests libsfcgal1 libsfcgal-dev postgresql-server-dev-9.6 libxml2-utils libcunit1-dev xsltproc docbook-xsl docbook-mathml dblatex libgeos-dev libjson0-dev libprotobuf-c0-dev libgdal-dev gdb libc6-dbg" failed and exited with 100 during .
```

PR #200, opened 2018-01-26
lwprint_double: Avoid undefined behaviour with infinity

Algunenano:
lwprint.c:491:12: runtime error: inf is outside the range of representable values of type 'int'

Detected in test_wkt_in_point with "POINT(1e700 0)"


Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/4005

PR #201, opened 2018-01-31
Example 4D geojson support

Komzpa:
Building an example for the 4D support


Komzpa:
@belug23 are you interested in continuing the work on this? :)

PR #203, opened 2018-02-04
Single-sided buffer

stev-0:
See https://lists.osgeo.org/pipermail/postgis-devel/2018-January/026852.html for discussion.

There is also a seperate branch on trunk (this is built on top of 2.4)


Komzpa:
This is not going to be merged, as PR implements new feature in stable branch (this can go to trunk only).

PR #205, opened 2018-02-05
Fix broken median tests

Algunenano:
Changed the method to check the tolerance in the test to use the sum of the distances to the obtained point instead of comparing directly coordinates.

Also I had some issues with GCC and `-O3` as it introduces some errors that causes the condition (intermediate point included in the set) to be missed. It works fine with `-O0` or `-O2` and clang `-O3`.

### DIFF
--- a/liblwgeom/cunit/cu_algorithm.c
+++ b/liblwgeom/cunit/cu_algorithm.c
@@ -1175,10 +1175,10 @@ static void do_median_test(char* input, char* expected, int fail_if_not_converge
 				double distance_expected = test_weighted_distance(&expected_pt, points, npoints);
 				double distance_result = test_weighted_distance(&actual_pt, points, npoints);
 
-				passed = fabs(distance_expected - distance_result) < tolerance;
+				passed = distance_result <= (1.0 + tolerance) * distance_expected;
 				if (!passed)
 				{
-					printf("Diff: %.10f\n", distance_result - distance_expected);
+					printf("Diff: Got %.10f Expected %.10f\n", distance_result, distance_expected);
 				}
 				lwfree(points);
 			}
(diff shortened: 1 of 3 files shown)


Algunenano:
Related to trac issue: https://trac.osgeo.org/postgis/ticket/3997

Algunenano:
After further investigation it seems that the test relies in 80bit precision to work correctly. I'm working to find a test that works with just 64 bit floats.

Algunenano:
It seems than when looking for fine grained results certain issues dependent on how many bits are used for the float operations start to appear.

I've added (hidden behind `#ifdef 0`)  a test that fails when using 80-bit precision and another one that fails under 64bit doubles. 
Also I've added a test that fails under both (infinite loop). I think it should have been solved by #176 but isn't so a deeper look is needed (or maybe switching to a different algorithm), but I'd rather have the build bots green now and deal with this if a different issue.

pramsey:
I'm seeing regression failures in median now on OSX. 
```
Butterfly:~/Code/postgis-svn/liblwgeom/cunit pramsey$ ./cu_tester test_median_robustness

Running test 'test_median_robustness' in suite 'algorithm'.
median_test input MULTIPOINT ((1 -1 3), (1 0 2), (2 1 1)) (parsed MULTIPOINT(1 -1 3,1 0 2,2 1 1)) expected POINT(1 0 2) got POINT(1 0 2)
median_test input MULTIPOINT ZM ((0 0 20000 0.5),(0 0 59000 0.5),(0 48000 -20000 1.3),(0 -48000 -20000 1.3),(0 -3000 -3472.22222222222262644208967685699462890625 1),(0 3000 3472.22222222222262644208967685699462890625 1),(0 0 -1644.736842105263121993630193173885345458984375 1),(0 0 1644.736842105263121993630193173885345458984375 1)) (parsed MULTIPOINT(0 0 20000 0.5,0 0 59000 0.5,0 48000 -20000 1.3,0 -48000 -20000 1.3,0 -3000 -3472.22222222 1,0 3000 3472.22222222 1,0 0 -1644.73684211 1,0 0 1644.73684211 1)) expected POINT(0 0 0) got POINT(0 0 0)

    FAILED - asserts -  25 passed,   2 failed,  27 total.


  1. cu_algorithm.c:1171  - CU_ASSERT_TRUE(passed)
  2. cu_algorithm.c:1171  - CU_ASSERT_TRUE(passed)
```

Algunenano:
Have you enabled the test or are you testing something different from this PR?

I disabled it (and added another one disabled too) because they fail depending on the precision of your double operations (64 or 80 bit). 
I'm not sure if the failures are due to rounding or due to the modification to the algorithm (when finding a point in the input) breaking the fact that each iteration should be better than the previous. Until further investigation I'd rather have them disabled so the CI is green.

pramsey:
I'm not building the PR, this is just tests running in trunk.

PR #211, opened 2018-02-20
rtpg_wkb: Add missing include

Algunenano:


Komzpa:
is that for https://trac.osgeo.org/postgis/ticket/4021 ? :)

Algunenano:
Yes, it was introduced in https://trac.osgeo.org/postgis/ticket/2221 and I'm just waiting for the test to pass.

Komzpa:
Thanks! CI passed, patch committed :)

dustymugs:
How odd. Nothing about rtpg_wkb.c requires lwgeom...

Algunenano:
`RASTER_fromHexWKB` uses `text2cstring`.

dustymugs:
Yep. I see that. That wasn't intended. Should have been using text_to_cstring instead.

pramsey:
There's no particular reason we couldn't strip out our implementation. I cannot recall if I put it in because there was no text_to_cstring at the time, or because I just didn't find it in the pgsql code base.

Algunenano:
> There's no particular reason we couldn't strip out our implementation. I cannot recall if I put it in because there was no text_to_cstring at the time, or because I just didn't find it in the pgsql code base.

I've made a quick test and found there is at least issues with `lwgeom_in_geojson.c` as both Postgres `utils/builtins.h` and LibJSON `json-c/json.h` declare ```json_object```.

PR #213, opened 2018-02-20
Drop cstring2text/text2cstring in favor of Postgresql functions

Algunenano:
To avoid future issues (as discussed in #211) I've changed all calls of `cstring2text` / `text2cstring` for cstring_to_text / `text_to_cstring` from `utils/builtins.h`.

As `lwgeom_in_geojson.c` can't include both `json-c/json.h` and `utils/builtins.h` I've moved the old declarations there and left a warning since I'm pretty sure that error will come up in the future.

### DIFF
--- a/libpgcommon/lwgeom_pg.c
+++ b/libpgcommon/lwgeom_pg.c
@@ -31,40 +31,6 @@
 
 #define PGC_ERRMSG_MAXLEN 2048 //256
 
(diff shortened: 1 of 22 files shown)


Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/4022#ticket

PR #214, opened 2018-02-20
Fix cast from box3d to geometry

Mat2095:
Polyhedral Surface was not correctly connected


Mat2095:
Trac issue: https://trac.osgeo.org/postgis/ticket/4020

PR #222, opened 2018-02-26
[travis] codecov for geos/raster

Komzpa:


pramsey:
I was rather hoping we could have less (none) codecov in our CI builds, not more.

Komzpa:
@pramsey are you talking about the comment or the concept?

Comment is gone, may only appear in some PRs that were forked before config committed.

As a concept, I'll be happy if you show me another way to understand which part of lwgeom_geos.c has testcases that use the code and which does not. :)

PR #225, opened 2018-02-28
Bit-setting to optimize geoms for compression (precision-based)

dbaston:
This is a rework of #223 that is based on "digits of precision" (number of digits to the right of the decimal point). I think this is more useful than the "significant digits" approach of #223, and more in line with what users expect given other PostGIS functions (precision arg to ST_AsText, for example).

I also did some real-world testing using the TIGER MCD layer. This dataset is provided with six digits of precision, and the geometry column occupies 499 MB. Running `postgis_optimize_geometry` with a precision of `6` reduces this to 256 MB. Joining the updated table produces no differences when geometries are compared with `ST_AsText(geom, 6)`.  Here is an example of what the modified coordinates look like:

Original:

`MULTIPOLYGON(((-96.915638 37.214606,-96.915629 37.215194,-96.915626 37.21539,-96.914793 37.215383,`

Modified:

`MULTIPOLYGON(((-96.9156379699707 37.2146058082581,-96.9156289100647 37.2151939868927,-96.91562557220
`
Interestingly, a quick performance check (point-in-polygon) gets about 25% worse with the storage-optimized geometry.


liblwgeom/lwgeom.c
```diff
@@ -2395,3 +2396,59 @@ lwgeom_is_trajectory(const LWGEOM *geom)
...
+	return *((double*) double_bits);
+}
+
+static double trim_preserve_decimal_digits(double d, int32_t decimal_digits)
+{
+	if (d==0)
+		return 0;
+
+	int digits_left_of_decimal = (int) (1 + log10(fabs(d)));
+	uint32_t bits_needed = bits_for_precision(abs(decimal_digits) + digits_left_of_decimal);
```
mguzelevich: http://www.cplusplus.com/reference/clibrary/cmath/frexp/

dbaston: @mguzelevich Could you elaborate on how to use that here?

Komzpa:
>>Interestingly, a quick performance check (point-in-polygon) gets about 25% worse with the storage-optimized geometry.

can you share the way you performed a check? did you CLUSTER / VACUUM FULL the table after making an UPDATE?

dbaston:
There was no `UPDATE` involved, the procedure was basically:
```

CREATE TABLE cousub_a AS SELECT row_number() OVER() AS gid, geom FROM cousub;
CREATE TABLE cousub_b AS SELECT gid, postgis_optimize_geometry(geom, 6) FROM cousub_a;
CREATE INDEX ON cousub_a USING gist(geom);
CREATE INDEX ON cousub_b USING gist(geom);

SELECT count(*) FROM testpoints, cousub_a WHERE ST_Intersects(testpoints.geom, cousub_a.geom);
SELECT count(*) FROM testpoints, cousub_b WHERE ST_Intersects(testpoints.geom, cousub_b.geom);
```

Komzpa:
@dbaston a possible scenario: instead of going to TOAST tables because the values were large, they fit into the heap itself after compression, making *less* tuples per page :)

easy enough to peek with `select max((ctid::text::point)[1]) from cousub_a` for both tables.

PR #226, opened 2018-02-28
pgindent all the files

Komzpa:
what if we used pgindent to format all the files?


strk:
were the detected bugs reported upstream ?
https://git.postgresql.org/git/pg_bsd_indent.git
https://www.postgresql.org/list/pgsql-bugs/

PR #228, opened 2018-03-06
Fix invalid inputs for GEOS if it cannot process them

Komzpa:


Komzpa:
https://trac.osgeo.org/postgis/ticket/4037

PR #230, opened 2018-03-09
WIP: Trunk memory leaks

Algunenano:
Fixing trunk memory leaks

Work in progress


Algunenano:
Running liblwgeom tests without logs showed some issues with the new makeValid and subdivide functions, and enabling the logs showed another ~1500 errors. Should be ok now.

PR #231, opened 2018-03-11
Replace GEOSPreparedContains with internal winding number algorithm

bjornharrtell:
Replica of https://git.osgeo.org/gitea/postgis/postgis/pulls/24 for CI.

### DIFF
--- a/liblwgeom/lwgeom_topo.c
+++ b/liblwgeom/lwgeom_topo.c
@@ -5015,7 +5015,6 @@ _lwt_AddPoint(LWT_TOPOLOGY* topo, LWPOINT* point, double tol, int
     LWGEOM *g = lwline_as_lwgeom(e->geom);
     LWGEOM *prj;
     int contains;
-    GEOSGeometry *prjg, *gg;
     LWT_ELEMID edge_id = e->edge_id;
 
     LWDEBUGF(1, "Splitting edge %" LWTFMT_ELEMID, edge_id);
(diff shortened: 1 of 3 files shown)


topology/test/perf/TopoGeo_addLinestring.sql
```diff
@@ -8,3 +8,5 @@ SELECT count(*) FROM (
       generate_series(-15,15,5) y
   ) foo
 ) bar;
+\timing
```
Komzpa: can it be explicit `\timing on` or `\timing off`?

Komzpa:
@bjornharrtell can you also update NEWS and post a link to trac ticket?

liblwgeom/lwgeom_topo.c
```diff
@@ -616,46 +615,27 @@ _lwt_CheckEdgeCrossing( LWT_TOPOLOGY* topo,
...
-    GEOSGeom_destroy(nodegg);
-    if (contains == 2)
-    {
-      GEOSPreparedGeom_destroy(prepared_edge);
-      GEOSGeom_destroy(edgegg);
-      _lwt_release_nodes(nodes, num_nodes);
-      lwerror("GEOS exception on PreparedContains: %s", lwgeom_geos_errmsg);
-      return -1;
-    }
+    const POINT2D *pt = getPoint2d_cp(node->geom->point, 0);
```
Algunenano: For the sake of consistency, could you move variable declarations to the top of their block? Same for https://github.com/bjornharrtell/postgis/blob/950bf4743c97b9f61b501914d4d36febb1de35cc/liblwgeom/lwgeom_topo.c#L3322 and https://github.com/bjornharrtell/postgis/blob/950bf4743c97b9f61b501914d4d36febb1de35cc/liblwgeom/lwgeom_topo.c#L5046

strk:
Agreed, variables declared at start of block, for ansi compatibility

strk:
> can it be explicit `\timing on` or `\timing off`?

Yes, but I think \timing should just go away, replaced with
timing done by run_test.pl itself (when invoked with --perf
or something similar)

Komzpa:
@bjornharrtell what's left to do announce this finished?

bjornharrtell:
@Komzpa - I view this as an experiment on which I wanted feedback on the method itself before iterating it to a finished state with a proper issue at Trac and NEWS addition. I also think it needs a wider set of performance tests to verify it's really faster, I'm not sure it is for edges with large amounts of vertices. I added bbox contains short circuit too but removed again as I could not measure any improvement, but might be relevant for other topologies.

Komzpa:
oh, then you should request feedback explicitly, or we're going to think you know what you are doing! :)

personally I like how many lines you have removed without breaking any test, that is usually a sign of good change.

Komzpa:
@bjornharrtell I was going to merge this but see that you have commit rights. Can you merge this then? :)

bjornharrtell:
Ok, will do in near future. :)

liblwgeom/lwgeom_topo.c
bjornharrtell: hmm but that means I have to drop const. :( oh well.

bjornharrtell: um.. no it doesn't :) C is wierd..

Komzpa:
Thank you!

PR #235, opened 2018-03-30
GEOSNode is tainted, retreat to using only GEOSUnion in MakeValid

Komzpa:
https://trac.osgeo.org/postgis/ticket/4061


dbaston:
Can you open a ticket for GEOS and reference it with a comment in the source? It's not obvious what "tainted" means.

PR #237, opened 2018-04-18
fix memory malloc and free not matching

hexiaoting:
The variable within drv_set and itself is using rtalloc() method to allocate memory, but using pfree() to free memory ,This is not true.

ps: And I want to try to be a contributor for project postgis, Can I? Thanks

### DIFF
--- a/raster/rt_pg/rtpostgis.c
+++ b/raster/rt_pg/rtpostgis.c
@@ -373,11 +373,11 @@ rtpg_assignHookGDALEnabledDrivers(const char *enabled_drivers, void *extra) {
 		}
 
 		for (i = 0; i < drv_count; i++) {
-			pfree(drv_set[i].short_name);
-			pfree(drv_set[i].long_name);
-			pfree(drv_set[i].create_options);
+			rtdealloc(drv_set[i].short_name);
+			rtdealloc(drv_set[i].long_name);
+			rtdealloc(drv_set[i].create_options);
 		}
-		if (drv_count) pfree(drv_set);
+		if (drv_count) rtdealloc(drv_set);
 
 	}
 


dustymugs:
There are handlers for memory management from the postgres level down to the librtcore. I had to remember the hops we go through to make everything wrap cleanly.

https://github.com/postgis/postgis/blob/f9e9411add7a32e4bdc907967c6664242a257ac4/raster/rt_pg/rtpostgisrt_set_handlers.c#L483

https://github.com/postgis/postgis/blob/bd3177743db5cb786deb05d7327f39a5cec2daf6/raster/rt_core/rt_context.c

https://github.com/postgis/postgis/blob/26cf6593a7be0c7e88af8129008fb63f5be11b7b/raster/rt_core/librtcore.h#L235

PS: You're contributing right now... ;-)

Komzpa:
@hexiaoting can you please add a NEWS entry on this with your preferred name spelling, and file a ticket to https://trac.osgeo.org/postgis/, so I can get this merged? :)

dustymugs:
This doesn't need to merged. Everything is working as expected given the memory handlers that ensure all memory functions redirect to the handlers (pgalloc vs malloc) . Code at the rtpg level intentially use the postgres memory functions.

hexiaoting:
Sorry, @dustymugs  This is not a bug after my investigation deeply(I'm developing on this rtpostgis module), Because it is alloc is reset in PG_init() function：
`rt_set_handlers(rt_pg_alloc, rt_pg_realloc, rt_pg_free, rt_pg_error, rt_pg_debug, rt_pg_notice); `

But, This is not suitable using rtalloc but using pfree rather then rtdealloc
So, I think this should stilled be merged. @Komzpa  How do you think?

PR #239, opened 2018-04-24
GEOS ClipByBox2D notice

Komzpa:


doc/reference_processing.xml
```diff
@@ -490,16 +490,16 @@ FROM (SELECT ST_Buffer(
 		<title>Description</title>
 
     <para>
-Clips a geometry by a 2D box in a fast but possibly dirty way. The output
-geometry is not guaranteed to be valid (self-intersections for a polygon
-may be introduced). Topologically invalid input geometries do not result
-in exceptions being thrown.
+Clips a geometry by a 2D box. The output geometry is not guaranteed to be valid
```
dbaston: Is this still accurate?

liblwgeom/lwgeom_geos.c
```diff
@@ -884,8 +884,14 @@ lwgeom_clip_by_rect(const LWGEOM *geom1, double x1, double y1, double x2, double
 {
 	LWGEOM *result;
 	LWGEOM *tmp;
-	LWGEOM *envelope = (LWGEOM*)lwpoly_construct_envelope(geom1->srid, x1, y1, x2, y2);
 
+	/* This lwgeom_intersection should be a call to GEOSClipByRect:
+	 * g3 = GEOSClipByRect(g1, x1, y1, x2, y2);
+	 * Unfortunately as of GEOS 3.7 it chokes on practical inputs.
```
dbaston: GEOS 3.7 is not released

doc/reference_processing.xml
Komzpa: it is not guaranteed but happens to be valid.

PR #240, opened 2018-05-08
 Add ST_BandFileSize and ST_BandFileTimestamp, extend ST_BandMetadata()

rouault:
Those additions to PostGISRaster enables the GDAL PostGISRaster client to check that a outdb filename returned by the server matches the file the client can access, which makes the GDAL PostGISRaster client able to get the raster content directly. See https://github.com/rouault/gdal/tree/pgraster_outdb for the GDAL branch implementing outdb support


robe2:
Confirmed 9.5 has a Int64GetDatum but no UInt64GetDatum

https://github.com/postgres/postgres/blob/REL9_5_STABLE/src/include/postgres.h#L621

Seems the UInt64GetDatum was introduced in 9.6

https://github.com/postgres/postgres/blob/REL9_6_STABLE/src/include/postgres.h#L653

PR #246, opened 2018-05-21
Address compiler warnings (-Wall -Wextra)

Algunenano:


Komzpa:
can we also get a -Wall -Wextra line in travis.yml? :)

PR #255, opened 2018-06-05
PostGIS requires GEOS 3.5+, remove conditinals for lower versions

Komzpa:


dbaston:
Thanks for removing the pre-3.5 cruft. I'd suggest leaving the unrelated formatting changes out of this PR, though.

PR #257, opened 2018-06-05
ST_PointOnSurface -> liblwgeom

Komzpa:


liblwgeom/lwgeom_geos.c
```diff
@@ -829,6 +829,55 @@ lwgeom_centroid(const LWGEOM* geom)
...
+	if (!input_lwgeom_to_geos(&g1, geom, __func__)) return NULL;
+
+	g3 = GEOSPointOnSurface(g1);
+
+	if (!g3)
+	{
+		GEOSGeometry *g1v;
+		lwnotice("%s: GEOS Error: %s", __func__, lwgeom_geos_errmsg);
+
+		if (!GEOSisValid(g1))
```
dbaston: It would really be good to complete the discussion on this issue with the project team before continuing to make changes like this.

liblwgeom/lwgeom_geos.c
```diff
@@ -543,7 +543,7 @@ output_geos_as_lwgeom(GEOSGeometry** g, LWGEOM** geom, const int32_t srid, const
 	return LW_TRUE;
 }
 
-/* Output encoder and sanity checker for GEOS wrappers */
+/* Clean up and return NULL */
 inline static LWGEOM*
 geos_clean_and_fail(GEOSGeometry* g1, GEOSGeometry* g2, GEOSGeometry* g3, const char* funcname)
```
dbaston: What do you think about making these macros? It would avoid the awkwardness of passing `__func__` around everywhere.

liblwgeom/lwgeom_geos.c
pramsey: Concur, we have long shied away from automagically fixing people's geometry.

PR #261, opened 2018-06-14
Explain units for DBSCAN

Komzpa:
Per request in IRC


dbaston:
It seems confusing to link to `ST_Distance` which is not used for distance calculations here.

Komzpa:
@dbaston do we have any better spot to link to, to explain what the units are? (without writing "units are map units" everywhere)

dbaston:
What do you think about the note on `ST_ClusterWithin` ?

robe2:
Are you guys ready to commit this so we can close this pull request?

PR #263, opened 2018-07-03
MVT combine: Fix value offset bug

Algunenano:
WIP for https://github.com/CartoDB/postgis/issues/20

I still need to find a way of test this automatically within Postgis

### DIFF
--- a/postgis/mvt.c
+++ b/postgis/mvt.c
@@ -1054,7 +1054,6 @@ vectortile_layer_combine(const VectorTile__Tile__Layer *layer1, const VectorTile
 	j = 0;
 	for (i = 0; i < layer1->n_features; i++)
 		layer->features[j++] = tile_feature_copy(layer1->features[i], 0, 0);
-	value2_offset = j;
 	for (i = 0; i < layer2->n_features; i++)
 		layer->features[j++] = tile_feature_copy(layer2->features[i], key2_offset, value2_offset);
 


pramsey:
Could (urg) expose some functions to read a two MVT bytea and return a combined MVT bytea. This is where having the MVT stuff ported into liblwgeom would help (I could pick up that old project and try to complete it, but it's ugly for other reasons and seems to necessarily involve extra copying. Something to talk about at the sprint.) 

Anyways, as long as MVT is in ./postgis, the only testing is going to be integration testing, so maybe a _postgis_mvt_combine(bytea,bytea) is the "best" way.

Algunenano:
> Anyways, as long as MVT is in ./postgis, the only testing is going to be integration testing, so maybe a _postgis_mvt_combine(bytea,bytea) is the "best" way.

I'm not a fan of making internal functions public so they can be tested but with what we have it does seem the only/best way to do it. I also abhor testing with bytea since the only thing that you are testing is that it doesn't change binary wise and to check it you (and whoever comes next) have to do it manually.

For the future, I'd like to have functions to work with MVTs (decode, combine, extract layers/features...) but it isn't an easy task and I don't know how useful they'd be apart from testing, as I don't expect people to store final MVTs in a column. Something like MVT as a geometry type.

Another option would be to have a different test framework for MVTs and use external tools to do the validation, which is basically automating what I've been doing manually.  This would either force us to bring dependencies in or build the framework out. 

This is certainly something that requires more thinking and that I don't expect to achieve in any case for 2.5 so I'll get this fix merged and create a trac issue so this can be considered in the future.

PR #264, opened 2018-07-05
Improve lwgeom_geos clarity

dbaston:
This PR attempts to clarify some of the code in `lwgeom_geos.c` by removing function calls that require non-obvious arguments (which are often `NULL` or `__func__`) and replacing them with calls to variadic macros/functions.

### DIFF
--- a/liblwgeom/lwgeom_geos.c
+++ b/liblwgeom/lwgeom_geos.c
@@ -575,6 +575,7 @@ lwgeom_normalize(const LWGEOM* geom)
 	if (!(g = LWGEOM2GEOS(geom, AUTOFIX))) GEOS_FAIL();
 
 	if (GEOSNormalize(g) == -1) GEOS_CLEAN_AND_FAIL(g);
+	GEOSSetSRID(g, srid);
 
 	if (!(result = GEOS2LWGEOM(g, is3d))) GEOS_CLEAN_AND_FAIL(g);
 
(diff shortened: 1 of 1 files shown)


liblwgeom/lwgeom_geos.c
```diff
@@ -495,92 +535,59 @@ lwgeom_geos_version()
...
-}
-
-/* Input decoder and sanity checker for GEOS wrappers */
-inline static uint8_t
-input_lwgeom_to_geos(GEOSGeometry** g, const LWGEOM* geom, const char* funcname)
+/* Return the consistent SRID of all inputs, or issue an
+ * error in case of SRID mismatch. Intended to be called
+ * from RESULT_SRID macro */
+static int32_t
+get_result_srid(int count, const char* funcname, ...)
```
Algunenano: Since `sizeof` returns size_t I'd rather have both `count` and `i` as size_t too.

pramsey:
I like it, 👍

dbaston:
@Komzpa do you want to review this or should I merge it in?

liblwgeom/lwgeom_geos.c
```diff
@@ -49,6 +51,44 @@ lwgeom_geos_error(const char* fmt, ...)
 	va_end(ap);
 }
 
+/* Destroy any non-null GEOSGeometry* pointers passed as arguments */
+#define GEOS_FREE(...) \
```
Komzpa: is "destroy" = "free"?

Komzpa:
@dbaston I'm ok with merge of this. Only thing I'm at doubt of is scope of current freeze and thus if this should go to 3.0 or 2.5.

dbaston:
Thanks for looking, @Komzpa . I'm inclined to target 2.5 so that we don't have three very different versions of this file in the wild.

robe2:
Okay go ahead :+1:

PR #265, opened 2018-07-05
WIP: Clean memory sanitizer warnings around GEOS code

Algunenano:
Shut up some of the warnings I get when running unit tests with -fsanitizer=memory

### DIFF
--- a/liblwgeom/lwprint.c
+++ b/liblwgeom/lwprint.c
@@ -378,6 +378,8 @@ static char * lwdouble_to_dms(double val, const char *pos_dir_symbol, const char
 
 	/* Allocate space for the result.  Leave plenty of room for excess digits, negative sign, etc.*/
 	result = (char*)lwalloc(format_length + WORK_SIZE);
+	memset(result, 0, format_length + WORK_SIZE);
+
 	/* Append all the pieces together. There may be less than 9, but in that case the rest will be blank. */
 	strcpy(result, pieces[0]);
 	for (index = 1; index < NUM_PIECES; index++)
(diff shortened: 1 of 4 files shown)


liblwgeom/lwprint.c
```diff
@@ -378,6 +378,8 @@ static char * lwdouble_to_dms(double val, const char *pos_dir_symbol, const char
 
 	/* Allocate space for the result.  Leave plenty of room for excess digits, negative sign, etc.*/
 	result = (char*)lwalloc(format_length + WORK_SIZE);
+	memset(result, 0, format_length + WORK_SIZE);
```
dbaston: Do you think we should add a `lwcalloc` to `lwgeom` ?

Algunenano: I've thought about that and it might help, but at the end of of the day if you don't use `memset` you probably won't use `lwcalloc` either.

Also, this case looks like a false positive as I don't see any way for the memory outside of the set string to be read, but since it's an output function and memset shouldn't any noticeable impact in performance I've decided to set it and make the sanitizer output clearer.

dbaston: ` grep -R -B1 "memset(" --include "*.c" | grep alloc | wc` shows 20 or so usages. Seems worth it to me.

Algunenano: Most of those seem uses to set simple structs, but I wouldn't oppose for it to be added in a followup issue.

Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/4118

PR #202, opened 2018-02-01
ST_Subdivide

Komzpa:
Let ST_Subdivide reselect pivot based on current piece geometry and not the original bbox. Split line for large polygons is also shifted to pass through existing point that lets the subdivision converge faster and chop off larger simple boxes.

This PR passes tests on simple cases.

One of unit tests (about inverted geometries) gets stuck on geos internals:

```
kom@junocat ~ % sudo gdb -ex "set pagination 0" -ex "thread apply all bt" --batch -p 17978
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
0x00007fe8a2b75d44 in geos::operation::intersection::distance(geos::operation::intersection::Rectangle const&, double, double, double, double) () from /usr/lib/x86_64-linux-gnu/libgeos-3.7.0dev.so

Thread 1 (Thread 0x7fe8b65ff780 (LWP 17978)):
#0  0x00007fe8a2b75d44 in geos::operation::intersection::distance(geos::operation::intersection::Rectangle const&, double, double, double, double) () from /usr/lib/x86_64-linux-gnu/libgeos-3.7.0dev.so
#1  0x00007fe8a2b76134 in geos::operation::intersection::distance(geos::operation::intersection::Rectangle const&, std::vector<geos::geom::Coordinate, std::allocator<geos::geom::Coordinate> > const&) () from /usr/lib/x86_64-linux-gnu/libgeos-3.7.0dev.so
#2  0x00007fe8a2b773e5 in geos::operation::intersection::RectangleIntersectionBuilder::reconnectPolygons(geos::operation::intersection::Rectangle const&) () from /usr/lib/x86_64-linux-gnu/libgeos-3.7.0dev.so
#3  0x00007fe8a2b74c90 in geos::operation::intersection::RectangleIntersection::clip_polygon_to_polygons(geos::geom::Polygon const*, geos::operation::intersection::RectangleIntersectionBuilder&, geos::operation::intersection::Rectangle const&) () from /usr/lib/x86_64-linux-gnu/libgeos-3.7.0dev.so
#4  0x00007fe8a2b75103 in geos::operation::intersection::RectangleIntersection::clip_geom(geos::geom::Geometry const*, geos::operation::intersection::RectangleIntersectionBuilder&, geos::operation::intersection::Rectangle const&, bool) () from /usr/lib/x86_64-linux-gnu/libgeos-3.7.0dev.so
[...]

PR #268, opened 2018-07-10
Remove geometry autofix behavior, for now

dbaston:
This PR reverts the geometry autofixing behavior that was added for 2.5 to avoid conflict with a comprehensive, documented approach in 3.0.

### DIFF
--- a/doc/reference_processing.xml
+++ b/doc/reference_processing.xml
@@ -490,16 +490,14 @@ FROM (SELECT ST_Buffer(
 		<title>Description</title>
 
     <para>
-Clips a geometry by a 2D box. The output geometry is not guaranteed to be valid
-(self-intersections for a polygon may be introduced).
-Topologically invalid input geometries do not result in exceptions being thrown.
-    </para>
+Clips a geometry by a 2D box.</para>
 
 		<para>Performed by the GEOS module.</para>
 		<note><para>Requires GEOS 3.5.0+</para></note>
 
 		<para>Availability: 2.2.0 - requires GEOS &gt;= 3.5.0.</para>
-		<para>Changed: 2.5.0 - wrapper around ST_Intersection to work around GEOS bugs. </para>
+                <para>Changed: 2.5.0 - wrapper around ST_Intersection to work around GEOS bugs.
+                      No longer supports invalid input geometry.</para>
 
 	  </refsection>
 
--- a/liblwgeom/lwgeom_geos.c
+++ b/liblwgeom/lwgeom_geos.c
@@ -618,24 +618,6 @@ lwgeom_intersection(const LWGEOM* geom1, const LWGEOM* geom2)
 
 	g3 = GEOSIntersection(g1, g2);
 
-	if (!g3)
-	{
-		GEOSGeometry *g1v, *g2v;
-		lwnotice("%s: GEOS Error: %s", __func__, lwgeom_geos_errmsg);
-
-		if (!GEOSisValid(g1) || !GEOSisValid(g2))
-		{
-			lwnotice(
-			    "Your geometry dataset is not valid per OGC Specification. "
-			    "Please fix it with manual review of entries that are not ST_IsValid(geom). "
-			    "Retrying GEOS operation with ST_MakeValid of your input.");
-			g1v = LWGEOM_GEOS_makeValid(g1);
-			g2v = LWGEOM_GEOS_makeValid(g2);
-			g3 = GEOSIntersection(g1v, g2v);
-			GEOS_FREE(g1v, g2v);
-		}
-	}
-
 	if (!g3) GEOS_FREE_AND_FAIL(g1);
 	GEOSSetSRID(g3, srid);
 
@@ -730,24 +712,6 @@ lwgeom_difference(const LWGEOM* geom1, const LWGEOM* geom2)
 
 	g3 = GEOSDifference(g1, g2);
 
-	if (!g3)
-	{
-		GEOSGeometry *g1v, *g2v;
-		lwnotice("%s: GEOS Error: %s", __func__, lwgeom_geos_errmsg);
-
-		if (!GEOSisValid(g1) || !GEOSisValid(g2))
-		{
-			lwnotice(
-			    "Your geometry dataset is not valid per OGC Specification. "
-			    "Please fix it with manual review of entries that are not ST_IsValid(geom). "
-			    "Retrying GEOS operation with ST_MakeValid of your input.");
-			g1v = LWGEOM_GEOS_makeValid(g1);
-			g2v = LWGEOM_GEOS_makeValid(g2);
-			g3 = GEOSDifference(g1v, g2v);
-			GEOS_FREE(g1v, g2v);
-		}
-	}
-
 	if (!g3) GEOS_FREE_AND_FAIL(g1, g2);
 	GEOSSetSRID(g3, srid);
 
@@ -780,24 +744,6 @@ lwgeom_symdifference(const LWGEOM* geom1, const LWGEOM* geom2)
 	if (!(g2 = LWGEOM2GEOS(geom2, AUTOFIX))) GEOS_FREE_AND_FAIL(g1);
 
 	g3 = GEOSSymDifference(g1, g2);
-
-	if (!g3)
-	{
-		GEOSGeometry *g1v, *g2v;
-		lwnotice("%s: GEOS Error: %s", __func__, lwgeom_geos_errmsg);
-
-		if (!GEOSisValid(g1) || !GEOSisValid(g2))
-		{
-			lwnotice(
-			    "Your geometry dataset is not valid per OGC Specification. "
-			    "Please fix it with manual review of entries that are not ST_IsValid(geom). "
-			    "Retrying GEOS operation with ST_MakeValid of your input.");
-			g1v = LWGEOM_GEOS_makeValid(g1);
-			g2v = LWGEOM_GEOS_makeValid(g2);
-			g3 = GEOSSymDifference(g1v, g2v);
-			GEOS_FREE(g1v, g2v);
-		}
-	}
 
 	if (!g3) GEOS_FREE_AND_FAIL(g1, g2);
 	GEOSSetSRID(g3, srid);
@@ -864,23 +810,6 @@ lwgeom_pointonsurface(const LWGEOM *geom)
 
 	g3 = GEOSPointOnSurface(g1);
 
-	if (!g3)
-	{
-		GEOSGeometry *g1v;
-		lwnotice("%s: GEOS Error: %s", __func__, lwgeom_geos_errmsg);
-
-		if (!GEOSisValid(g1))
-		{
-			lwnotice(
-			    "Your geometry dataset is not valid per OGC Specification. "
-			    "Please fix it with manual review of entries that are not ST_IsValid(geom). "
-			    "Retrying GEOS operation with ST_MakeValid of your input.");
-			g1v = LWGEOM_GEOS_makeValid(g1);
-			g3 = GEOSPointOnSurface(g1v);
-			GEOS_FREE(g1v);
-		}
-	}
-
 	if (!g3) GEOS_FREE_AND_FAIL(g1);
 	GEOSSetSRID(g3, srid);
 
@@ -914,24 +843,6 @@ lwgeom_union(const LWGEOM* geom1, const LWGEOM* geom2)
 	if (!(g2 = LWGEOM2GEOS(geom2, AUTOFIX))) GEOS_FREE_AND_FAIL(g1);
 
 	g3 = GEOSUnion(g1, g2);
-
-	if (!g3)
-	{
-		GEOSGeometry *g1v, *g2v;
-		lwnotice("%s: GEOS Error: %s", __func__, lwgeom_geos_errmsg);
-
-		if (!GEOSisValid(g1) || !GEOSisValid(g2))
-		{
-			lwnotice(
-			    "Your geometry dataset is not valid per OGC Specification. "
-			    "Please fix it with manual review of entries that are not ST_IsValid(geom). "
-			    "Retrying GEOS operation with ST_MakeValid of your input.");
-			g1v = LWGEOM_GEOS_makeValid(g1);
-			g2v = LWGEOM_GEOS_makeValid(g2);
-			g3 = GEOSUnion(g1v, g2v);
-			GEOS_FREE(g1v, g2v);
-		}
-	}
 
 	if (!g3) GEOS_FREE_AND_FAIL(g1, g2);
 	GEOSSetSRID(g3, srid);
--- a/regress/clipbybox2d_expected
+++ b/regress/clipbybox2d_expected
@@ -2,11 +2,7 @@
 2|BOX(5 5,8 8)
 3|BOX(2 2,8 8)
 4|MULTIPOINT(0 0,2 2)
-NOTICE:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
-NOTICE:  Self-intersection
-NOTICE:  Your geometry dataset is not valid per OGC Specification. Please fix it with manual review of entries that are not ST_IsValid(geom). Retrying GEOS operation with ST_MakeValid of your input.
-NOTICE:  Self-intersection
-5|MULTIPOLYGON(((2 2,5 5,8 2,2 2)),((5 5,2 8,8 8,5 5)))
+ERROR:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
 6|MULTIPOLYGON(((2.5 2,5 4,5 5,10 5,10 2,2.5 2)))
 7|POLYGON((2 2,2 5,5 5,5 2,2 2))
 8|SRID=3857;POLYGON EMPTY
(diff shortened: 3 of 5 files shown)


regress/clipbybox2d_expected
```diff
@@ -2,11 +2,7 @@
 2|BOX(5 5,8 8)
 3|BOX(2 2,8 8)
 4|MULTIPOINT(0 0,2 2)
-NOTICE:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
-NOTICE:  Self-intersection
-NOTICE:  Your geometry dataset is not valid per OGC Specification. Please fix it with manual review of entries that are not ST_IsValid(geom). Retrying GEOS operation with ST_MakeValid of your input.
-NOTICE:  Self-intersection
-5|MULTIPOLYGON(((2 2,5 5,8 2,2 2)),((5 5,2 8,8 8,5 5)))
+ERROR:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
```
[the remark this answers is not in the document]
dbaston: If I recall, those guarantees were broken by earlier and unrelated work on `ST_Subdivide`

pramsey:
Without looking at the code (I know! I suck!) but just at the test cases exposed here, can we catch the exceptions and run the isvalidreason on them for a nicer output, ala the current approach, but without automatically fixing things?

pramsey:
Actually, from a "user practicality" point of view, one of the issues the old "toss exception" behaviour has is that the user cannot easily figure out precisely what features are at issue. I guess maybe the isvalid reason w/ coordinates would be a quick'n'dirty path to find the approximate areas of concern.

dbaston:
>  can we catch the exceptions and run the isvalidreason on them for a nicer output

I did this initially, but we end up with duplicated error messages, unless we decide to swallow the original error message from GEOS. After a failure, GEOS is already checking the validity of its inputs. Maybe it should be modified to include the coordinate in the exception message?

regress/clipbybox2d_expected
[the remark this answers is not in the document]
Komzpa: Work on `ST_Subdivide` was a bugfix of box clipping that shot down our processing, and was the reason for the need to MakeValid things in lwgeom_intersection :)

dbaston:
> Maybe it should be modified to include the coordinate in the exception message?

It already does this, duh. We just don't include it in the `_expected` files (as we shouldn't)

regress/clipbybox2d_expected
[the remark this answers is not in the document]
dbaston: Updated docs.

regress/tickets_expected
```diff
@@ -326,31 +326,11 @@ NOTICE:  Too few points in geometry component at or near point 0 0
...
-NOTICE:  lwgeom_symdifference: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
-NOTICE:  Self-intersection
-NOTICE:  Your geometry dataset is not valid per OGC Specification. Please fix it with manual review of entries that are not ST_IsValid(geom). Retrying GEOS operation with ST_MakeValid of your input.
-NOTICE:  Self-intersection
-#4037.3|MULTIPOLYGON(((0 0,2 2,8 2,10 0,0 0)),((5 5,2 2,2 8,5 5)),((8 2,5 5,8 8,8 2)),((2 8,0 10,10 10,8 8,2 8)))
-NOTICE:  lwgeom_union: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
-NOTICE:  Self-intersection
-NOTICE:  Your geometry dataset is not valid per OGC Specification. Please fix it with manual review of entries that are not ST_IsValid(geom). Retrying GEOS operation with ST_MakeValid of your input.
-NOTICE:  Self-intersection
-#4037.4|POLYGON((0 0,2 2,2 8,0 10,10 10,8 8,8 2,10 0,0 0))
```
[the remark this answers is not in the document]
dbaston: Updated `NEWS`.

regress/clipbybox2d_expected
```diff
@@ -2,11 +2,7 @@
...
 6|MULTIPOLYGON(((2.5 2,5 4,5 5,10 5,10 2,2.5 2)))
```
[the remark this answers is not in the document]
dbaston: This change is necessary because `ST_ClipByBox2D` now delegates to `ST_Intersection`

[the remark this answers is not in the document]
Komzpa: Here, input is invalid, and ST_Intersection does not barf on it (and produces a counter-intuitive geometry).

liblwgeom/lwgeom_geos.c
```diff
@@ -781,24 +745,6 @@ lwgeom_symdifference(const LWGEOM* geom1, const LWGEOM* geom2)
 
 	g3 = GEOSSymDifference(g1, g2);
 
-	if (!g3)
```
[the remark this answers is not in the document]
dbaston: I think you're referring to line ~1488? I don't think there is a possibility for the noding to cause incorrect results. Is there?

doc/reference_processing.xml
```diff
@@ -490,16 +490,14 @@ FROM (SELECT ST_Buffer(
...
-    </para>
+Clips a geometry by a 2D box.</para>
 
 		<para>Performed by the GEOS module.</para>
 		<note><para>Requires GEOS 3.5.0+</para></note>
 
 		<para>Availability: 2.2.0 - requires GEOS &gt;= 3.5.0.</para>
-		<para>Changed: 2.5.0 - wrapper around ST_Intersection to work around GEOS bugs. </para>
+                <para>Changed: 2.5.0 - wrapper around ST_Intersection to work around GEOS bugs.
+                      No longer supports invalid input geometry.</para>
```
Komzpa: This change breaks MVT output in cases simplified but not yet clipped geometry is invalid.

regress/clipbybox2d_expected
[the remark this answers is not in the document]
dbaston: Sorry, I'm having trouble understanding the changes you're requesting.

doc/reference_processing.xml
dbaston: I guess MVT code could be modified to check validity on intersection failure, and then run makevalid if needed?

regress/clipbybox2d_expected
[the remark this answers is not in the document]
Algunenano: Sorry I didn't notice the ping, I'll try to have a look, possibly on Monday.
TBH, the more I use MVTs the more I wish it would only do coordinate transformations (or all the operations were optional and configurable).

doc/reference_processing.xml
Algunenano: Recently I've been comparing the outputs of using St_AsMVT and Mapnik ([node-mapnik](https://github.com/mapnik/node-mapnik) + [mapnik-vector-tile](https://github.com/mapbox/mapnik-vector-tile)) with the same geometries as input. Here is an [image](https://user-images.githubusercontent.com/664253/41592513-15865562-73bd-11e8-920e-98c21b0b0c9c.png) with some of the tests I've run, or the WIP PR with the [geometries tested](https://github.com/Algunenano/Windshaft/blob/74ab2faa75341d51eb98533180b3a677d61f4c3e/test/acceptance/mvt.js#L649).

As you can see there, Mapnik drops the geometry when in finds a self-intersection, so the equivalent here would be ST_AsMVTGeom returning NULL. Now, for reasons I don't want to discuss right now (but probably will in the future) it was decided that our implementation was going to autofix geometries and that's what it has been doing up until now.

Take this query as an example:
```sql
SELECT 'PG43 - ON ', ST_AsText(ST_AsMVTGeom(
	ST_GeomFromText('POLYGON((-1 -1, 101 101, -1 101, 101 -1, -1 -1))'),
	ST_MakeBox2D(ST_Point(0, 0), ST_Point(100, 100)),
	10, 0, true));
```

With the current trunk implementation you get some NOTICE messages and the autofixed geometry:
```
NOTICE:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection at or near point 50 50 at 50 50
NOTICE:  Self-intersection at or near point 50 50
NOTICE:  Your geometry dataset is not valid per OGC Specification. Please fix it with manual review of entries that are not ST_IsValid(geom). Retrying GEOS operation with ST_MakeValid of your input.
NOTICE:  Self-intersection at or near point 50 50
  ?column?  |                         st_astext                          
------------+------------------------------------------------------------
 PG43 - ON  | MULTIPOLYGON(((0 10,5 5,10 10,0 10)),((5 5,0 0,10 0,5 5)))
```

With the changes proposed in this PR you get an error and the query is aborted:
```
ERROR:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection at or near point 50 50 at 50 50
template_postgis=#
```

Mainly for performance optimization, I would like St_AsMVTGeom to be more customizable, which in this case it'd mean that you could either choose to autofix the geometry or drop it ASAP, but I won't push for that in 2.5. For 2.5, I can live with either ST_AsMVTGeom returning NULL (better performance and conformance with Mapnik's implementation) or with it autofixing the geometry (current implementation), but I certainly cannot live with it throwing an exception and killing the query.


> I guess MVT code could be modified to check validity on intersection failure, and then run makevalid if needed? Test case would be great if you have one.

I'm not very familiar with the geos code, but since you are calling `GEOS_FREE_AND_FAIL` which ends up calling `lwerror`, the mvt function doesn't regain control of the program so you can't check for an error and run `lwgeom_make_valid`. The only option would be to run it before clipping but I'm not sure how big of a performance hit that would mean for valid geometries. I'll try to have a look and make some perf tests.

Some extra notes:
- In the process of looking into this issue I've noticed that in some corner cases we could avoid the clipping if we recalculate the bbox of the geometry after the simplification steps, but that recalculation isn't worth it in the general case. It's true that forcing validation before clipping would regenerate it, so there is that.,

- I've also noticed that https://github.com/postgis/postgis/pull/267 can be used too for buffer == 0. I'll add a bunch of tests to trunk but that doesn't affect this issue.

dbaston: Hi @Algunenano , thank you for the analysis. I see two options here:

- You're right that we can't call `lwgeom_intersection` in the MVT code, but we could call `GEOSIntersection` directly and handle the error ourselves (drop or autofix, I have no opinion).
- Alternatively, we could make `lwgeom_clip_by_rect` again call `GEOSClipByRect` (so it would no longer fail on invalid geometry), and then modify `ST_Subdivide` to call `lwgeom_intersection` instead of `lwgeom_clip_by_rect`. @Komzpa if I remember correctly, your changes to `lwgeom_clip_by_rect` were motivated by a `ST_Subdivide` issue, right?

Algunenano: So I've gone ahead and test the changes forcing validation:
```c
		if (!gbox_contains_2d(&bgbox, lwgeom_gbox))
		{
			if (lwgeom->type == POLYGONTYPE ||
				lwgeom->type == MULTIPOLYGONTYPE ||
				lwgeom->type == COLLECTIONTYPE)
			{
				// TODO: Handle memory and errors properly
				lwgeom = lwgeom_make_valid(lwgeom);
			}
			double x0 = bgbox.xmin;
			double y0 = bgbox.ymin;
			double x1 = bgbox.xmax;
			double y1 = bgbox.ymax;
			lwgeom = lwgeom_clip_by_rect(lwgeom, x0, y0, x1, y1);
			POSTGIS_DEBUG(3, "mvt_geom: no geometry after clip");
			if (lwgeom == NULL || lwgeom_is_empty(lwgeom))
				return NULL;
		}
```

I don't have a dataset with invalid geometries to test, but TBH it's not the test case I'd want to optimize for; so I have compare the performance using a dataset of valid multipolygons:
[mvt_pg_trunk_vs_mvt_pg_geos.pdf](https://github.com/postgis/postgis/files/2197904/mvt_pg_trunk_vs_mvt_pg_geos.pdf)

In the first tests there is too big of a variance to extract any useful information based on the number of runs done, but from there on the version without the autofix looks consistently 5-10% faster. Based on that, I'm fine for now with adding the `lwgeom_make_valid` call and removing the autofix behavior. It needs some work as there are now 2 calls to `lwgeom_make_valid` which creates a copy of the geometry and I prefer cleaning it instead of leaving it to Postgresql memory management, but it should be easy to do.

dbaston: @Algunenano I'd be curious if there is a  performance impact to removing `lwgeom_make_valid` and changing `lwgeom_clip_by_rect` to again call `GEOSClipByRect`.

dbaston: @Komzpa The performance impact seems worth knowing in case of a 2.4 to 2.5 regression, but @Algunenano would know better than me if that's worth testing (I'm not going to duplicate his benchmark process). FWIW I can't replicate the "endless loop" issue (which would be severe).

Algunenano: > @Algunenano I'd be curious if there is a performance impact to removing lwgeom_make_valid and changing lwgeom_clip_by_rect to again call GEOSClipByRect.

I'll give it a try. I'll have to dig in history to retrieve it, so if you have a branch with it restored on trunk it'll save me time

Algunenano: Switching to the old build hangs the subdivide tests. Nonetheless I'll run the benchmark to see time difference in the implementations. Changes can be seen [here](https://github.com/Algunenano/postgis/commit/4fe67ddf9422d4f2ec7083170a818b36395b58749)

dbaston: @Algunenano that link doesn't work for me. Here's what I arrived at: [link](https://github.com/dbaston/postgis/commit/ec2b48462dc23dd12ac89b363d29b3da8d5de21e)

Algunenano: Sorry, it had an extra digit. [Here it is](https://github.com/Algunenano/postgis/commit/4fe67ddf9422d4f2ec7083170a818b36395b5874) but it's pretty much the same.

Algunenano: Here is the comparison using this PR + GEOSClipByRect:
* Vs current trunk ([mvt_pg_trunk_vs_mvt_old_subdivide.pdf](https://github.com/postgis/postgis/files/2198350/mvt_pg_trunk_vs_mvt_old_subdivide.pdf)): Using `GEOSClipByRect` is between 0-20% better in MVT generation
* Vs this PR + make_valid before clipping ([mvt_pg_geos_vs_mvt_old_subdivide.pdf](https://github.com/postgis/postgis/files/2198355/mvt_pg_geos_vs_mvt_old_subdivide.pdf)): Depends on the test. Seems that when there is more simplification beforehand, this PR is faster; but when there isn't (bigger geometries after the simplify steps), using GEOSClipByRect is faster.

Since the subdivide tests hang (at least on my PC using geos master) I think we shouldn't revert to GEOSClipByRect.

dbaston: @Algunenano Thanks again for this. Based on what you're saying, I would suggest to consider my second option: make `lwgeom_clip_by_rect` again call `GEOSClipByRect`, and then modify `ST_Subdivide` to call `lwgeom_intersection` (so that we don't get the infinite loop @Komzpa discovered.) The net change from 2.4 to 2.5 would then be:

- MVT: nothing on this front
- ST_Subdivide: new algorithm, plus switched from using `GEOSClipByRect` to a generic intersection (which is IMO more appropriate; I think the caveats listed with `GEOSClipByRect` make it less suitable for the point-in-polygon queries often associated with `ST_Subdivide`).

dbaston: @Komzpa Don't we avoid an endless loop and still incorporate your `ST_Subdivide` algorithm changes with my proposal? We also avoid a 2.4 -> 2.5 change in the semantics of `ST_ClipByBox2D`.

Komzpa: Endless loop is in GEOSClipByRect, stacktrace is in https://github.com/postgis/postgis/pull/202#issue-166583659 - endless loop is in principle reachable from ST_ClipByBox2D or ST_AsMVT, just needs some patience in building a failing case.

dbaston: OK, let's pause discussion until we have a test case.

Algunenano: My understanding is that this PR (plus tweaks in MVT to avoid exceptions) improves performance, but I'm not sure why. Also, the performance of GEOSClipByRect vs lwgeom_intersection seems equivalent but GEOSClipByRect has a known bug, so I'd rather have make_valid (either in MVT or at the start of lwgeom_clip_by_rect if you want to keep the old guarantees and avoid multiple transformations).

pramsey:
None from me. I would like to leave behaviour as is and address this globally in Boston.

Algunenano:
> Objections to merging this PR as-is?

Ok for me, I agree this should be discussed so leaving the behaviour as it was is the way to go for now. As mentioned before, this will break MVTs (I've already added a test to check it, so expect broken builds) so, when you merge this, please open a issue and I'll have a look to either call make_valid before `lwgeom_clip_by_rect` or implement a handle to catch this error and retry with validation (whatever ends up being faster in normal path).

Komzpa:
@Algunenano also, if ST_MakeValid is added inside `lwgeom_clip_by_rect` at the start, that will keep "works on invalid" semantics for it.

Algunenano:
> @Algunenano also, if ST_MakeValid is added inside lwgeom_clip_by_rect at the start, that will keep "works on invalid" semantics for it.

Yes, but I was getting ahead of myself and thinking about avoiding the double conversion to geos (one in `lwgeom_make_valid` and another in `lwgeom_intersection`), but I'll do that in a different PR.

PR #269, opened 2018-07-11
Fix SP-GiST warnings on PG11

Komzpa:
references https://trac.osgeo.org/postgis/ticket/1847

doc will follow.


dbaston:
A bit difficult to review this with all of the formatting changes.

PR #270, opened 2018-07-13
Fix postgis_tiger_geocoder.control generation

ChristophBerg:
Commit 639308e94df7b77c42e3ef73798d277fa80c971b moved the
postgis_tiger_geocoder.control into the Makefile, but forgot to remove
the file from configure.ac, which still generates the file, so the
Makefile rule is not effective and the resulting .control file has the
raw @EXTVERSION@ line from the .in file.

Cc: @strk


Komzpa:
Committed, thank you! :)

PR #272, opened 2018-07-16
ST_AsMVTGeom: Clip using tile coordinates also for buffer 0

Algunenano:
Does several things:
- Avoid undefined behaviour when extent == 0, which meant that the check could be removed by the compiler.
- Clips using tile coordinates also for buffer == 0.
- Adds some more extra tests for clipping.
- Adds a test with an invalid polygon so issues as the one being discussed in https://github.com/postgis/postgis/pull/268 are reflected somewhere in the tests.

### DIFF
--- a/postgis/mvt.c
+++ b/postgis/mvt.c
@@ -696,11 +696,7 @@ LWGEOM *mvt_geom(LWGEOM *lwgeom, const GBOX *gbox, uint32_t extent, uint32_t buf
 	gridspec grid;
 	double width = gbox->xmax - gbox->xmin;
 	double height = gbox->ymax - gbox->ymin;
-	double resx = width / extent;
-	double resy = height / extent;
-	double res = (resx < resy ? resx : resy)/2;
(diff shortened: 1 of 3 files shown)


postgis/mvt.c
```diff
@@ -726,8 +728,7 @@ LWGEOM *mvt_geom(LWGEOM *lwgeom, const GBOX *gbox, uint32_t extent, uint32_t buf
 	{
 		// We need to add an extra half pixel to include the points that
 		// fall into the bbox only after the coordinate transformation
-		double buffer_map_xunits = !buffer ?
-			0.0 : nextafterf(resx * (buffer + 0.5), 0.0);
+		double buffer_map_xunits = nextafterf(res, 0.0) + resx * buffer;
 		GBOX bgbox;
 		const GBOX *lwgeom_gbox = lwgeom_get_bbox(lwgeom);;
```
Komzpa: ;;

PR #275, opened 2018-07-21
Provide a way to opt out from autofix

Komzpa:
In https://trac.osgeo.org/postgis/ticket/4040 there is a request to provide a way that works for users that are concerned with data integrity.

In https://github.com/postgis/postgis/pull/268 it is expected that data integrity can be preserved by reverting the change. However, it's not the case: GEOS can produce an incorrect result from overlay operation on invalid geometries without throwing any exception. Example:

```
SELECT ST_AsGeoJSON(ST_Intersection('POLYGON((0 0,5 4,5 6,0 10,10 10,5 6,5 4,10 0,0 0))', ST_MakeEnvelope(2,2,10,5)));
------
{"type":"GeometryCollection","geometries":[{"type":"LineString","coordinates":[[5,5],[5,4],[7.5,2]]},{"type":"Polygon","coordinates":[[[2.5,2],[5,4],[5,5],[10,5],[10,2],[7.5,2],[2.5,2]]]}]} 
```
input:
image

result:
image

To satisfy people who are concerned with data integrity, we can provide them with a way of replacing the functions with their versions checking the inputs. Then, even with this example, it feels like "old" behavior and provides some guarantees:

```
01:55:27 [kom] > SELECT ST_AsGeoJSON(ST_Intersection('POLYGON((0 0,5 4,5 6,0 10,10 10,5 6,5 4,10 0,0 0))', ST_MakeEnvelope(2,2,10,5)));
ERROR:  P0001: Input geometry of type ST_Polygon with 9 points is invalid at location POINT(5 4): Self-intersection
CONTEXT:  PL/pgSQL function _st_failoninvalid(geometry) line 7 at RAISE
SQL function "st_intersection" statement 1
LOCATION:  exec_stmt_raise, pl_exec.c:3725
Time: 1,514 ms
```

### DIFF
--- a/postgis/harden.sql
+++ b/postgis/harden.sql
@@ -1,0 +1,64 @@
+-- This file replaces standard PostGIS functions with their versions
+-- actively checking input to be valid.
+
+begin;
+create or replace function _ST_FailOnInvalid(geom geometry)
+returns boolean as
+$$
+declare
+	v valid_detail;
+begin
+	v = ST_IsValidDetail(geom);
+	if (not v.valid) then
+		raise exception 'Input geometry of type % with % points is invalid at location %: %',
+		ST_GeometryType(geom), ST_NPoints(geom), ST_AsText(v.location), v.reason;
+	end if;
+	return true;
+end
+$$
+language plpgsql volatile strict parallel safe cost 1;
+
+alter function ST_Intersection(geometry, geometry) rename to ST_Intersection_real;
+create function ST_Intersection(geom1 geometry, geom2 geometry)
+returns geometry
+as $$
+select ST_Intersection_real(geom1, geom2) where _ST_FailOnInvalid(geom1) and _ST_FailOnInvalid(geom2)
+$$ language sql parallel safe;
+
+alter function ST_Union(geometry, geometry) rename to ST_Union_real;
+create function ST_Union(geom1 geometry, geom2 geometry)
+returns geometry
+as $$
+select ST_Union_real(geom1, geom2) where _ST_FailOnInvalid(geom1) and _ST_FailOnInvalid(geom2)
+$$ language sql parallel safe;
+
+alter function ST_Difference(geometry, geometry) rename to ST_Difference_real;
+create function ST_Difference(geom1 geometry, geom2 geometry)
+returns geometry
+as $$
+select ST_Difference_real(geom1, geom2) where _ST_FailOnInvalid(geom1) and _ST_FailOnInvalid(geom2)
+$$ language sql parallel safe;
+
+alter function ST_SymDifference(geometry, geometry) rename to ST_SymDifference_real;
+create function ST_SymDifference(geom1 geometry, geom2 geometry)
+returns geometry
+as $$
+select ST_SymDifference_real(geom1, geom2) where _ST_FailOnInvalid(geom1) and _ST_FailOnInvalid(geom2)
+$$ language sql parallel safe;
+
+alter function ST_Buffer(geometry, float) rename to ST_Buffer_real;
+create function ST_Buffer(geom1 geometry, r float)
+returns geometry
+as $$
+select ST_Buffer_real(geom1, r) where _ST_FailOnInvalid(geom1)
+$$ language sql parallel safe;
+
+alter function ST_PointOnSurface(geometry) rename to ST_PointOnSurface_real;
+create function ST_PointOnSurface(geom1 geometry)
+returns geometry
+as $$
+select ST_PointOnSurface_real(geom1) where _ST_FailOnInvalid(geom1)
+$$ language sql parallel safe;
+
+commit;
+


robe2:
Only issue I have with this is that I don't think it will play well with extension upgrade.

What I am forseeing, you install PostGIS 2.5.0, then you install the opt out script.

then you upgrade to 2.5.1, you are back to same behavior (non-opt out) which is fine,
but then I think the renamed functions are now part of 2.5.0 and the old ones are not so maybe you can't upgrade or upgrade just skips over them or brings them in as part of extension.

Komzpa:
Standard update is blocked with this script:

```sql
16:08:14 [kom] > alter extension postgis update TO "2.5.0beta1next";
ERROR:  42P13: cannot change name of input parameter "geom1"
HINT:  Use DROP FUNCTION st_buffer(geometry,double precision) first.
LOCATION:  ProcedureCreate, pg_proc.c:481
Time: 76,817 ms
```

This means harden_uninstall.sql should be provided, so that update procedure is:
 - try normal upgrade, see it failing, read the manual;
 - uninstall the checking versions of functions;
 - upgrade;
 - reinstall script.

Re: auto-fix speed, it is as fast as non-fixing version except for cases where the fix is required. The concern expressed in the ticket is about data integrity, not speed, so to provide certain level of data integrity guarantees one should either check geometries upon creation themselves (and then presence of autofix will not bother them), or check upon every operation. Checking `TopologyException` just isn't the right way to check for correctness of operation.

There is observed speed difference in ST_ClipByBox2D, but it has some more clean-up logic when wired to lwgeom_intersection, so this is expected and not about auto-fix.

Komzpa:
The way I see this is the following:


  | Autofix on exception | Autofix on entry | Failure on exception | Failure on entry
-- | -- | -- | -- | --
  | PostGIS 2.5 | PostGIS 3.0? | PostGIS 2.4 | harden.sql
Data Integrity when user takes care of inputs | v | v | v | v
Data Integrity when MakeValid is correct way of interpreting invalid | x | v | x | v
Strict Data Integrity when MakeValid does not catch nuances of data | x | x | x | v
Workflow smoothness when user MakeValid is OK | v | v | x | x
Workflow smoothness when user takes care of input | x | v | x | v
Query is fast on valid input | v | x | v | x
Total v | 3 | 4 | 2 | 4

PostGIS 2.4 way of dealing with invalids (Failure on exception) is neither guaranteeing a failure on invalid calculation nor is convenient when user agrees to any input interpretation.

PostGIS 2.5.0beta1 (Autofix on exception) fixed workflow smoothness for people who are OK with MakeValid's input interpretation.

harden.sql from this PR (Failure on entry) provides strict guarantees on any PostGIS version, but leaves the user with broken geometry and does not give a good way of fixing.

postgis/harden.sql
```diff
@@ -0,0 +1,64 @@
...
+begin
+	v = ST_IsValidDetail(geom);
+	if (not v.valid) then
+		raise exception 'Input geometry of type % with % points is invalid at location %: %',
+		ST_GeometryType(geom), ST_NPoints(geom), ST_AsText(v.location), v.reason;
+	end if;
+	return true;
+end
+$$
+language plpgsql volatile strict parallel safe cost 1;
```
Algunenano: I think this should be `IMMUTABLE`, not `VOLATILE`. For example, in a query like this:
```sql
Select * from my_table where ST_Intersection(geom, ST_MakeEnvelope(10, 10, 11, 11, 4326))
```
You want the envelope to be checked just once, not once per row.

The rest of the functions need also the rest of the properties set: `IMMUTABLE STRICT COST 2048`, and unless we drop support for releases prior to 9.6, the parallel property needs to be set in a configure step.

Algunenano:
I'm not sure about this.

On one hand it's nice that the functions just substitute the existing ones, so it someone has other functions or scripts using them and want to have a strict run they can just use this file; on the other hand, due to its performance impact, I probably would only use this on a very special occasion  so I wouldn't want to substitute the functions forever, just for that run and the go back which doesn't look a straightaway operation (load harden, run query, unload harden).

As a similar idea, we could have a GUC (POSTGIS_HARDEN ?) that changes the `lwgeom_from_gserialized` to both deserialize and check for validity if enabled, throwing an exception it isn't valid. The main issue I see there is the performance impact it would have, so I'm thinking it should be flagged so the `geometry_serialize` would set it on output. Does that sound sane? Probably not sane enough for 2.5.

robe2:
GUCs are out for PostGIS 2.5, too late in release cycle to add those.  PostGIS 3.0 maybe

dbaston:
I can see configuring the behavior with a GUC and/or function arguments. Loading and unloading a special file that overwrites core PostGIS functions doesn't strike me as a good user experience.

postgis/harden.sql
Komzpa: It also seems to be not parallel safe, "cannot start commands in parallel scan" says postgres.

PR #277, opened 2018-07-23
Wkt enforce whitespace

Algunenano:
Fix for https://trac.osgeo.org/postgis/ticket/4109

It'd great if someone with experience with Bison could have a look.


Komzpa:
If it does not break an existing test, I'd say we merge this.
(WKT input is also tested via oss-fuzz so a major crash, if any, will be found.)

pramsey:
I find the solution breathtakingly ugly so I'm looking for a better way...

Algunenano:
> I find the solution breathtakingly ugly so I'm looking for a better way...

I'm not going to disagree with you on this one. It is extremely ugly :japanese_ogre:

pramsey:
I think the following two-liner is sufficient to fix the issue:
```
--?(([0-9]+\.?)|([0-9]*\.?[0-9]+)([eE][-+]?[0-9]+)?) {  
+-?(([0-9]+\.?)|([0-9]*\.?[0-9]+)([eE][-+]?[0-9]+)?)[ \,\)] {
       LWDEBUG(5,"DOUBLE");
       wkt_yylval.doublevalue = atof(wkt_yytext);
+      yyless(wkt_yyleng-1);
       return DOUBLE_TOK;
       }
```
The pattern change insists that all doubles must be anchored at the end by a space, comma or close bracket. The call to `yyless` puts that anchor back into the input stream so that it can subsequently be picked up by things like the comma and bracket patterns.

Algunenano:
> The pattern change insists that all doubles must be anchored at the end by a space, comma or close bracket. The call to yyless puts that anchor back into the input stream so that it can subsequently be picked up by things like the comma and bracket patterns.

What about multiple spaces, tabs or CRs? I'll try to add them with a few tests

pramsey:
I'm not sure how much more lenient we want to be WRT CRs. Tabs maybe. Multiples should end up being eaten by the whitespace rule at the bottom, so don't add them to the double pattern.

Algunenano:
I've added single `\t` `\n` `\r`. This might still not be enough for Windows (`\r\n`); is it worth it to address it?

pramsey:
I don't think so. If your WKT is
```
POINT(123.
4123)
```
you're SOL, sorry.

pramsey:
👍

PR #279, opened 2018-07-26
Revert to old ST_ClipByBox2d behaviour with invalids

Algunenano:
After some tests with MVTs I didn't see any significant performance difference between doing the validation before calling `lwgeom_clip_by_rect` or implementing my own call to GEOSIntersection with error handling, probably because the input was always correct and not so many of the geometries needed clipping (most fall inside the tile completely), so I decided to kill 2 birds and adapt the call directly in `lwgeom_clip_by_rect` so we get the old behaviour (validity tolerance) in 2.5.

This should fix https://trac.osgeo.org/postgis/ticket/4135 and https://trac.osgeo.org/postgis/ticket/4134.

BTW, here are some of the stuff I tested in case you are curious:
- Implementation in MVT_geom: https://github.com/Algunenano/postgis/blob/bb8d4512f5edc9813130ce6064f98db1fb27c109/postgis/mvt.c#L744
- Implementation forcing validation with memory handling: https://github.com/Algunenano/postgis/blob/1085acd64ac84e86ca1a7e6f377e9b796eab174b/postgis/mvt.c#L742
- Without memory handling: https://github.com/Algunenano/postgis/blob/33c39109bc2f16553daa8304f4e2f26dd1096598/postgis/mvt.c#L747

### DIFF
--- a/doc/reference_processing.xml
+++ b/doc/reference_processing.xml
@@ -490,14 +490,16 @@ FROM (SELECT ST_Buffer(
 		<title>Description</title>
 
     <para>
-Clips a geometry by a 2D box.</para>
+Clips a geometry by a 2D box. The output geometry is not guaranteed to be valid
+(self-intersections for a polygon may be introduced).
+Topologically invalid input geometries do not result in exceptions being thrown.
+    </para>
 
 		<para>Performed by the GEOS module.</para>
 		<note><para>Requires GEOS 3.5.0+</para></note>
 
 		<para>Availability: 2.2.0 - requires GEOS &gt;= 3.5.0.</para>
-                <para>Changed: 2.5.0 - wrapper around ST_Intersection to work around GEOS bugs.
-                      No longer supports invalid input geometry.</para>
+		<para>Changed: 2.5.0 - wrapper around ST_Intersection to work around GEOS bugs. </para>
 
 	  </refsection>
 
--- a/liblwgeom/cunit/cu_clip_by_rect.c
+++ b/liblwgeom/cunit/cu_clip_by_rect.c
@@ -64,6 +64,15 @@ static void test_lwgeom_clip_by_rect(void)
 	//tmp = lwgeom_to_ewkt(out); printf("%s\n", tmp); lwfree(tmp);
 	CU_ASSERT(lwgeom_is_empty(out));
 	lwgeom_free(out); lwgeom_free(in);
+
+	/* Invalid polygon (lineal self-intersection) */
+	wkt = "POLYGON((0 0, 10 10, 0 10, 10 0, 0 0))";
+	in = lwgeom_from_wkt(wkt, LW_PARSER_CHECK_NONE);
+	out = lwgeom_clip_by_rect(in, 2, 2, 8, 8);
+	tmp = lwgeom_to_ewkt(out);
+	printf("%s\n", tmp);
+	CU_ASSERT_STRING_EQUAL("MULTIPOLYGON(((2 2,5 5,8 2,2 2)),((5 5,2 8,8 8,5 5)))", tmp);
+	lwfree(tmp); lwgeom_free(out); lwgeom_free(in);
 }
 
 /*
--- a/liblwgeom/lwgeom_geos.c
+++ b/liblwgeom/lwgeom_geos.c
@@ -858,19 +858,52 @@ LWGEOM *
 lwgeom_clip_by_rect(const LWGEOM *geom1, double x1, double y1, double x2, double y2)
 {
 	LWGEOM *result;
-	LWGEOM *tmp;
+	GEOSGeometry *geos_orig, *geos_env, *geos_result;
 
 	/* This lwgeom_intersection should be a call to GEOSClipByRect:
 	 * g3 = GEOSClipByRect(g1, x1, y1, x2, y2);
 	 * Unfortunately as of GEOS 3.7 it chokes on practical inputs.
 	 * GEOS ticket: https://trac.osgeo.org/geos/ticket/865
-	 */
+	 * Instead we manually call GEOSIntersection and handle exceptions to guarantee
+	 * the function works with invalid inputs */
+
+	if (!geom1)
+		return NULL;
+	if (lwgeom_is_empty(geom1))
+		return lwgeom_clone_deep(geom1);
+
+	initGEOS(lwnotice, lwgeom_geos_error);
+
+	if (!(geos_orig = LWGEOM2GEOS(geom1, LW_TRUE)))
+		GEOS_FAIL();
 
 	LWGEOM *envelope = (LWGEOM *)lwpoly_construct_envelope(geom1->srid, x1, y1, x2, y2);
-	result = lwgeom_intersection(geom1, envelope);
+	geos_env = LWGEOM2GEOS(envelope, LW_TRUE);
 	lwgeom_free(envelope);
+	if (!geos_env)
+		GEOS_FREE_AND_FAIL(geos_orig);
 
-	if (!result) return NULL;
+	if (!(geos_result = GEOSIntersection(geos_orig, geos_env)))
+	{
+		// Handle invalid input
+		GEOSGeometry* geos_orig_fixed;
+		if ((geos_orig_fixed = LWGEOM_GEOS_makeValid(geos_orig)))
+		{
+			geos_result = GEOSIntersection(geos_orig_fixed, geos_env);
+			GEOSGeom_destroy(geos_orig_fixed);
+		}
+	}
+
+	GEOSGeom_destroy(geos_orig);
+	GEOSGeom_destroy(geos_env);
+
+	if (!geos_result)
+		return NULL;
+
+	if (!(result = GEOS2LWGEOM(geos_result, FLAGS_GET_Z(geom1->flags))))
+		GEOS_FREE_AND_FAIL(geos_result);
+	GEOS_FREE(geos_result);
+	result->srid = geom1->srid;
 
 	/* clipping should not produce lower dimension objects */
 	if (
@@ -879,7 +912,8 @@ lwgeom_clip_by_rect(const LWGEOM *geom1, double x1, double y1, double x2, double
 	    /* output may have different things inside */
 	    result->type == COLLECTIONTYPE)
 	{
-		tmp = lwcollection_as_lwgeom(lwcollection_extract(lwgeom_as_lwcollection(result), lwgeom_dimension(geom1) + 1));
+		LWGEOM *tmp = lwcollection_as_lwgeom(lwcollection_extract(lwgeom_as_lwcollection(result),
+																  lwgeom_dimension(geom1) + 1));
 		lwfree(result);
 		result = tmp;
 		if (!result) return NULL;
--- a/regress/clipbybox2d_expected
+++ b/regress/clipbybox2d_expected
@@ -2,7 +2,8 @@
 2|BOX(5 5,8 8)
 3|BOX(2 2,8 8)
 4|MULTIPOINT(0 0,2 2)
-ERROR:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
+NOTICE:  Self-intersection
+5|MULTIPOLYGON(((2 2,5 5,8 2,2 2)),((5 5,2 8,8 8,5 5)))
 6|MULTIPOLYGON(((2.5 2,5 4,5 5,10 5,10 2,2.5 2)))
 7|POLYGON((2 2,2 5,5 5,5 2,2 2))
 8|SRID=3857;POLYGON EMPTY
--- a/regress/mvt_expected
+++ b/regress/mvt_expected
@@ -44,9 +44,6 @@ PG41 - ON |LINESTRING(0 10,0 4,0 2,0 0,1 0)
 PG41 - OFF|LINESTRING(0 10,0 4,0 2,0 0,1 0)
 PG42 - ON |LINESTRING(0 10,0 0,1 0)
 PG42 - OFF|LINESTRING(0 10,0 0,1 0)
-NOTICE:  lwgeom_intersection: GEOS Error: TopologyException: Input geom 0 is invalid: Self-intersection
-NOTICE:  Self-intersection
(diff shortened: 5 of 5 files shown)


dbaston:
I'm still wondering why we `lwgeom_clip_by_rect` can't just call `GEOSClipByRect` as it did in 2.4, but if there's no performance gain to be had I guess there's not much point in it. @strk I think you're the most familiar with `lwgeom_clip_by_rect`. Do you have a test case where the performance difference is significant?

Algunenano:
Wasn't there a bug with `GEOSClipByRect`? Last time I tried it (last week?) one of the queries in the regress tests got stuck.

dbaston:
@Algunenano I'm not sure. I haven't seen a failing input for `GEOSClipByRect`. The hanging regression test might be from how `ST_Subdivide` handles the output of `GEOSClipByRect`. Maybe @Komzpa knows more.

Algunenano:
It was 10 days ago: [comment](https://github.com/postgis/postgis/pull/268#discussion_r202718588)

I can retest this tomorrow, but I think there hasn't been anything pushed into GEOS since.

dbaston:
@Algunenano what I'm trying to say is that the failure may be caused by a bug in `ST_Subdivide`, not `GEOSClipByRect`.

Algunenano:
Ah, I wasn't following you.

I'm having a look (built with the equivalent of the old function and run the test):
```sql
with inverted_geom as (
    select ST_Difference(
               ST_Expand('SRID=3857;POINT(0 0)' :: geometry, 20000000),
               ST_Buffer(
                   'SRID=3857;POINT(0 0)' :: geometry,
                   1,
                   1000
               )
           ) as geom
)
select '#3744', ST_Area(ST_Simplify(ST_Union(geom), 2))::numeric
from (
         select ST_Subdivide(geom) geom
         from inverted_geom
     ) z;
```

It seems to get stuck in GEOS code when  `GEOSClipByRect_r` is called. I've attached several times to the process and it's not moving from:
```
(gdb) bt
#0  geos::operation::intersection::distance (rect=..., x1=3.0481343214686657e-14, y1=-1, x2=3.0481343214686657e-14, y2=-1)
    at RectangleIntersectionBuilder.cpp:204
#1  0x00007ff35828801e in geos::operation::intersection::distance (rect=..., ring=...) at RectangleIntersectionBuilder.cpp:247
#2  geos::operation::intersection::RectangleIntersectionBuilder::reconnectPolygons (this=0x7ffe21add490, rect=...)
    at RectangleIntersectionBuilder.cpp:409
#3  0x00007ff3582855a4 in geos::operation::intersection::RectangleIntersection::clip_polygon_to_polygons (this=<optimized out>, 
    g=0x55c337797dc0, toParts=..., rect=...) at RectangleIntersection.cpp:535
#4  0x00007ff358285d43 in geos::operation::intersection::RectangleIntersection::clip_polygon (this=<optimized out>, g=0xbfffffff, 
    parts=..., rect=..., keep_polygons=true) at RectangleIntersection.cpp:551
#5  geos::operation::intersection::RectangleIntersection::clip_geom (this=0x7ffe21add5a0, g=0x55c337797dd8, parts=..., rect=..., 
    keep_polygons=true) at RectangleIntersection.cpp:652
#6  0x00007ff358286176 in geos::operation::intersection::RectangleIntersection::clip (this=0x55c337797dd8)
    at RectangleIntersection.cpp:695
#7  geos::operation::intersection::RectangleIntersection::clip (g=..., rect=...) at RectangleIntersection.cpp:686
#8  0x00007ff358bf0790 in GEOSClipByRect_r (extHandle=0x55c33779a090, g=0x55c337797dd8, xmin=<optimized out>, ymin=<optimized out>, 
    xmax=<optimized out>, ymax=<optimized out>) at geos_ts_c.cpp:2569
#9  0x00007ff358ec0582 in lwgeom_clip_by_rect (geom=0x55c33779b2c0, x1=3.0481343214686657e-14, y1=-20000000, x2=20000000, y2=-1)
    at lwgeom_geos.c:964
#10 0x00007ff358e81256 in lwgeom_subdivide_recursive (geom=<optimized out>, maxvertices=256, depth=3, col=<optimized out>)
    at lwgeom.c:2394
```

I'm trying to extract a minimal test case for GEOS, either with the call to `GEOSClipByRect_r` or the call to the first `geos::operation::intersection::distance`.

Algunenano:
I've created a PR in Geos that I can use to reproduce the issue there: https://github.com/libgeos/geos/pull/109
I've only tested it in my main machine, but it loops endlessly every time.

Komzpa:
let's merge this?

Komzpa:
Revert can be done conditionally after a fix is in released version.

strk:
I'm ok with skipping GEOS code when bogus, so conditional seems the
way to go to me.

--strk;

Algunenano:
> I'm ok with skipping GEOS code when bogus, so conditional seems the
way to go to me.

The problem I see with that is that we'd need to adapt the tests to work with both branches which is a pain with anything on top of lwgeom_clip_by_rect. You can see the current differences [here](https://github.com/Algunenano/postgis/commit/082e1e53a7514f8183f88fa03bd439687c66f781).

How big is the performance difference between GEOSClipByRect and GEOSIntersection anyway? Is it measured anywhere?

Komzpa:
@Algunenano we can solve tests as an improvement after release of fixed GEOS version. For now, ClipByBox-based Subdivide in your commit loses 0.2% of area, while Intersection-based does not - so there are definitely more issues there.

Algunenano:
I decided to go ahead and test both versions:
image

Blue is this branch (using `GEOSIntersection`) and orange is [this other one](https://github.com/Algunenano/postgis/commit/082e1e53a7514f8183f88fa03bd439687c66f781) (using `GEOSClipByRect`).

Datasets are:
- Line 2: 1 million points from NYC taxi dataset. All valid geometries.
- Line 3: 1.1 million polygons from NYC buildings. All valid geometries.

Running the following query:
```sql
SELECT ST_ClipByBox2d(:GEOM_COLUMN, '0103000020110F000001000000050000006EF8B664BB7A5FC116CDBFFB8E555F416EF8B664BB7A5FC11E42CFC368555F41766DC62C957A5FC11E42CFC368555F41766DC62C957A5FC116CDBFFB8E555F416EF8B664BB7A5FC116CDBFFB8E555F41'::geometry) FROM :TABLE AS cdbq;
```
The geometry corresponds to tile 18/77092/777341 in webmercator projection.

The results are the number of transactions using pgbench in 30 seconds using 6 clients (1 thread per client).

Given all this, I don't see any noticeable performance differences between the 2 alternatives, so I'm all about dropping `GEOSClipByRect` since it's less precise and doesn't offer any improvement.

Komzpa:
@Algunenano for somewhat more computational test, may I ask you to benchmark a polygon version that will hit less short-circuitry?

`Select ST_ClipByBox2D(:geom_column, ST_Expand(:geom_column, -10)) from :table as cdbq;`

strk:
How complex (number of points) are the 1.1 million polygons ?

Algunenano:
> @Algunenano for somewhat more computational test, may I ask you to benchmark a polygon version that will hit less short-circuitry?

I'll give it a try. I'll probably will need to limit the size since 100k points seem to take ~30 seconds.

> How complex (number of points) are the 1.1 million polygons ?

9 points on average:
```sql
Select max(np), min(np), avg(np), count(np) from (Select ST_NPoints("the_geom_webmercator") as np from geo_export_2c10327b_1c8d_4698_8f6f_e6915dc51449) _a;
 max  | min |        avg         |  count  
------+-----+--------------------+---------
 2580 |   4 | 9.0136167367849505 | 1082932
(1 row)
```

Algunenano:
Without the short circuit `GEOSClipByRect` is 35x faster:
image

I withdraw what I said, having `GEOSClipByRect` is worth it.

dbaston:
> Revert can be done conditionally after a fix is in released version.

If the bug is so severe that PostGIS needs to handle it specially, then the PostGIS bugfix should have been backported to 2.3 and 2.4. Given that we have no reports of this in the wild, and the changes were not backported to 2.3 and 2.4, I think it can wait for a GEOS release of which we are in complete control.

> For now, ClipByBox-based Subdivide in your commit loses 0.2% of area, while Intersection-based does not - so there are definitely more issues there.

Those differences should probably be reported as a bug. But more broadly, it sounds like that the contract of `lwgeom_clip_by_rect` is inappropriate for ST_Subdivide, so it is ST_Subdivide that should be updated, not `lwgeom_clip_by_rect`.

Komzpa:
> Given that we have no reports of this in the wild, and the changes were not backported to 2.3 or 2.4,

Report of this in the wild is now unit test that you want to revert. If I'm not wild enough for you remind me to bite you in Boston :)

Backpatching of the change should be done too, thanks for the reminder.

pramsey:
So, is a fair 2.5 course of action as follows:

- for functions that need to clip things (MVT, Subdivide) the function chooses whether they want to use clipbyrect or intersection and internally use that function
- clipbyrect SQL binding is returned to using actual rectangle clipping
- clipbyrect doco gets note that it sometimes returns invalidity
- hanging test case is included in test suite for 2.5 but commented out and ticketed so that when a GEOS with a fixed implementation comes out it can be turned on with a version guard

Yes?

Algunenano:
> for functions that need to clip things (MVT, Subdivide) the function chooses whether they want to use clipbyrect or intersection and internally use that function

I don't think the 2 functions are interchangeable. One autofixes any error that may happen (validity), the other one throws an error that stops the query ( `lwerror`).

> clipbyrect SQL binding is returned to using actual rectangle clipping

Are we willing to live the possibility of an infinite loop? Keep in mind that the query won't be stopped by a timeout and, at least in my tests, it doesn't stop either with `pg_cancel_backend` or `pg_terminate_backend`. I've had to use `kill -9 <pid>`.

It's not like you have to chose between _precision with autofix_ (`GEOSClipByRect`) or _speed_ (`GEOSIntersection`); you have to choose between _precision with autofix_ or _speed with the possibility of having to ssh into the machine to kill a hung process_.

From my point of view, the issue has appeared with `ST_Subdivide` but it is as likely to happen with `ST_AsMVTGeom` or `ST_ClipByBox2d`. People are likely to live with 2.5 for a year, so I'd rather not ship it with the ability to waste CPU cycles until killed.


Here is my proposal instead.
- Subdivide and MVT_AsGeom keep using `lwgeom_clip_by_rect`. Other functions keep using lwgeom_intersection`.
- lwgeom_clip_by_rect is changed as:
```c
#ifdef GEOS_VERSION_WITH_BUG_FIXED
    // Old implementation using `GEOSClipByRect`
#else
    // Implementation using `GEOSIntersection` (current PR)
#endif
```
- Adapt tests to accept both outputs as valid and report any unexpected difference upstream.
- Whenever geos makes a new release (the fix is already merged) we change `GEOS_VERSION_WITH_BUG_FIXED` to `GEOS_NUMERIC_VERSION >= 30702` (or whatever it ends up being).
- Hanging test case is always enabled since it should be fixed in any case. 


Another proposal is to get GEOS 3.7 out with the fix and drop support for any previous release in Postgis. :smile:

dbaston:
Can we take `ST_Subdivide` out of the discussion? Can we agree that it has no business using a routine that doesn't claim to be accurate?

Note also that the GEOS bug fix will be available in 3.5.2, 3.6.3, and 3.7.0.

Komzpa:
@dbaston I see the issue in insisting "ClipByRect doesn't claim to be accurate".

Thing it claimed was "Validity is not checked and not enforced on output". It's similar to Buffer: it doesn't validate input, combines back usually something valid, but interpretation of invalid may be not the one you needed.

Original ticket for implementation had it as just a special case of Intersection: https://trac.osgeo.org/geos/ticket/699

It can be expected that on valid input there is no difference in using either Intersection, or special case optimized ClipByRect. For now it does not hold, so optimization should not be used.

Is there any problem with solution by @Algunenano in
 https://github.com/postgis/postgis/pull/279#issuecomment-409157737 ?

dbaston:
@Komzpa `ST_ClipByBox2D` was introduced in PostGIS 2.2. The docs are [here](https://postgis.net/docs/manual-2.2/ST_ClipByBox2D.html), and you can read what claims are made about the function. If you read the Trac ticket that you linked, you'll see that it was originally thought of as a special case for intersection, but that possibility was discarded during development. In particular, you shouldn't expect that it produces the same output as `ST_Intersection`.

If @Algunenano wants to change the MVT code to not use `lwgeom_clip_by_rect`, or to conditionally not use it, I don't see a problem with that (though it does sound like a testing headache.) I do see an issue with completely changing the semantics of `lwgeom_clip_by_rect`, which has been present in PostGIS for three releases, without a compelling argument.

Algunenano:
> Is it too terrible to have the next release also delegate the reliability of GEOS to... GEOS?

That's a great point; I hadn't thought about it that way. And it's not like the bug it's not going to get fixed upstream; it already is and. according to @dbaston, will be released as part of 3.5.2, 3.6.3, and 3.7.0. I agree that we should push users and maintainers to update their libraries instead of hacking around bugs as much as possible. I agree with your proposal.

That also makes me want to remove any pending `GEOS_NUMERIC_VERSION` or `POSTGIS_GEOS_VERSION` referencing 3.4 or before.

robe2:
Just a note, we should remove all references of 3.4 in PostGIS 2.5 code.  GEOS 3.5 is our minimum for PostGIS 2.5.  Also I agree with lets force people to upgrade to the latest micro by not trying to check patch versions.  I hate checks for micro versions as it just clutters our code, means people can't do runtime upgrades easily and we forget to remove them.  We should put a big disclaimer in the release notes -- **Make sure you are running with GEOS 3.5.2+, 3.6.3+, or 3.7.0+.**

When are we going to fix the MVT regress failure?  Debbie and Winnie are building with latest 3.7 GEOS head and still have this error.

Algunenano:
> When are we going to fix the MVT regress failure? Debbie and Winnie are building with latest 3.7 GEOS head and still have this error.

I'll be fixed by https://github.com/postgis/postgis/pull/282


I'm closing this since it's superseeeded by https://github.com/postgis/postgis/pull/281 and https://github.com/postgis/postgis/pull/282

PR #280, opened 2018-07-30
Use user provided CFLAGS in topology and address standardizer

Algunenano:
For both extensions, use the CFLAGS provided in the `configure` step.

For topology I've decided to not change the function type signatures to get this into 2.5, so I've left several `TODO`s as the implementation is buggy if more rows are turned that `INT_MAX`. I've tried to make it so if this happens only information would be lost instead of having undefined behaviour.

### DIFF
--- a/liblwgeom/liblwgeom_topo.h
+++ b/liblwgeom/liblwgeom_topo.h
@@ -188,6 +188,7 @@ typedef struct LWT_BE_CALLBACKS_T {
    * @param ids an array of element identifiers
    * @param numelems input/output parameter, pass number of node identifiers
    *                 in the input array, gets number of node in output array.
+   *	TODO: Should be uint64 to match SPI_processed
    * @param fields fields to be filled in the returned structure, see
    *               LWT_COL_NODE_* macros
    *
@@ -212,6 +213,7 @@ typedef struct LWT_BE_CALLBACKS_T {
    * @param numelems output parameter, gets number of elements found
    *                 if the return is not null, otherwise see @return
    *                 section for semantic.
+   *	TODO: Should be uint64 to match SPI_processed
    * @param fields fields to be filled in the returned structure, see
    *               LWT_COL_NODE_* macros
    * @param limit max number of nodes to return, 0 for no limit, -1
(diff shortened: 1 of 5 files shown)


dbaston:
I don't see any issue changing the signatures for 2.5 (these aren't user-facing functions); others may disagree.

liblwgeom/liblwgeom_topo.h
```diff
@@ -188,6 +188,7 @@ typedef struct LWT_BE_CALLBACKS_T {
    * @param ids an array of element identifiers
    * @param numelems input/output parameter, pass number of node identifiers
    *                 in the input array, gets number of node in output array.
+   *	TODO: Should be uint64 to match SPI_processed
```
strk: We don't need a match, btw, as this module is PostgreSQL agnostic. Possibly we do need a bigger integer, to _allow_ for returning more than 4 million faces

Algunenano:
> I don't see any issue changing the signatures for 2.5 (these aren't user-facing functions); others may disagree.

I don't mind going either way.

> I see many changes unrelated to the PR subject

The changes are to fix the warnings/errors that arise once the flags are correctly passed, since otherwise the builds [will break](https://github.com/postgis/postgis/blob/1f63d85c5f2ed8bc07949c062664e8f6b242bb8c/.travis.yml#L5).

liblwgeom/liblwgeom_topo.h
Algunenano: True, it's design as platform agnostic, but I've only seen this used in the `topology/` folder which is tightly tied to Postgres.

strk:
You'll find [librttopo](https://git.osgeo.org/gitea/rttopo/librttopo) being a 1:1 mapping of the liblwgeom version of topology support, except that one is using the re-entrant GEOS API (we should do as well).

Algunenano:
> You'll find librttopo being a 1:1 mapping of the liblwgeom version of topology support, except that one is using the re-entrant GEOS API (we should do as well).

Nice to know.

Mainly to avoid introducing possible issues in a beta, I'm going to merge this as is and defer changing the types to 3.0. I'll open a ticket to serve as a reminder.

PR #281, opened 2018-07-31
Make lwgeom_subdivide use lwgeom_intersection for clipping

dbaston:
Thjs PR updates `lwgeom_subdivide` to use `lwgeom_intersection` for
clipping instead of `lwgeom_clip_by_rect`. No tests are changed.

`lwgeom_clip_by_rect` describes itself as clipping in a "fast but possibly
dirty way." This is not a good fit for `lwgeom_subdivide`, a primary use of
which is the optimization of point-in-polygon processes. That application
requires that the area covered by the subdivided geometry be the same as
the area covered by the original geometry, which in turn requires that a
robust intersection routine be used.

As discussed in #279 . Merging this allows `lwgeom_clip_by_rect` to be
reverted to its 2.4 state without affecting `lwgeom_subdivide`.

### DIFF
--- a/liblwgeom/lwgeom.c
+++ b/liblwgeom/lwgeom.c
@@ -2249,10 +2249,10 @@ lwgeom_grid(const LWGEOM *lwgeom, const gridspec *grid)
 
 
 /* Prototype for recursion */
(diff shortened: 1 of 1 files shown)


dbaston:
Travis failure is from MVT tests; subdivide tests are OK.

dbaston:
Any objections to this?

PR #282, opened 2018-08-01
Revert lwgeom_clip_by_rect changes

dbaston:
This PR reverts `lwgeom_clip_by_rect` to its implementation in PostGIS 2.2 - 2.4.

It makes trivial modifications to the MVT tests to reflect the different results returned by `GEOSClipByRect` instead of `GEOSIntersection`. The changed results are topologically equivalent to the previous results, and their orientation is unchanged.

No tests have been commented out or made version-dependent.


dbaston:
Note that Travis failure is from a compiler warning in third-party code.

PR #283, opened 2018-08-02
Set up Travis build matrix using prebuilt Docker images

dbaston:
This PR sets up a Travis build matrix for PostGIS using prebuilt Docker images with various combinations of PostgreSQL, GEOS, PROJ, and GDAL.

The current configuration tests PostgreSQL 9.4-12dev, GEOS 3.5-3.7dev, GDAL 1.11/2.2/2.3/2.4dev, and PROJ 4.9/5.1/5.2dev. The whole cartesian product of these is not tested; I tried to come up with some sensible combinations.

Some advantages relative to the current setup:

- We can test against a range of dependencies, including unreleased versions
- Build times are substantially reduced (> 20 minutes to ~5 minutes, though some of the difference may be from extra tools in the current configuration that we would want to replicate in this configuration)

And a primary disadvantage:

- We have the maintenance burden of keeping the Docker images up-to-date. Currently they are generated by a script in [this repository](https://github.com/dbaston/postgis-build-env).

What do others think about this approach?


Algunenano:
I think this is a nice approach to broaden the tests. The only thing I'm not so sure it's using a non-organization repo but I guess we can give a try and migrate in the future if needed.

Some questions from someone that has never set up Docker:
- Can we customize CFLAGS? I guess it's possible, but I don't see them in the `build.py`, and checking for warnings (`-Werror -Wall -Wextra -Wno-unused-parameter`) was something that was only done in this project.
- How often is the `latest` image updated? Is it rebuilt with every PR, weekly, manually? In the past I've had some issues with PRs building against images with old packages, which is not that big of an issue for `old` but it is for something like `latest`. Let's say GEOS 3.5.2 and 3.6.3 get released, when would we see them in the builds using `GEOS='svn-3.6',`?
- How difficult it would be to add a new build with new compilers that use a different script (not `ci/travis/run_tests.sh`) and flags? At some point I would like to test liblwgeom under sanitizers (https://trac.osgeo.org/postgis/ticket/3881), and I was wondering if this will make it easier or more difficult.

dbaston:
Hi @Algunenano ,

I agree that we would want to move the code under the `postgis` repo if we go this route. For your questions:

> Can we customize CFLAGS?

Yes. Probably the easiest place to specify them is in `run_tests.sh` (which should probably be named `build_and_test.sh` or something).

> How often is the latest image updated?

Whenever we choose to build it. That's the maintenance burden. I think a weekly cron job seems reasonable, unless we have a specific reason to trigger a build manually (for example, if a fix was just pushed to Postgres master).

> Let's say GEOS 3.5.2 and 3.6.3 get released, when would we see them in the builds using GEOS='svn-3.6',?

The images are all built from git branches rather than specific releases. So `geos35` just reflects whatever is in the GEOS 3.5 branch at the time we build image.

> How difficult it would be to add a new build with new compilers that use a different script (not ci/travis/run_tests.sh) and flags?

I don't think this PR makes it any easier or difficult. We could either modify `run_tests.sh` to add some extra stuff (I'd like to have it valgrind the cunit tests, for example), or we could have Travis run more than one script.

Komzpa:
Things that need to be done:

 - automate images build. can this be done by Travis itself, so there is no reliance on external tools? check for latest GEOS/Proj/GDAL commit on branch, try pull that image from docker repository, if failed - build it and push?

 - put back codecov support.

Building our own GEOS can be really beneficial: for --coverage build, we can produce a report of lines in GEOS code that are really utilized by PostGIS tests. That can assist debugging issues spanning from Postgres input to GEOS internal exceptions, without need to isolate intermediate test cases.

dbaston:
> * automate images build. can this be done by Travis itself, so there is no reliance on external tools? check for latest GEOS/Proj/GDAL commit on branch, try pull that image from docker repository, if failed - build it and push?

I think the simplest thing is to rebuild the images at some frequency (weekly?) with the ability to manually rebuild if desired. I'm happy to set this up as a cron job on my laptop; I realize that creates a dependency on non-project infrastructure, but I have no idea how we would get project infrastructure for this. And the code to build all of the images can be put in a project repo, so it doesn't seem like it's a very storng dependency. Alternatively, the builds could be done using GitLab CI and triggered manually; I'm not sure about doing it on a timer.

> * put back codecov support.

Sure, can you let me know what to add?

Komzpa:
It seems it's rather trivial to push from travis directly:
https://ops.tips/blog/travis-ci-push-docker-image/ (a new account for travis itself would be needed)

This way travis can:
 - got and curl somewhere the latest revision of GEOS, GDAL, Proj, Postgres, whatever;
 - attempt `docker pull postgis-travis:gdal-GDALREV-geos-GEOSREV-proj-PROJREV...`;
 - if attempt fails, build it;
 - run postgis tests;
 - push layers to registry.

In this scheme, it survives even a dead registry, and pulls up everything as fixes get pushed.

For coverage: you need to build with `-O0 --coverage`, make sure postgres instance can write to the directory with sources (currently I remap user id's), and after test run the codecov's uploader bash script.

dbaston:
@Komzpa are you suggesting that the images be built as part of the PostGIS build? It seems like that would cause a PostGIS build failure if any of the dependent project builds fails, and would introduce huge uncertainty into the PostGIS CI runtimes. I don't think the build environments need to be up to the latest commit for any of the dependencies, unless we have a specific reason to think that we do, in which case the build environments can be manually updated.

Are you interested in supplying a patch for the codecov stuff?

Komzpa:
@dbaston If any of dependent project builds fail on their release branch, it's a good reason to check what's happening there. Otherwise we can have sudden issues like sfcgal being lost from ubuntu as it can't be compiled with new cgal, as we faced in Ubuntu Bionic which unfortunately got stuck in this state 
until it's EOL in April 2023. Nobody thought about updating anything in that scenario. That's part of the reason why Continuous Integration is called this way :)

I also believe that Dockerfile and build script should be a part of postgis repo. Can you commit the files into .ci/travis? Just to the trunk, without connecting them anywhere for now. I can then have a take on it on a free day.

dbaston:
Checking that SFCGAL builds against the newest CGAL sounds like the role of the SFCGAL build process, not the PostGIS build process. This PR attempts to build against the latest buildable versions of dependencies, while making sure that the PostGIS build process never fails before starting to actually build PostGIS. (Yes, this is CI.)

You can download the Dockerfiles from the linked repo. It may make sense to copy them into the PostGIS repo, or they may need to be kept separate (if you want GitLab CI or Travis to do the image build separately from the PostGIS build, I think they'd need to be separate).

ci/travis/run_tests.sh
```diff
@@ -0,0 +1,22 @@
...
+LDFLAGS_COV="--coverage"
+
+/usr/local/pgsql/bin/pg_ctl -l /tmp/logfile start
+./autogen.sh
+
+CFLAGS="${CFLAGS_STD}" LDFLAGS="${LDFLAGS_COV}" ./configure
+make -j2
+make check
+
+CFLAGS="${CFLAGS_COV}" LDFLAGS="${LDFLAGS_COV}" ./configure
```
Komzpa: `--enable-debug` on this configure call should get coverage inside some ifdefs.

ci/travis/run_tests.sh
```diff
@@ -0,0 +1,22 @@
...
+make -j2
```
Komzpa: these two can be just `make -j check` - it will build everything it needs to and interleave that with check.

dbaston:
Merged at  faac4cf2ea

PR #286, opened 2018-08-10
Speedup MVT column parser

Algunenano:
Speedup MVT columns parsing by:
* Avoid multiple `strlen` when parsing JSON columns.
* Use `heap_deform_tuple` instead of `GetAttributeByNum` for normal columns.
* Cache the table index and oid for columns to avoid recalculation. Reuse the arrays for Datum and null values instead of allocating and deallocating them for each row. Same for the `TupleDesc`.

I'm attaching some bechmarks for a dataset where the previous process was a big issue. It has one geometry column (already prepared for MVT), a couple of strings and the rest of the columns (**~640**) are doubles:

```bash
$ cat mc.pgb 
-- Run with something like:
-- pgbench -c 4 -T 30 -f mc.pgb -U postgres mc_perf

SELECT St_AsMVT(q) FROM (
SELECT * FROM tilertest.tract9double
) q;
```

* Before:
```
$ pgbench -c 4 -T 30 -f mc.pgb -U postgres mc_perf
starting vacuum...ERROR:  relation "pgbench_branches" does not exist
(ignoring this error and continuing anyway)
ERROR:  relation "pgbench_tellers" does not exist
(ignoring this error and continuing anyway)
ERROR:  relation "pgbench_history" does not exist
(ignoring this error and continuing anyway)
end.
transaction type: mc.pgb
scaling factor: 1
query mode: simple
number of clients: 4
number of threads: 1
duration: 30 s
number of transactions actually processed: 68
latency average = 1870.297 ms
tps = 2.138698 (including connections establishing)
tps = 2.138903 (excluding connections establishing)
```

* After:
```
$ pgbench -c 4 -T 30 -f mc.pgb -U postgres mc_perf
starting vacuum...ERROR:  relation "pgbench_branches" does not exist
(ignoring this error and continuing anyway)
ERROR:  relation "pgbench_tellers" does not exist
(ignoring this error and continuing anyway)
ERROR:  relation "pgbench_history" does not exist
(ignoring this error and continuing anyway)
end.
transaction type: mc.pgb
scaling factor: 1
query mode: simple
number of clients: 4
number of threads: 1
duration: 30 s
number of transactions actually processed: 806
latency average = 149.464 ms
tps = 26.762339 (including connections establishing)
tps = 26.765364 (excluding connections establishing)
```


Algunenano:
Travis is [having some issues](https://www.traviscistatus.com/incidents/pxj3py3xzn14) with its network and some builds are failing, but since 3/4 went ok I'll proceed anyway.

PR #293, opened 2018-09-11
GiST and SP-GiST n-D indexes (based on GIDX) with overlaps, contains, within, and equals operators

estebanzimanyi:
The current version of PostGIS provides GiST n-D indexes (based on GIDX) only with the &&& (overlaps) operator. This PR extends the GiST n-D indexes for three additional operators:  ~~ (contains) @@ (within) ~~= (equals). It also implements SP-GiST n-D indexes that support these four operators.


alesuiss:
N.b. that there is still some instability in the results for GiST n-D indexes (cf. ticket [#4139](https://trac.osgeo.org/postgis/ticket/4139))

Komzpa:
@alesuiss may I ask you to split this into two parts then, so we can merge it easier?

a green one with sp-gist comparing itself to non-indexed scan, and a failing to-fix one exposing the bug in gist? this can probably be achived by temporarily commenting out failing gist test.

alesuiss:
I guess the coverage issue is because we started from an earlier version of svn-trunk. Do we need to do something about that?

PR #300, opened 2018-09-14
Topology: Use the qualified names of the extension tables

Algunenano:
Fixes https://trac.osgeo.org/postgis/ticket/2503

Not sure how to test this (except manually) and how long do we need to backport it (at least 2.4). I'll have a look.


Komzpa:
Congratulations with a get! (#300)

PR #302, opened 2018-09-16
WIP: Rename C functions to match SQL names

Komzpa:


Komzpa:
@pramsey what are the downsides?

For symlink upgrade we can create a new `legacy.c` file that will create all the old functions and raise `lwerror("Please finish PostGIS upgrade process");`.

pramsey:
I think that big changes should more appropriately answer: what are the upsides? Churn for churns sake? We've made it this long without changing them, and the upside of staying the same has been maintaining easy compatibility across upgrades, which has frequently saved people's skins when things have gone pear-shaped.

robe2:
With @Komzpa  other suggestion of having a legacy.c, we'll remember what they used to be :)

I think it will keep our sanity if our exposed function names more closely match the SQL function names they support.

There is no upgrade downside I see, and if we add in the 2.1 functions we took out, we can get everyone on board for 3.0 and less reason for people to be using 2 series when we are done.

PR #303, opened 2018-09-17
ST_AsMVT: Proper handling of id column

Algunenano:
Trac: https://trac.osgeo.org/postgis/ticket/4128

Missing documentation update


Komzpa:
@Algunenano thanks for getting this polished! :)

Algunenano:
Np, I've been looking into the MVT code lately so I knew where to look.

joedjc:
@Algunenano This is great! It will be really useful in combination with the new feature-state mapbox feature. Do you have an idea when this will be available to use? I notice it's listed as part of the 3.0 release, but according to the roadmap that could be a year away? Would be great if it can make the next release if it's ready

Komzpa:
@joedjc it will be around next september, PostGIS 2.5.0 just went out. You can use development builds for now.

joedjc:
@Komzpa Thank you for the quick reply. Ok I will try that as it would be a shame not to be able to use this now. I've not used a dev build before so if you have any reference on how to use (and if it's possible to use with a AWS managed RDS) i'd be very grateful.

PR #304, opened 2018-09-19
St_AsMVTGeom: Avoid polygon type transformation

Algunenano:
Polygons could be transformed into lines in the lwgeom_make_valid call.
This doesn't play nicely with MVT renderers so we drop them instead.

Trac issue: https://trac.osgeo.org/postgis/ticket/4181


Komzpa:
Can we merge this so it gets into 2.5.0 which @robe2 plans to get out before sprint? :)

Algunenano:
I was planning on merging this on Monday but I'll see if I can merge this today.

PR #305, opened 2018-09-21
Update ST_ChaikinSmoothing to also smooth start/end point of polygon …

Komzpa:
…by default

https://trac.osgeo.org/postgis/ticket/4156


Algunenano:
Could you add the change to the default in the docs and `postgis/postgis.sql.in`?

Komzpa:
@Algunenano I expect we can get this into 2.5, so we never have a release that requires passing (,,false) for all the polygons and it's not a change then.

Algunenano:
Seems fine to be me then :+1:

PR #306, opened 2018-09-21
Mvt clipping

Algunenano:
WIP for https://trac.osgeo.org/postgis/ticket/4183

Changes are on top of https://github.com/postgis/postgis/pull/304 which I intend to merge separately..


Algunenano:
I'm currently benchmarking to see what's the less impactful option. One thing I hadn't considered is that to compare bboxes after clipping you need to generate the bbox of the clipped geometry, which I think we didn't before.

Algunenano:
For this specific case it's 1.4x faster to return NULL than to fix the geometry and retry so, for now, I'll leave it to return an empty geometry and wait until there is configuration about autofixing to add other options.

PR #308, opened 2018-09-27
ST_Subdivide with OpenMP

Komzpa:


Komzpa:
good openmp manual: https://bisqwit.iki.fi/story/howto/openmp/

PR #313, opened 2018-10-02
St_AsMVTGeom: Handle type changes in geometry collections

Algunenano:
Fixes for https://trac.osgeo.org/postgis/ticket/4181

This behaviour was being hit in the test for 3922 but it was hidden behind the `length(bytea` call since the test only intended to test a crash.


Algunenano:
Merged in https://github.com/postgis/postgis/commit/4385284de6156a83a51cd640b5362ccd22d90492

PR #325, opened 2018-11-08
Added a row in Synopsis

SunveerSingh:
According to https://trac.osgeo.org/postgis/ticket/2767 a raw in Synopsis need to be added about AddRasterConstraints tablename and fieldname

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -29,7 +29,7 @@ PostGIS 3.0.0
   - #4215, Use floating point compare in ST_DumpAsPolygons (Darafei Praliaskouski)
   - #4155, Support for GEOMETRYCOLLECTION, POLYGON, TIN, TRIANGLE in
     ST_LocateBetween and ST_LocateBetweenElevations (Darafei Praliaskouski)
-  - #2767, Added "opt" and Boolean values for AddRasterConstraint (Sunveer Singh)  
+  - #2767, Documentation for AddRasterConstraint optional parameters (Sunveer Singh)
 
 PostGIS 2.5.0
 2018/09/23
--- a/NEWS
+++ b/NEWS
@@ -29,6 +29,7 @@ PostGIS 3.0.0
   - #4215, Use floating point compare in ST_DumpAsPolygons (Darafei Praliaskouski)
   - #4155, Support for GEOMETRYCOLLECTION, POLYGON, TIN, TRIANGLE in
     ST_LocateBetween and ST_LocateBetweenElevations (Darafei Praliaskouski)
+  - #2767, Added "opt" and Boolean values for AddRasterConstraint (Sunveer Singh)  
 
 PostGIS 2.5.0
 2018/09/23
--- a/doc/reference_raster.xml
+++ b/doc/reference_raster.xml
@@ -452,25 +452,25 @@ VALUES (1,
 			<paramdef><type>name </type>
 			<parameter>rastcolumn</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>srid</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>scale_x</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>scale_y</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>blocksize_x</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>blocksize_y</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>same_alignment</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
+			<paramdef choice="opt"><type>boolean </type>
 			<parameter>regular_blocking</parameter></paramdef>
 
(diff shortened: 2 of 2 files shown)


Komzpa:
@SunveerSingh since this change is going to get applied to already released branches, could you also add a line about it to NEWS file in the root directory of repo, stating ticket number, change description and your name?

SunveerSingh:
@Komzpa kan you run the travis again, I am having some network problem, that is why it failed.

Komzpa:
Tests failed:
```
psql:/src/postgis/regress/00-regress-install/share/contrib/postgis/raster_comments.sql:4: ERROR:  function addrasterconstraints(name, name) does not exist
```

So, there is no such simple function. Sorry for misguidance. (It's good to have tests that check documentation :) )

Looking at https://github.com/postgis/postgis/blob/svn-trunk/raster/rt_pg/rtpostgis.sql.in#L8303 it's in fact a longer call that sets most of arguments by default. Can you go over all the AddRasterConstraints definitions in file https://github.com/postgis/postgis/blob/svn-trunk/raster/rt_pg/rtpostgis.sql.in and see whether they have a matching entry in documentation?

SunveerSingh:
Ok we dont have tablename name; feildname; here https://github.com/postgis/postgis/blob/svn-trunk/raster/rt_pg/rtpostgis.sql.in

Komzpa:
@SunveerSingh the one on line 8303 is the one that can be called as (tablename,fieldname), just that it has more optional parameters.

SunveerSingh:
And I guess tablename and fieldname is not required so that is why the user "can" use, and as you said they are optional so we can instead of writing it here we can add a note "The user can call AddRasterConstraints tablename, fieldname.(Optional)"

Komzpa:
@SunveerSingh "Note:" is never a good choice. :)

Can you look at XML near the ones you constructed? there is something with "opt" (that means parameter is optional), and some of the names have =true on them (meaning that when it's not set, it supposed to be True). So you need to add the tail of optional arguments to the block in your PR that will match the order and default values of the one on line 8303.

SunveerSingh:
ok so first should I add tablename name; and fieldname name; here: https://github.com/postgis/postgis/blob/svn-trunk/raster/rt_pg/rtpostgis.sql.in#L8303?

Komzpa:
@SunveerSingh they are already there. Names in code and in docs don't match, 'rasttable' is the same thing as 'tablename', and 'fieldname' is the same thing as 'rastcolumn'. If you dislike that they don't match you can go greater length and change that too, but that can be done after we're done with this change as separate PR :)

SunveerSingh:
if they are already then what should I do now?

Komzpa:
The way I see it:

1. remove entry you added;
2. on entry on line 447 in xml file, mark things that are optional and their default value according to order of things in L8303;
3. add NEWS file change you did in #326 into this PR, #325.

SunveerSingh:
1. remove entry you added-done!

SunveerSingh:
2.on entry on line 447 in xml file, mark things that are optional and their default value according to order of things in L8303;-Done!

SunveerSingh:
Done!

SunveerSingh:
3.add NEWS file change you did in #326 into this PR, #325.-Done!

NEWS
```diff
@@ -29,6 +29,7 @@ PostGIS 3.0.0
   - #4215, Use floating point compare in ST_DumpAsPolygons (Darafei Praliaskouski)
   - #4155, Support for GEOMETRYCOLLECTION, POLYGON, TIN, TRIANGLE in
     ST_LocateBetween and ST_LocateBetweenElevations (Darafei Praliaskouski)
+  - #2767, Added "opt" and Boolean values for AddRasterConstraint (Sunveer Singh)
```
Komzpa:
```suggestion
  - #2767, Documentation for AddRasterConstraint optional parameters (Sunveer Singh)
```

Komzpa: note that you've added spaces at the end here, and tests may complain :)

doc/reference_raster.xml
```diff
@@ -489,7 +489,7 @@ VALUES (1,
 				<parameter>extent=true</parameter>
 			</paramdef>
 		  </funcprototype>
-
+
```
Komzpa: also, spaces on the end. be careful when you press Enter on github :)
```suggestion

```

doc/reference_raster.xml
```diff
@@ -452,26 +452,26 @@ VALUES (1,
...
 
-			<paramdef><type>boolean </type>
-			<parameter>same_alignment</parameter></paramdef>
+			<paramdef choice="opt"><type>boolean </type>
+			<parameter>same_alignment=true</parameter></paramdef>
 
-			<paramdef><type>boolean </type>
-			<parameter>regular_blocking</parameter></paramdef>
+			<paramdef choice="opt"><type>boolean </type>
+			<parameter>regular_blocking=false</parameter></paramdef>
```
Komzpa: :+1: you found that tricky one :)

NEWS
SunveerSingh: Thanks for suggestions

doc/reference_raster.xml
SunveerSingh: Thanks to you!

SunveerSingh:
What is left now?

Komzpa:
Let's get a green tick from travis to see that we've not mixed it all up again, and we're probably good to go :)

SunveerSingh:
Thanks a lot! can we close this: https://trac.osgeo.org/postgis/ticket/2767#comment:6

Komzpa:
@SunveerSingh patience! a green tick from Travis, a merge of PR into release branch, and it will get closed. :)

SunveerSingh:
No Worries, Thank a lot for helping me out!

Komzpa:
@SunveerSingh congratulations, merged :)

SunveerSingh:
@Komzpa Thanks a lot! :)

PR #328, opened 2018-11-11
Specified 1-indexing in ST_SetValues

SunveerSingh:
According to ticket https://trac.osgeo.org/postgis/ticket/2835 ST_SetValues needed to be specified about raw and column are 1-indexed.

### DIFF
--- a/doc/reference_raster.xml
+++ b/doc/reference_raster.xml
@@ -5314,7 +5314,7 @@ GROUP BY (foo.geomval).val;
 			<refsection>
 				<title>Description</title>
 				<para>
-					Returns modified raster resulting from setting specified pixels to new value(s) for the designated band.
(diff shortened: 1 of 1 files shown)


doc/reference_raster.xml
```diff
@@ -5314,7 +5314,7 @@ GROUP BY (foo.geomval).val;
 			<refsection>
 				<title>Description</title>
 				<para>
-					Returns modified raster resulting from setting specified pixels to new value(s) for the designated band.
+					Returns modified raster resulting from setting specified pixels to new value(s) for the designated band. Row and Column are 1-indexed.
```
Komzpa:
```suggestion
					Returns modified raster resulting from setting specified pixels to new value(s) for the designated band. <varname>columnx</varname> and <varname>rowy</varname> are 1-indexed.
```

This can also go to separate paragraph just below this one.

SunveerSingh:
@Komzpa why it is failing?

PR #330, opened 2018-11-12
Enable debug code in Travis build

dbaston:
Make sure we catch compile or test failures with `--enable-debug`.


dbaston:
I think this accomplishes what is needed, but I'll leave it until after 75ad14a lands in trunk.

Algunenano:
I've merged my changes into trunk. Can you close https://trac.osgeo.org/postgis/ticket/4235 when you merge this one please?

PR #334, opened 2018-11-14
Added info about maxerr in ST_Rescale

SunveerSingh:
According to ticket #2921 some information about maxerr needed to be added. https://trac.osgeo.org/postgis/ticket/2921

### DIFF
--- a/doc/reference_raster.xml
+++ b/doc/reference_raster.xml
@@ -6568,7 +6568,7 @@ FROM (
 
 				<para>When the new scalex or scaley is not a divisor of the raster width or height, the extent of the resulting raster is expanded to encompass the extent of the provided raster. If you want to be sure to retain exact input extent see <xref linkend="RT_ST_Resize" /></para>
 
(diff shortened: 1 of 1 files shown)


doc/reference_raster.xml
```diff
@@ -6568,8 +6568,8 @@ FROM (
 
 				<para>When the new scalex or scaley is not a divisor of the raster width or height, the extent of the resulting raster is expanded to encompass the extent of the provided raster. If you want to be sure to retain exact input extent see <xref linkend="RT_ST_Resize" /></para>
 
-				<para>A maxerror percent of 0.125 is used if no <varname>maxerr</varname> is specified.</para>
-
+				<para>A maxerr is the threshold for transformation approximation by the resampling algorithm (in pixel units).A maxerror percent of 0.125 is used if no <varname>maxerr</varname> is specified, which is the same value used in GDAL gdalwarp utility. if set to zero, no approximation takes place. </para>
```
Komzpa:
```suggestion
				<para><varname>maxerr</varname> is the threshold for transformation approximation by the resampling algorithm (in pixel units). A default of 0.125 is used if no <varname>maxerr</varname> is specified, which is the same value used in GDAL gdalwarp utility. If set to zero, no approximation takes place.</para>
```

SunveerSingh:
Is it okay?

SunveerSingh:
@komzpa can we merge it?

PR #301, opened 2018-09-14
failing test case for #4139

alesuiss:
Test case based on svn-trunk to show ticket [#4139](https://trac.osgeo.org/postgis/ticket/4139) in action. Calls the &&& operator a bunch of times on mixed n-D geometries, first with no index, then with n-D GiST index (dropped and recreated every time). 

Results without the index are consistent as expected; results with the index are different every time.

PR #336, opened 2018-11-20
GiST test case for #4139

Komzpa:
See previous discussion in https://github.com/postgis/postgis/pull/301

References https://trac.osgeo.org/postgis/ticket/4139

### DIFF
--- a/postgis/gserialized_gist_nd.c
+++ b/postgis/gserialized_gist_nd.c
@@ -164,10 +164,12 @@ gidx_merge(GIDX **b_union, GIDX *b_new)
 	Assert(b_new);
 
 	/* Can't merge an unknown into any thing */
+	/* Q: Unknown is 0 dimensions. Should we reset result to unknown instead? */
 	if (gidx_is_unknown(b_new))
 		return;
 
 	/* Merge of unknown and known is known */
+	/* Q: Unknown is 0 dimensions. Should we never modify unknown instead? */
 	if (gidx_is_unknown(*b_union))
 	{
 		*b_union = b_new;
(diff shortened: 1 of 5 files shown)


Komzpa:
@Algunenano the problem was indeed during index construction. gidx_merge was adding more constraining dimensions instead of dropping unshared.

Tests pass their counts, but I get memory warnings right in Postgres output that I don't yet know how to even approach:
```
--- regress_gist_index_nd_expected      2018-11-21 11:04:05.293839600 +0300
+++ /tmp/pgis_reg/test_66_out   2018-11-22 01:06:45.174052881 +0300
@@ -1,3 +1,10 @@
+WARNING:  detected write past chunk end in GiST temporary context 0x5635287b01a0
+WARNING:  detected write past chunk end in GiST temporary context 0x5635287b0168
+WARNING:  detected write past chunk end in GiST temporary context 0x5635287afcc8
+WARNING:  detected write past chunk end in GiST temporary context 0x5635287b0108
+WARNING:  detected write past chunk end in GiST temporary context 0x5635287afc68
+WARNING:  problem in alloc set GiST temporary context: bad single-chunk 0x5635287b0140 in block 0x5635287ad570
+WARNING:  problem in alloc set GiST temporary context: found inconsistent memory block 0x5635287ad570
 &&&|180502|Seq Scan|180502|Index Scan
 ~~ |39682|Seq Scan|39682|Index Scan
 @@ |39682|Seq Scan|39682|Index Scan
/tmp/pgis_reg/test_66_diff (END)
```

postgis/gserialized_gist_nd.c
```diff
@@ -180,15 +179,16 @@ gidx_merge(GIDX **b_union, GIDX *b_new)
 
 	POSTGIS_DEBUGF(4, "merging gidx (%s) into gidx (%s)", gidx_to_string(b_new), gidx_to_string(*b_union));
 
-	if (dims_new > dims_union)
+	/* Shrink unshared dimensions */
+	if (dims_new < dims_union)
```
Algunenano: Is this shrinking really necessary now? If b_union has 10 dimensions and b_new 2, I'd expect the merge to have 10, not 2; that is, a merge between 2 boxes should be bigger, never smaller.

Komzpa: Think of 3DM box: it has (minx, maxx, miny, maxy, -FLT_MAX, FLT_MAX, minm, maxm ) for padding, it is a 4D gidx.

When this gets into calculations here, it resets whatever was there in Z to -flt_max / flt_max, in gidx_merge logic. That is equivalent of dropping the dimension. Thus, unshared dimensions should be dropped to correctly summarize a group of other mixed-dimensions boxes. 

Is there a fault in this logic?

Algunenano: > Think of 3DM box: it has (minx, maxx, miny, maxy, -FLT_MAX, FLT_MAX, minm, maxm ) for padding, it is a 4D gidx.

That's what I wasn't expecting. I'd expect that a XYM GBOX would result in a 3-dimension gidx, but as you say it's done by adding an infinite dimension instead [1]. Based on the preferences taken ([2]), I'm unsure whether `gidx_merge` the union should contain that extra dimension or not, but it appears that this was in fact the issue here.

[1] https://github.com/postgis/postgis/blob/svn-trunk/libpgcommon/gserialized_gist.c#L275
[2] https://lists.osgeo.org/pipermail/postgis-devel/2015-February/024759.html

Algunenano:
Great job with this @Komzpa !

PR #337, opened 2018-11-20
[WIP] Travis: Enable UBSan with clang

Algunenano:


Algunenano:
@dbaston @Komzpa What's the proper way to include a recent release of clang into the docker images?

It's not urgent since I'm expecting the tests to crash :smile:

Komzpa:
@Algunenano commit it as dependency into Dockerfile here:

https://github.com/postgis/postgis-build-env

and get someone with push access to Dockerhub run `python3 build.py`. I can do this in the evening. 

(If you want a fun engineering task, Travis has push rights and credentials set up, but build takes more than Travis limit - if you can chop it into quasi-independent stages faster tha then you can do it without me or @dbaston).

Komzpa:
clang now in image. I restarted the build but it fails weirdly.
```
gcc: error: unrecognized command line option '-fsanitize-trap=undefined'; did you mean '-fsanitize=undefined'?
```

Algunenano:
I've changed it to use the old deprecated flag (that is common with gcc) and now it's working and crashing as expected. Thanks a lot @Komzpa

Algunenano:
> Question is, why is it switching to gcc when clang was asked explicitly.

Because Postgres was compiled with GCC and PGXS enforces it. The parts of Postgis that deal with Postgresql itself are always built using Postgres flags (both CC and CFLAGS).

Komzpa:
In this case, would it be beneficial to build a second version of the image, where Postgres is built with clang, so we can also check for UB in that code with clang's ubsan?

Algunenano:
> In this case, would it be beneficial to build a second version of the image, where Postgres is built with clang, so we can also check for UB in that code with clang's ubsan?

Yes, I haven't found any issue around that code in my local tests, but I wouldn't discard it in the future. I'd also take advantage of the change and enable JIT by building Postgres with `--with-llvm`.

PR #338, opened 2018-11-20
ST_3DDistance: Handle invalid polygons

Algunenano:
- Added guards in "measures.h" since it didn't have them.
- Simplified `define_plane` to calculate the plane with the first 3 non collinear points of the ptarray.
- Added tests in liblwgeom as it was easier to debug.
- Avoid division by zero in lw_dist3d_ptarray_poly when the point is the same plane as the polygon.
- Add a bunch of extra tests (some copied from the 2d tests, some extras around the same plane issue).
- When a 3d polygon is defined by collinear points, consider it as a 3d linestring (by ignoring any inner ring).

Trac issue: https://trac.osgeo.org/postgis/ticket/4246

### DIFF
--- a/liblwgeom/measures3d.c
+++ b/liblwgeom/measures3d.c
@@ -1167,13 +1167,14 @@ int
 define_plane(POINTARRAY *pa, PLANE3D *pl)
 {
 	const uint32_t POL_BREAKS = 3;
+	uint32_t unique_points = pa->npoints - 1;
 	uint32_t i;
 
 	/* Calculate the average point */
 	pl->pop.x = 0.0;
 	pl->pop.y = 0.0;
 	pl->pop.z = 0.0;
-	for (i = 0; i < pa->npoints; i++)
+	for (i = 0; i < unique_points; i++)
 	{
 		POINT3DZ p;
 		getPoint3dz_p(pa, i, &p);
@@ -1182,9 +1183,9 @@ define_plane(POINTARRAY *pa, PLANE3D *pl)
 		pl->pop.z += p.z;
 	}
 
-	pl->pop.x /= pa->npoints;
-	pl->pop.y /= pa->npoints;
-	pl->pop.z /= pa->npoints;
+	pl->pop.x /= unique_points;
+	pl->pop.y /= unique_points;
+	pl->pop.z /= unique_points;
 
 	pl->pv.x = 0.0;
 	pl->pv.y = 0.0;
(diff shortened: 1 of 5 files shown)


liblwgeom/measures3d.c
```diff
@@ -58,7 +59,7 @@ get_3dcross_product(VECTOR3D *v1,VECTOR3D *v2, VECTOR3D *v)
 	v->y=(v1->z*v2->x)-(v1->x*v2->z);
 	v->z=(v1->x*v2->y)-(v1->y*v2->x);
 
-	return LW_TRUE;
+	return ((fabs(v->x) > DBL_EPSILON) || (fabs(v->y) > DBL_EPSILON) || (fabs(v->z) > DBL_EPSILON));
```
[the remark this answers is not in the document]
Algunenano: :+1: I didn't remember about that macro

liblwgeom/measures3d.c
```diff
@@ -1133,63 +1164,46 @@ the plane is stored as a point in plane (plane.pop) and a normal vector (plane.p
 int
 define_plane(POINTARRAY *pa, PLANE3D *pl)
```
nicklasaven: I am not sure about the simplification of define_plane. The reason I did it this complicated was an attempt to avoid small numbers giving large errors. If the first 3 points is very close to each other a very small error (floating point error) can make the plane quite wrong which can give quite big error if the polygon is large. This is far away in my head no, since it was a long time ago. I will look more into it.

Algunenano: Maybe we could try to add some test with really close numbers (but bigger than FP_TOLERANCE) to see if it reproduces it.

nicklasaven: Yes, I think we need to think about stability for cases like that and also "dirty" polygons, that has more or less coplanar vertex points.

Algunenano: BTW, since now when we can't calculate the plane it defaults to checking the external ring itself as a line, we could add an extra check in `define_plane` and ignore any point whose distance to the previous ones is less than FP_TOLERANCE or something like this.

nicklasaven:
Here is the thoughts behind how the define_plane function is designed now.

https://trac.osgeo.org/postgis/wiki/3DDistancecalc

nicklasaven:
Yes, I think it looks good. If I understand it right the functionality is the same as before, but you use a fixed 3 samples instead of my walabout ending with 3-5 samples. Is that correct?
If line 1197 - 1202 is safe from any pitfalls I think it looks very good. Much cleaner than before.
I would have put some floor in the above mentioned rows just because I am not that confident about the integer divisions. But if this is safe it is great.

Algunenano:
Looking into it again I see that I'm using npoints instead of (npoints - 1) so I'm repeating the first point (equal to last). I'll change that before merging.

Int division always truncates. Floor would just cast it to float and back, which hopefully will be optimized away by the compiler and do nothing ;D

PR #343, opened 2018-11-29
list multisurface support in linetocurve

stvno:
The documentation doesn't mention that ST_LineToCurve can also convert a MultiSurface into a MultiPolygon. I only discovered this by accident, and one of the major Dutch OpenData sets uses MultiSurfaces so this issue will pop up more often now.


Algunenano:
I guess that you are referencing `ST_CurveToLine`, not `ST_LineToCurve` right? The patch applies to that.

How do you want to appear in the commit log (Name / Github username / other)?

stvno:
Oh indeed I meant `ST_CurveToLine` 

"Steven Ottens" is fine

Algunenano:
Merged. Thanks!

PR #345, opened 2018-11-30
Remove SFCGAL ST_Area

Komzpa:
https://trac.osgeo.org/postgis/ticket/4258


robe2:
@Komzpa   I approved too early.  Looks like you are mixing a deprecation restructure in here.

Those should be done as separate commits.

Also put in notes that ST_Area no longer uses SFCGAL when sfcgal backend is in place.  Though I guess if we are removing the GUC, we would have a big splashing about the backend gone.

PR #354, opened 2018-12-16
DACCEPT_USE_OF_DEPRECATED_PROJ_API_H for proj6

Komzpa:
https://trac.osgeo.org/postgis/ticket/4254

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -825,7 +825,7 @@ if test ! "x$PROJDIR" = "x"; then
 			AC_MSG_RESULT([Using user-specified proj directory: $PROJDIR])
 
 			dnl Add the include directory to PROJ_CPPFLAGS
-			PROJ_CPPFLAGS="-I$PROJDIR/include"
+			PROJ_CPPFLAGS="-I$PROJDIR/include -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H"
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -825,19 +825,22 @@ if test ! "x$PROJDIR" = "x"; then
...
 	fi
 fi
 
-
 dnl Check that we can find the proj_api.h header file
 CPPFLAGS_SAVE="$CPPFLAGS"
 CPPFLAGS="$PROJ_CPPFLAGS"
-AC_CHECK_HEADER([proj_api.h], [], [AC_MSG_ERROR([could not find proj_api.h - you may need to specify the directory of a PROJ.4 installation using --with-projdir])])
+AC_CHECK_HEADER([proj_api.h], [], [AC_MSG_ERROR([could not find proj_api.h - you may need to specify the directory of a PROJ.4 installation using --with-projdir]),
+[
```
Algunenano: There is an error here, the error clause goes until the end so the `include` clause isn't being used. It should be good with: 
```
dnl Check that we can find the proj_api.h header file
AC_CHECK_HEADER([proj_api.h], [], [AC_MSG_ERROR([could not find proj_api.h - you may need to specify the directory of a PROJ.4 installation using --with-projdir])],
[
#define ACCEPT_USE_OF_DEPRECATED_PROJ_API_H 1
])
```

PR #358, opened 2019-01-14
Mvt serialized

Algunenano:
Implements https://trac.osgeo.org/postgis/ticket/4294

I've also moved input checking to the input function and added tests around it


postgis/lwgeom_out_mvt.c
```diff
@@ -45,26 +45,76 @@ Datum ST_AsMVTGeom(PG_FUNCTION_ARGS)
...
+		{
+			/* Shortcut to drop geometries smaller than the resolution */
+			double geom_width = gserialized_box.xmax - gserialized_box.xmin;
+			double geom_height = gserialized_box.ymax - gserialized_box.ymin;
+			double geom_area = geom_width * geom_height;
+
+			double bounds_width = (bounds->xmax - bounds->xmin) / extent;
+			double bounds_height = (bounds->ymax - bounds->ymin) / extent;
+
+			/* We use half the area of the grid square as minimun resolution */
```
Komzpa:
```suggestion
			/* We use half the area of the grid square as minimum resolution */
```

postgis/lwgeom_out_mvt.c
```diff
@@ -45,26 +45,76 @@ Datum ST_AsMVTGeom(PG_FUNCTION_ARGS)
...
 	if (PG_ARGISNULL(1))
+	{
 		elog(ERROR, "%s: parameter bounds cannot be null", __func__);
-	bounds = (GBOX *) PG_GETARG_POINTER(1);
+		PG_RETURN_NULL();
+	}
+	bounds = (GBOX *)PG_GETARG_POINTER(1);
+	if (bounds->xmax - bounds->xmin <= 0 || bounds->ymax - bounds->ymin <= 0)
+	{
+		elog(ERROR, "%s: bounds width or height cannot be 0", __func__);
```
Komzpa: why? why not just return null?

postgis/lwgeom_out_mvt.c
```diff
@@ -45,26 +45,76 @@ Datum ST_AsMVTGeom(PG_FUNCTION_ARGS)
...
+		PG_RETURN_NULL();
+	}
+
 	extent = PG_ARGISNULL(2) ? 4096 : PG_GETARG_INT32(2);
+	if (extent <= 0)
+	{
+		elog(ERROR, "%s: extent cannot be 0", __func__);
```
Komzpa: check is <=, message is about =0 sharp. can lead to hours of looking for 0 in code :)

postgis/lwgeom_out_mvt.c
```diff
@@ -45,26 +45,76 @@ Datum ST_AsMVTGeom(PG_FUNCTION_ARGS)
...
+
 	buffer = PG_ARGISNULL(3) ? 256 : PG_GETARG_INT32(3);
 	clip_geom = PG_ARGISNULL(4) ? true : PG_GETARG_BOOL(4);
-	// NOTE: can both return in clone and in place modification so
-	// not known if lwgeom_in can be freed
+
+	geom_in = PG_GETARG_GSERIALIZED_P_COPY(0);
+	type = gserialized_get_type(geom_in);
+
+	/* If possible, peak into the bounding box before deserializating it to discard small geometries
```
Komzpa:
```suggestion
	/* If possible, peek into the bounding box before deserializing it to discard small geometries
```

postgis/lwgeom_out_mvt.c
```diff
@@ -45,26 +45,76 @@ Datum ST_AsMVTGeom(PG_FUNCTION_ARGS)
...
+	 * We don't check COLLECTIONTYPE since that might be a collection of points */
+	if (type == LINETYPE || type == POLYGONTYPE || type == MULTILINETYPE || type == MULTIPOLYGONTYPE)
+	{
+		GBOX gserialized_box;
+		/* We only apply the optimization if the bounding box is available */
+		if (gserialized_read_gbox_p(geom_in, &gserialized_box) == LW_SUCCESS)
```
Komzpa: there was a _peek function that also worked for two-point linestrings without box, maybe use it?

Algunenano: I wanted to avoid repeating work in the common case (where this speedup isn't applied): in that case he bounding box would be calculated first for this check and later when deserializing the geometry.

postgis/lwgeom_out_mvt.c
Komzpa: then still <=0 and =0 mixup at least needs patching :)

postgis/lwgeom_out_mvt.c
Komzpa: here, _peek is only used in trivial cases:

https://github.com/postgis/postgis/blob/d1e5a63aac0078b6699702d8758dc1f0c7714841/liblwgeom/g_serialized.c#L639

I suggest going first two if-branches, not all three. It feels it can be refactored so that _peek calls _read first instead of failing on present box, and here you just call a _peek.

postgis/lwgeom_out_mvt.c
Algunenano: I've tried but I haven't found a way to create an invalid box (where max < min). Do you prefer using  FP_LTEQ instead?

Komzpa: I see that we're checking for one thing and outputting a message about failing different check. It's never a good idea.

<= 0 are usually checks for math sanity, so there are no things like div0 or negative unsigned area, so FP_* aren't that good for them. If there wasn't such a check here, your later box size shortcut will filter all the geometries, so technically we can just emit a warning here that box is too small but output a valid empty thing instead of crashing the query. WDYT?

postgis/lwgeom_out_mvt.c
Komzpa: Box is not written out for everything that _peek has code for. 

https://github.com/postgis/postgis/blob/682db5db8a2113d41214458143a2f74125ad178e/liblwgeom/lwgeom.c#L1181

postgis/lwgeom_out_mvt.c
Algunenano: > so technically we can just emit a warning here that box is too small but output a valid empty thing instead of crashing the query. WDYT?

I don't like it. Passing a tiny (or invalid) bounding box is clearly an error. I'm not saying that it can't be done, but when I use ST_AsMVT functions I iterate over the geometry, not the bounding box, so in that case I'd end up with an empty tile without a way to know that this was a programming error.

> I see that we're checking for one thing and outputting a message about failing different check. It's never a good idea.

I'll change it to something along the lines of `The geometric bounds are too small`.

PR #361, opened 2019-01-16
ST_Subdivide endless loop on coordinates near coincident to bounds

Komzpa:
https://trac.osgeo.org/postgis/ticket/4301#ticket


Algunenano:
LGTM. In my opinion it'd be a nice improvement if we also tested the difference between the area of the original geometry and the sum of it's parts after subdivide (should be pretty close to 0). What do you think?

PR #366, opened 2019-01-22
Only build static liblwgeom

Algunenano:
Related to https://trac.osgeo.org/postgis/ticket/4260

Since we don't want to link liblwgeom dynamically anymore there is no point in still building it.


Komzpa:
+3.46% code coverage? I take it :)

PR #370, opened 2019-01-30
Update codecov

Algunenano:
This is a test to see if coverage increases using long names in the cov files.


Algunenano:
> Codecov script can now be stripped of repo now that they applied the fix for our duplicate file naes upstream.

Thanks! I'll remove it and leave the rest as is since I don't see any difference with the new long-names option.

Komzpa:
codecov-env is part of Docker build and can't be dropped as such:

https://github.com/postgis/postgis/blob/svn-trunk/.travis.yml#L29

Algunenano:
Merged

PR #371, opened 2019-02-06
Wagyu build improvements

Algunenano:
- If possible (it uses a really basic heuristic) use the same compiler to build wagyu as the one used to build postgresql to reduce friction. Otherwise CXX will be used.
- Add the linker flag to postgis.so for the C++ Standard Library when compiling wagyu. It will tey to use `lc++`, `libstdc++` or nothing in that order.

@pramsey Any chance you could try this with OSX, please? I've managed to compile it in a VM, but I'm still missing a lot of pieces to be able to run the tests. It isn't a very welcoming OS to program from the command line :smile: 

Related to https://trac.osgeo.org/postgis/ticket/4321

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1494,6 +1494,7 @@ if test "x$HAVE_PROTOBUF" = "xyes"; then
 
             WAGYU_CXX=`"$PG_CONFIG" --cc`
             CPPFLAGS="-x c++"
+            CFLAGS=""
             CXX="$WAGYU_CXX"
             AC_PROG_CXX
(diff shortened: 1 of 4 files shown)


pramsey:
This builds for me now.

Algunenano:
> This builds for me now.

Thanks a lot. I'll try to see if I can come up with a better way to "guess" the right compiler; maybe just using  `$(pg_config --cc) -x c++` which works in gcc, clang and icc and would work for builds using full paths.

PR #372, opened 2019-02-08
Improve detection of packages (pkg-config)

Algunenano:
- Use pkg-config if available.
- Only add `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` if using PROJ6.
- Add PROJ_CPPFLAGS to the CPPFLAGS, which should allow us to remove `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H` from the code itself.


pramsey:
I was just looking at what it would take to move to the new `proj.h` API for proj6+ (apparently I cannot `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H`), if you're already there, any interest in doing that? Otherwise, merge at your earliest convenience and I'll carry on.

Algunenano:
> apparently I cannot `ACCEPT_USE_OF_DEPRECATED_PROJ_API_H`

That is odd. Tests with current trunk in travis pass. :man_shrugging: 

> if you're already there, any interest in doing that?

You are not the first to ask me that question lately ;D In my plans, I intend to at least have a look to see how hard it'd be to adapt and what benefits it'd bring, but I don't have the bandwidth right now. 

> Otherwise, merge at your earliest convenience and I'll carry on.

Sadly it's not passing the PROJ6 build (it's optional but it was passing before) as the CPPFLAGS aren't propagated. I'm checking that now.

Algunenano:
So, I decided to do to the same with `json-c`; and the next one is `libpcre`, which is being used in in the address parser.

This all comes from the fact that I expect: `./autogen.sh && ./configure && make` to just work in OSX and it wasn't. I'm tempted to remove the `with-XX=dir` and make pkg-config mandatory but I don't want to screw Windows too much.

PR #373, opened 2019-02-11
Make lwgeom_mindist3d solid aware

Komzpa:
Perform raycast to check if we're completely inside.

Lets us remove SFCGAL counterparts completely.

https://trac.osgeo.org/postgis/ticket/4278


liblwgeom/measures3d.c
```diff
@@ -335,6 +335,58 @@ lwgeom_mindistance3d(const LWGEOM *lw1, const LWGEOM *lw2)
...
 
+static inline int
+gbox_contains_3d(const GBOX *g1, const GBOX *g2)
+{
+	return (g2->xmin >= g1->xmin) && (g2->xmax <= g1->xmax) && (g2->ymin >= g1->ymin) && (g2->ymax <= g1->ymax) &&
+	       (g2->zmin >= g1->zmin) && (g2->zmax <= g1->zmax);
+}
+
+static inline int
+lwgeom_solid_contains_lwgeom(const LWGEOM *solid, const LWGEOM *g)
```
Komzpa: @mhugo can you have a look, is there something obvious that I missed here?

mhugo:
@Komzpa Thanks !
Isn't there a possible problem when the start point of the geometry to test has the same x,y as one of the vertex of the solid ?
In this case, lwpoly_contains_point will return true for each touching polygon, which could double the number of intersections. This will say a point is outside a solid when it is inside.

Komzpa:
@mhugo true, this can happen. Is there a simple way to fix this corner case? It's not enough to just check for vertex equality - a thing can be on some diagonal edge.

Technically ptarray_contains_point inside lwpoly_contains_point can output LW_BOUNDARY as hit test result. Probably I can teach lwpoly_contains_point to pass it outside and then count all the LW_BOUNDARY's as half-hits?

mhugo:
@Komzpa 
> Probably I can teach lwpoly_contains_point to pass it outside and then count all the LW_BOUNDARY's as half-hits?

Half counting is not enough I think. There are situations where the ray may hit the intersection vertex of 3 polygons, or 4 polygons, or even more ...

In 2D, the crossing test for intersection uses an orientation to distinguish edge cases. See http://geomalgorithms.com/a03-_inclusion.html for example.

Another possible issue: what happens if the solid contains a polygon that is parallel to the Z axis and the ray intersects it ? Converting it to 2D will result in a point or linestring that will get ignored, right ?

Komzpa:
Okay. So how about just printing a warning on LW_BOUNDARY with a ticket link then, and deal with it when someone actually hits it? :)

Another option I see is to wrap it into a loop that will perform random rotation of both shapes each time LW_BOUNDARY happens. It feels like in finite-length solids with non-zero distance from interior point to edge there has to be a rotation that will hit only insides of polygons.

Komzpa:
@mhugo check it out - I'm now randomly skewing geometry if ray hits boundary until it doesn't. Do you maybe have a test case?

mhugo:
@Komzpa I don't really like when an algorithm relies on randomness to finish ...
But, apparently CGAL does something similar (see https://doc.cgal.org/latest/Polygon_mesh_processing/classCGAL_1_1Side__of__triangle__mesh.html).

Some test cases that triggers the rotations would be welcomed indeed.

liblwgeom/measures3d.c
```diff
@@ -361,29 +362,66 @@ lwgeom_solid_contains_lwgeom(const LWGEOM *solid, const LWGEOM *g)
...
+					break;
+				}
+			}
+
+			lwcollection_free(c);
+
+			if (is_boundary)
+			{
+				/* randomly skew a bit and restart*/
+				double cx = lwrandom_uniform()-0.5;
```
mwtoews: To make this random, `lwrandom_set_seed(0)` should be done somewhere at the top of the `while` loop.

Algunenano: At the same time, `ST_3DIntersects` can't be IMMUTABLE if it relies on random behaviour. Could always setting the same seed be the way [to go](https://xkcd.com/221/) as it doesn't need true randomness, just a series of pseudo-random numbers?

mhugo: The randomness is not supposed to change the result. It will change the execution time, not the result. So immutable could stay I think.
However, I agree we do not need true randomness here.

Komzpa:
@mhugo added a box with a box on top of it aligned with a box inside along Z. Skew is hit now and resolves the case. Any other cases?

mhugo:
@Komzpa Thanks. I would say a last one when the ray hits a vertical face. Something like this (in 2D):
3dintersects_test

The solid could be:
```
select st_astext(st_extrude(st_translate('polygon((0 0 1,0 0 2,1 0 -1,0 0 -1,-1 0 -1,-1 0 1,0 0 1))',0,-1,0),0,2,0));
=>
POLYHEDRALSURFACE Z (((0 -1 1,-1 -1 1,-1 -1 -1,0 -1 -1,1 -1 -1,0 -1 2,0 -1 1)),((0 1 1,0 1 2,1 1 -1,0 1 -1,-1 1 -1,-1 1 1,0 1 1)),((0 -1 1,0 1 1,-1 1 1,-1 -1 1,0 -1 1)),((-1 -1 1,-1 1 1,-1 1 -1,-1 -1 -1,-1 -1 1)),((-1 -1 -1,-1 1 -1,0 1 -1,0 -1 -1,-1 -1 -1)),((0 -1 -1,0 1 -1,1 1 -1,1 -1 -1,0 -1 -1)),((1 -1 -1,1 1 -1,0 1 2,0 -1 2,1 -1 -1)),((0 -1 2,0 1 2,0 1 1,0 -1 1,0 -1 2)))
```
and the point: `POINT(0 0 0)`

Komzpa:
If I ST_Tesselate the input the code's not yet working, investigating.

Komzpa:
Okay. This works for polyhedrals but not TINS, about which it gets a message. Adding TIN can't be just simple "if POLYGONTYPE || TRIANGLETYPE" as memory layout of TRIANGLE and POLYGON is different (which is not obvious when you count (()) in representation that are aligned with POLYGON).

mhugo:
@Komzpa Thanks for your work

PR #378, opened 2019-02-28
Spatial index support via Pg12 support request API

pramsey:
Using the [support request API](https://github.com/postgres/postgres/blob/4598a99cf22de0dbab975d9c87ce16fd53146aca/src/include/nodes/supportnodes.h) it is possible to add index support to functions without recourse to SQL inlining. So, we can add an "overlaps" index condition to the plan for a query that has `ST_Intersects` in the `WHERE` or `JOIN` clause, invisibly, by implementing an appropriate "support function". 

A side effect of moving to this means of activating indexes is allowing us to re-cost functions with higher cost, but without causing index failures due to [inlining failing](https://carto.com/blog/postgres-parallel/), as we have seen in the past.


postgis/gserialized_supportfn.c
```diff
@@ -0,0 +1,308 @@
...
+
+			if (needsSpatialIndex(funcid, &idxfn))
+			{
+				Node *indexarg, *otherarg, *radiusarg;
+				Oid indexdatatype, otherdatatype;
+				Oid oproid;
+				int nargs = list_length(clause->args);
+
+				/*
+				* Only add an operator condition for GIST and SPGIST indexes.
```
Komzpa: brin mentioned in code?

postgis/gserialized_supportfn.c
```diff
@@ -0,0 +1,308 @@
...
+				oproid = get_opfamily_member(opfamilyoid, indexdatatype, otherdatatype, idxfn.strategy_number);
+				if (oproid == InvalidOid)
+					elog(ERROR, "no spatial operator found for opfamily %u strategy %d", opfamilyoid, idxfn.strategy_number);
+
+				/*
+				* For the ST_DWithin variants we need to build a more complex return.
+				* We want to expand the non-indexed side of the call by the
+				* radius and then apply the operator.
+				* st_dwithin(g1, g2, radius) yields this, if g1 is the indexarg:
+				* g1 && st_expand(g2, radius)
```
Komzpa: geography had _st_expand?

Algunenano:
Wow, i was tracking this but i missed it getting merged upstream.

Do you mind keeping this open for a few days / a week so I can have a look and setup some tests, please?

> A side effect of moving to this means of activating indexes is allowing us to re-cost functions with higher cost, but without causing index failures due to inlining failing, as we have seen in the past.

With this I'm all for updating the cost of all functions and telling people they should use PG12 for proper cost.

OTOH, lately I've been thinking about how could Postgres support costs based on input (for example St_Area costs 5*avg points per geometry in input table squared). Do you know if there is something like that in the horizon (either for PG12 or 13) where we can help/push?

pramsey:
Yes, dynamic costing is available now 

https://github.com/postgres/postgres/blob/4598a99cf22de0dbab975d9c87ce16fd53146aca/src/include/nodes/supportnodes.h#L131-L143

However, in order to be truly dynamic I think we'd need to enhance our stats tables to not only store grids of feature density but also grids of vertex density. Or even just save a single slot for the average vertices/feature. Or I guess maybe there's already enough data available in the stats slot... (object size and number of objects)... can you tell I haven't looked at the stats system in a while?

postgis/gserialized_supportfn.c
pramsey: Hrm, TBD, I see a bunch of inlined functions in geography.sql.in that I have not handled yet.

postgis/gserialized_supportfn.c
```diff
@@ -0,0 +1,254 @@
...
+	expandfn_args[1] = radius_datatype;
+	expandfn_oid = LookupFuncName(expandfn_name, 2, expandfn_args, noError);
+	if (expandfn_oid == InvalidOid)
+		elog(ERROR, "%s: unable to lookup 'st_expand(Oid[%u], Oid[%u])'", __func__, geo_datatype, radius_datatype);
+	return expandfn_oid;
+}
+
+Datum postgis_index_supportfn(PG_FUNCTION_ARGS);
+PG_FUNCTION_INFO_V1(postgis_index_supportfn);
+Datum postgis_index_supportfn(PG_FUNCTION_ARGS)
```
robe2:
```suggestion
PGDLLEXPORT Datum postgis_index_supportfn(PG_FUNCTION_ARGS)
```

The function gets left out of my windows build without the PGDLLEXPORT clause

pramsey: I can add it, but boy I would love to know *why* it works, and why 90% of our functions work without it, but some have it. Any help? I see spgist support functions have it. Actually all other instances of it seem to trace back to spgist. But still: why only for that one symbol?

robe2: I did an experiment of moving the 

Datum postgis_index_supportfn(PG_FUNCTION_ARGS);

declarator

to the top of the file right after the includes and that seems to work such that I don't need the PGDLLEXPORT later.

That might be why our others have worked because the header defs were at the top.  Seems very fragile I know.

pramsey: Weird.

pramsey:
@strk I really need your help figuring out exactly what st_function(geometry,geometry) is causing the topology failures... something that breaks it down to an actual "these inputs are returning an incorrect answer" situation. I cannot tease that out of topology. https://travis-ci.org/postgis/postgis/jobs/500590893#L2870

postgis/gserialized_supportfn.c
pramsey: Geography has been brought into the fold now too.

postgis/geography_measurement.c
```diff
@@ -267,10 +268,86 @@ Datum geography_distance(PG_FUNCTION_ARGS)
...
+	GSERIALIZED *g2 = PG_GETARG_GSERIALIZED_P(1);
+	double tolerance = 0.0;
+	bool use_spheroid = true;
+	bool dwithin = geography_dwithin_impl(fcinfo, g1, g2, tolerance, use_spheroid);
+	PG_FREE_IF_COPY(g1, 0);
+	PG_FREE_IF_COPY(g2, 1);
+	PG_RETURN_BOOL(dwithin);
+}
+
+#if 0
```
Komzpa: can this be removed?

postgis/gserialized_supportfn.c
```diff
@@ -0,0 +1,372 @@
...
+	const char *fn_name;
+	int   strategy_number; /* Index strategy to add */
+	int   nargs;           /* Expected number of function arguments */
+	int   expand_arg;      /* Radius argument for "within distance" search */
+} IndexableFunction;
+
+/*
+* Metadata currently scanned from start to back,
+* so most common functions first. Could be sorted
+* and searched with binary search.
```
Komzpa: what if we make many support functions, for each original - will it be even faster without a check?

postgis/geography_measurement.c
pramsey: will do

postgis/gserialized_supportfn.c
pramsey: We could make one per strategy, but honestly it's not worth it. The function is called once at the planning stage, so this wee bit of inefficiency is not an issue.

PR #386, opened 2019-03-25
Remove errorIfGeometryCollection checking on file lwgeom_geos.c

estebanzimanyi:
Remove the restriction that functions ST_Equal, ST_Contains, ST_Relate, etc. cannot accept GeometryCollection arguments. This was a restriction of an older version of GEOS that according to my understanding is no longer true.


Algunenano:
Can you add tests (either in SQL/regress or in C/cunit) to check this, please? Then we can rely on travis to make sure it's available in GEOS 3.6+

Algunenano:
Hi,

You are modifying at least 11 functions: 
```
  C function       -- SQL function
- overlaps         -- ST_Overlaps(geom1 geometry, geom2 geometry)
- contains         -- ST_Contains(geom1 geometry, geom2 geometry)
- containsproperly -- ST_ContainsProperly(geom1 geometry, geom2 geometry)
- covers           -- ST_Covers(geom1 geometry, geom2 geometry)
- coveredby        -- ST_CoveredBy(geom1 geometry, geom2 geometry)
- crosses          -- ST_Crosses(geom1 geometry, geom2 geometry)
- touches          -- ST_Touches(geom1 geometry, geom2 geometry)
- disjoint         -- ST_Disjoint(geom1 geometry, geom2 geometry)
- relate_pattern   -- ST_Relate(geom1 geometry, geom2 geometry,text)
- relate_full      -- ST_Relate(geom1 geometry, geom2 geometry) // ST_Relate(geom1 geometry, geom2 geometry, int4)
- ST_Equals        -- ST_Equals(geom1 geometry, geom2 geometry)
``` 

Can you please add simple test for all of them (at least one per C function passing geometry collections in all parameters)?

Also, in terms of documentation prefer adding a line like https://github.com/postgis/postgis/blob/ba1baf9f5e3d3ed79fc1eefa0e40915a9803a207/doc/reference_processing.xml#L1172
```
 <para>Enhanced: 3.0.0 enabled support for GeometryCollections</para> 
```

PR #387, opened 2019-03-27
Fix postgis_type_name('GEOMETRYM',3) 

mattbretl:
`postgis_type_name('GEOMETRYM',3)` currently returns null. It should return GeometryM.

This PR contains the fix and an updated regression test.


Algunenano:
Thanks, I've merged it and I'll backport it to 2.3

PR #391, opened 2019-04-15
Test extension on Travis

Komzpa:


strk:
Note that `make installcheck` already does the extension test, no need to override!

strk:
Rather, it would help if travis did run _upgrade_ tests (currently only run by other bots which not many people like to listen to...)

ci/travis/run_tests.sh
```diff
@@ -16,4 +16,4 @@ export CUNIT_VALGRIND_FLAGS="--leak-check=full --error-exitcode=1"
 
 # Standard build
 ./configure CFLAGS="${CFLAGS_STD}" LDFLAGS="${LDFLAGS_STD}"
-bash ./ci/travis/logbt -- make -j check RUNTESTFLAGS="--verbose --extension"
+bash ./ci/travis/logbt -- make -j installcheck RUNTESTFLAGS="--verbose"
```
strk: We want _both_ `make check` and `make installcheck`. The first checks the installation via manual scripts, the second checks via extension

strk: Of course the `installcheck` must be done _after_ `make install`

Komzpa:
we have drone now.

PR #398, opened 2019-04-24
Adding ubsan implicit conversion sanitizer to clang

Algunenano:
WIP.
- https://github.com/postgis/postgis-build-env/pull/3 should to be merged too :pray: 
- I've already cleaned the errors in liblwgeom and core regress, but raster and topology is still pending,  thus that build doesn't pass the tests yet.

### DIFF
--- a/raster/rt_core/rt_statistics.c
+++ b/raster/rt_core/rt_statistics.c
@@ -113,7 +113,7 @@ rt_band_get_summary_stats(
 	uint64_t *cK, double *cM, double *cQ
 ) {
 	uint32_t x = 0;
-	uint32_t y = 0;
+	int64_t y = 0;
 	uint32_t z = 0;
 	uint32_t offset = 0;
 	uint32_t diff = 0;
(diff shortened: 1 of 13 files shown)


Algunenano:
Raster is now passing, I'm addressing topology/ next.

raster/rt_core/rt_statistics.c
```diff
@@ -1032,7 +1032,7 @@ rt_band_get_quantiles_stream(
 	uint32_t j = 0;
 	uint32_t k = 0;
 	uint32_t x = 0;
-	uint32_t y = 0;
+	int64_t y = 0;
```
Komzpa: why is x different from y?

Algunenano: x goes from 0 to `band->width` but y is initially setup as  `-1` and then it can either be replaced by an uint32_t value or added to a random bounded uint32_t. 

To be honest, it is a 500 LOC function and reviewing how and this could be changed to use unsigned would take a lot of time. I'm good with doubling the precision and keeping the sanitizer happy.

Komzpa:
How about we merge this at least in part? that would be better already :)

Algunenano:
It's on my list. I indent to finish this (having all warnings/errors addressed) as soon as I can get back to postgis/trunk (soon :tm:)

PR #399, opened 2019-04-25
Improve doc Reference categories

dr-jts:
Add sections Inputs
Add subsections to Inputs, Outputs, Operators, Spatial Relationships
Standardize Constructors and Accessors descriptions
Various minor text changes/fixes

Ticket: https://trac.osgeo.org/postgis/ticket/4332


doc/reference_accessor.xml
```diff
@@ -7,8 +7,7 @@
 	  <refnamediv>
 		<refname>GeometryType</refname>
 
-		<refpurpose>Returns the type of the geometry as a string. Eg:
-			'LINESTRING', 'POLYGON', 'MULTIPOINT', etc.</refpurpose>
+		<refpurpose>Returns the type of the geometry as text.</refpurpose>
```
Algunenano: Small note about consistency (now that you are improving things quite a bit):
- Here it refers to the geometry as `the geometry`.
- In https://github.com/postgis/postgis/pull/399/commits/ef2b137635c4be8dbd1e9045b45a16ccacef0137#diff-f976dc5a4064695bcf05b055d1c2d912R93 as `the Geometry`.
- In https://github.com/postgis/postgis/pull/399/commits/ef2b137635c4be8dbd1e9045b45a16ccacef0137#diff-f976dc5a4064695bcf05b055d1c2d912R210 and https://github.com/postgis/postgis/pull/399/commits/ef2b137635c4be8dbd1e9045b45a16ccacef0137#diff-f976dc5a4064695bcf05b055d1c2d912R264 as `a Geometry`.
- In https://github.com/postgis/postgis/pull/399/commits/ef2b137635c4be8dbd1e9045b45a16ccacef0137#diff-f976dc5a4064695bcf05b055d1c2d912R280 as `this Geometry object`.

Should we care about this?

Komzpa: given that this goes into COMMENT, can we lose articles and get it condensed so it's a bit easier to fit into screen?

doc/reference_accessor.xml
```diff
@@ -331,9 +723,9 @@ ST_Dimension
 	  <refsection>
 		<title>Description</title>
 
-		<para>Returns the last point of a <varname>LINESTRING</varname> geometry
-		as a <varname>POINT</varname> or <varname>NULL</varname> if the input
-		parameter is not a <varname>LINESTRING</varname>.</para>
+		<para>Returns the last point of a LINESTRING
+		as a POINT.  Returns <varname>NULL</varname> if the input
```
Komzpa: Why is POINT removed as varname but LINESTRING isn't?
In any case they're not variables so semantically it's incorrect to mark them this way. Perhaps find some part of the doc that mentions geometry types and link there?

doc/reference_accessor.xml
```diff
@@ -382,8 +774,7 @@ SELECT ST_AsEWKT(ST_EndPoint('LINESTRING(1 1 2, 1 2 3, 0 0 5)'));
 		  <refnamediv>
 			<refname>ST_Envelope</refname>
 
-			<refpurpose>Returns a geometry representing the double precision (float8) bounding box of the
-			supplied geometry.</refpurpose>
+			<refpurpose>Returns a geometry representing the double precision (float8) bounding box of a geometry.</refpurpose>
```
Komzpa: float8 is confusing to newbies. "is it less precise than usual or more precise than usual".

doc/reference_accessor.xml
```diff
@@ -658,7 +1046,7 @@ LINESTRING(0 0 1,1 1 1,1 2 1,1 1 1,0 0 1)
 		</note>
 
 		<note>
-		  <para>If you want to extract all geometries, of a geometry, ST_Dump is more efficient and will also work for singular geoms.</para>
+		  <para>If you want to extract all geometries, of a geometry, <xref linkend="ST_Dump" /> is more efficient, and also works for singular geoms.</para>
```
Komzpa: thanks for linking ST_Dump!

doc/reference_accessor.xml
```diff
@@ -1184,10 +1568,7 @@ postgis=# SELECT ST_IsClosed('MULTIPOINT((0 0), (1 1))'::geometry);
 	<refentry id="ST_IsCollection">
 	  <refnamediv>
 		<refname>ST_IsCollection</refname>
-
-                <refpurpose>Returns <varname>TRUE</varname> if the argument is a
-                  collection (<varname>MULTI*</varname>, <varname>GEOMETRYCOLLECTION</varname>, ...)
-                </refpurpose>
+      <refpurpose>Tests if a geometry is a geometry collection type.</refpurpose>
```
Komzpa: this change lost an important note that MULTI* is also a collection according to this.

doc/reference_accessor.xml
```diff
@@ -1269,7 +1650,7 @@ postgis=# SELECT ST_IsCollection('GEOMETRYCOLLECTION(POINT(0 0))'::geometry);
 	  <refnamediv>
 		<refname>ST_IsEmpty</refname>
 
-		<refpurpose>Returns true if this Geometry is an empty geometrycollection, polygon, point etc.</refpurpose>
+		<refpurpose>Tests if a geometry is empty.</refpurpose>
```
Komzpa: It's not obvious if collection of empty is also empty.

Komzpa: maybe "presence of any coordinate in geometry object" or something like this?

doc/reference_accessor.xml
```diff
@@ -1345,8 +1726,7 @@ SELECT ST_IsEmpty(ST_GeomFromText('POLYGON((1 2, 3 4, 5 6, 1 2))'));
 	<refnamediv>
 	  <refname>ST_IsRing</refname>
 
-	  <refpurpose>Returns <varname>TRUE</varname> if this
-	  <varname>LINESTRING</varname> is both closed and simple.</refpurpose>
+	  <refpurpose>Tests if a LineString is closed and simple.</refpurpose>
```
Komzpa: interesting if it's a part of spec
```
gis=# select ST_IsRing('POLYGON((0 0, 1 1, 2 3, 0 0))');
ERROR:  ST_IsRing() should only be called on a linear feature
```

doc/reference_accessor.xml
```diff
@@ -1492,20 +1870,25 @@ FROM (SELECT 'LINESTRING(0 0, 0 1, 1 0, 1 1, 0 0)'::geometry AS the_geom) AS foo
 	  <refsection>
 		<title>Description</title>
 
-		<para>Test if an ST_Geometry value is well formed.  For geometries that are invalid,
-			the PostgreSQL NOTICE will provide details of why it is not valid.  For more
-			information on the OGC's definition of geometry simplicity and validity, refer
+		<para>Test if an ST_Geometry value is well-formed according to the OGC rules.
```
Komzpa: "ST_Geometry" shoudn't be a thing.

doc/reference_accessor.xml
```diff
@@ -1633,7 +2016,7 @@ SELECT ST_IsValidReason('LINESTRING(220227 150406,2220227 150407,222020 150410)'
 	  <refnamediv>
 		<refname>ST_IsValidDetail</refname>
 
-		<refpurpose>Returns a valid_detail (valid,reason,location) row stating if a geometry is valid or not and if not valid, a reason why and a location where.</refpurpose>
+		<refpurpose>Returns a <varname>valid_detail</varname> row stating if a geometry is valid, and if not a reason why and a location.</refpurpose>
```
Komzpa: valid_detail probably can be references, it had a page?

doc/reference_accessor.xml
```diff
@@ -1740,8 +2122,8 @@ SELECT * FROM ST_IsValidDetail('LINESTRING(220227 150406,2220227 150407,222020 1
 	  <refsection>
 		<title>Description</title>
 
-		<para>Return the M coordinate of the point, or NULL if not
-			available. Input must be a point.</para>
+		<para>Return the M crdinate of a Point, or NULL if not
```
Komzpa: as a non-native English speaker it's painful to learn word "ordinate". they're all "coordinates" in Russian, "ordinate" is specifically Y axis in version of maths I was taught - "abscissa" is for X. Is "M ordinate" valid thing in English, or can we keep "M coordinate" which also matches Russian usage?
```suggestion
		<para>Return the M ordinate of a Point, or NULL if not
```

doc/reference_accessor.xml
```diff
@@ -1773,11 +2155,79 @@ SELECT * FROM ST_IsValidDetail('LINESTRING(220227 150406,2220227 150407,222020 1
 	  </refsection>
 	</refentry>
 
+	<refentry id="ST_MemSize">
+	  <refnamediv>
+		<refname>ST_MemSize</refname>
+
+		<refpurpose>Returns the amount of memory space a geometry takes.</refpurpose>
```
Komzpa: `bytes, serialized, uncompressed.` it's a size of GSERIZLIZED which gets compressed on disk, and decompressed even more in memory when becoming LWGEOM or GEOS geometry. Many postgres sizes are in 8kb pages which means bytes should be mentioned.

doc/reference_accessor.xml
```diff
@@ -2130,9 +2577,7 @@ GROUP BY gid, field1,field2;
 	  <refnamediv>
 		<refname>ST_PatchN</refname>
 
-		<refpurpose>Return the 1-based Nth geometry (face) if the geometry is a
-			POLYHEDRALSURFACE, POLYHEDRALSURFACEM.
-			Otherwise, return NULL.</refpurpose>
+		<refpurpose>Returns the Nth geometry (face) of a PolyhedralSurface.</refpurpose>
```
Komzpa: 1-based is lost.

doc/reference_input.xml
```diff
@@ -0,0 +1,1822 @@
...
+
+			 <para>&sfs_compliant; s3.2.6.2</para>
+			 <para>Performed by the GEOS module.</para>
+			 <para>Availability: 1.1.0</para>
+		  </refsection>
+
+		  <refsection>
+			<title>Examples</title>
+
+			<programlisting>Forthcoming</programlisting>
```
Komzpa: can we comment out whole section then?

doc/reference_input.xml
```diff
@@ -0,0 +1,1822 @@
...
+				</note>
+
+				 <para>&sfs_compliant; s3.2.6.2</para>
+				 <para>Performed by the GEOS module.</para>
+				 <para>Availability: 1.1.0</para>
+			</refsection>
+
+			<refsection>
+				<title>Examples</title>
+				<programlisting>Forthcoming</programlisting>
```
Komzpa: can we also comment it out?

doc/reference_accessor.xml
dr-jts: @Algunenano Yes. absolutely we should have consistency in referring to geometry, and also specific geometry types.  In aid of this I will create a PostGIS Doc style guide in the repo, as a place for discussion and to capture decisions.

For this particular issue, there's lots of options, but here's my current thinking:
* Refer to general geometry objects as "geometry"
* Refer to specific types as LineString, Polygon, etc
* Not sure about using a different font.  Not doing that is simpler.  At least should not be fonted in the summary text

dr-jts: @Komzpa Yes, the comment should be shorter.  I will fix.

dr-jts: @Komzpa Actually realized I already made this shorter.  Not so keen on removing all articles.  That hasn't been done elsewhere, and there are many longer comments.

doc/reference_accessor.xml
dr-jts: Fixing this.

And good point about reason for not specifying geometry type names as varnames.  (I think the goal was just to get them in a code font, but either this should be done with a different tag, or just not done at all.  This is something for the style guide.

doc/reference_accessor.xml
dr-jts: Fixed in comment.  Will leave the longer more accurate description in the main text.  Newbies can look it up :)

doc/reference_accessor.xml
dr-jts: The detail is described in the function page.  Should the comment be made longer/more descriptive as well?  In general I've been trying to make the summary comments more concise.

And on a related note, there should be a better section describing ALL the geometry types supported by PostGIS, which can be linked to.  This is on my todo list...

doc/reference_accessor.xml
dr-jts: Ok, good to know.

doc/reference_accessor.xml
dr-jts: Not following.  Can you provide some text for this?

doc/reference_accessor.xml
dr-jts: With the goal of making the summaries shorter and simpler I removed the note about being 1-based.  It's in the main function page description.  

Is there a reason to leave it in the summary?

doc/reference_input.xml
dr-jts: Yes, makes sense.

The descriptions for these two functions need a lot of work...

doc/reference_input.xml
dr-jts: Yes

dr-jts:
Thanks for the comments all.  If there's no further discussion needed, I will work on committing this.

Komzpa:
https://github.com/postgis/postgis/commit/62f01fbb2efc2c9eb18293ca36d41b7acbf13209 this was merged? can this be closed?

PR #403, opened 2019-05-09
Upgrade ST_Union without drop

Komzpa:
based on https://git.osgeo.org/gitea/postgis/postgis/pulls/30 test and prior implementation of parallel safe patch up.


postgis/postgis_drop_after.sql
```diff
@@ -242,3 +242,5 @@ IF _postgis_scripts_pgsql_version()::integer >= 96 THEN
 END IF;
 END;
 $$;
+
+UPDATE pg_aggregate SET aggtransfn = 'pgis_geometry_union_transfn', aggtranstype='internal'::regtype where aggfnoid='st_union(geometry)'::regprocedure;
```
strk: I don't like the implications of such a thing.
Conceptually, an UPDATE does not go in a "drop_after".

Rather, if you want to do things manually (I'm not against the "changed but upgrader helper no touch" idea), the manual approach could go in the main script, with some conditional code in a DO block.

But if this could be done all directly with an upgrade helper I'd feel much better...

Algunenano:
The signature change was reverted, so we "only" kept previous backwards incompatible changes which, if we keep going, will belong to deprecated releases soon :smiley: Should we close this @Komzpa ?

PR #404, opened 2019-05-15
Make TIN default output for Delaunay Triangles.

Komzpa:
https://trac.osgeo.org/postgis/ticket/4393#ticket

### DIFF
--- a/doc/reference_processing.xml
+++ b/doc/reference_processing.xml
@@ -1114,11 +1114,12 @@ Return a <ulink
 url="http://en.wikipedia.org/wiki/Delaunay_triangulation">Delaunay
 triangulation</ulink> around the vertices of the input geometry.
 Output is a COLLECTION of polygons (for flags=0) or a MULTILINESTRING
-(for flags=1) or TIN (for flags=2).  The tolerance, if any, is used to snap input vertices
+(for flags=1) or TIN (for flags=2, default).  The tolerance, if any, is used to snap input vertices
 togheter.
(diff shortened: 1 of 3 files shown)


dbaston:
I think we decided not to do this, right?

Komzpa:
Reason given by @pramsey was "there's no reason to give users TIN if there are no TIN-specific functions that they can work with". Then we discussed what those functions can be, and I ticketed them:

 - https://trac.osgeo.org/postgis/ticket/4397 drape
 - https://trac.osgeo.org/postgis/ticket/4396 3DSlope
 - https://trac.osgeo.org/postgis/ticket/4400 Isolines
 - https://trac.osgeo.org/postgis/ticket/4401 ST_Boundary(TIN)
 - https://trac.osgeo.org/postgis/ticket/4399 support in output

As I understand merging this should be revisited after (at least some of) those are done.

pramsey:
You may have read more into my interest in TIN-consuming functions than I meant: I think more TIN consuming functions is a great idea, but I think changing the output signatures of long-standing functions is generally not, unless we're pretty sure we're not going to be harming existing users of said functions or placing more burden on existing users just for our own particular use cases. The only output change I can remember is the 2.0 change in WKT/WKB output to the (standard) ISO format, which only affected people with 3D outputs (a very small sub-set of WKT/WKB users).

robe2:
I'm closing this as a No no buddy @Komzpa

PR #406, opened 2019-05-31
Wagyu on

Algunenano:


robe2:
@Algunenano I tried building --with-wagyu on and running into errors


https://trac.osgeo.org/postgis/ticket/4408

Any thoughts on this -- we can enable wagyu on winnie to try to troubleshoot this.  I'm fine with this not being a show-stopper for committing this change.

Then we could see if others run into the issue.

Algunenano:
@robe2 I can push this change this weekend if you want to release the alpha2.

Also, it's pretty easy to disable wagyu for mingw in the meantime if you think that's better until I have time to investigate the issue.

robe2:
Yah I'll disable it before alpha2 release if I can't figure it out.

In meantime I have freebsd (bessie bots) and debbie now configured with mvt and wagyu.  Bessie runs passed.  I'm sure debbie's will pass too.

So I think we should commit this patch.  I'll do when I get up if you don't get to it before then.

robe2:
maybe I'll just do it now :)

PR #407, opened 2019-06-02
Support NODATA=NaN in Raster

Komzpa:
https://trac.osgeo.org/postgis/ticket/4412

Helps import Facebook population dataset.

### DIFF
--- a/raster/loader/raster2pgsql.c
+++ b/raster/loader/raster2pgsql.c
@@ -492,11 +492,8 @@ static void calc_tile_size(
 			}
 			r = r - (double) d;
 
-			if (
-				FLT_EQ(_r, -1.0) ||
-				(r < _r) ||
-				FLT_EQ(r, _r)
-			) {
+			if (FLT_EQ(_r, -1.0) || (r < _r) || FLT_EQ(r, _r))
+			{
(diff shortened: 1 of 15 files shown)


robe2:
See my note - https://trac.osgeo.org/postgis/ticket/4412

in short - I'm okay with this in 3.0, but introducing it in 2.5.3 feels a little dirty as it risks disstabilizing a stable release.

dbaston:
Seems OK in concept, but I didn't review the mixed formatting and logic changes. Should be noted as a breaking change in NEWS, I think. Is `ST_PixelAsPolygons` really the only affected function?

Komzpa:
@dbaston formatting is specially a separate commit, https://github.com/postgis/postgis/pull/407/commits/7c2366715d9658b07fa65bb078d4c97ae1dae586 shows change without formatting (mostly 0 -> 0.0, as isnan(0) is not possible).

With ST_PixelAsPolygons we have at least one way of escaping from such raster into vector land. 

It is not a breaking change as there's no working behavior we break, people who had a raster with NODATA NaN can't work with it at all using PostGIS - it comes out all-NODATA which is an obvious breakage.

PR #408, opened 2019-06-03
Cloud Optimized GeoTIFF tile size autodetect

Komzpa:
Use tile sizes aligned with GeoTIFF "Natural Blocks" to speed up import 10x.


robe2:
Looks fine to me

dbaston:
I think some tests might be helpful here.

This isn't strictly related to COGs, is it? It's defaulting to multiples of the block size, whatever that is?

Komzpa:
It's not about cloud optimized only but it also saves much time on them:
```
kom@nucat:~/downloads$ time ~/proj/postgis/raster/loader/raster2pgsql /vsicurl/https://github.com/postgis/postgis/raw/690cdd624b6d5ad62c497481832bbf0425457687/raster/test/regress/loader/testraster2.tif  -t auto 0_fb_africa_population_raster >/dev/null
Processing 1/1: /vsicurl/https://github.com/postgis/postgis/raw/690cdd624b6d5ad62c497481832bbf0425457687/raster/test/regress/loader/testraster2.tif
INFO: Using computed tile size: 256x192

real    0m13,444s
user    0m1,901s
sys     0m0,105s
kom@nucat:~/downloads$ time raster2pgsql /vsicurl/https://github.com/postgis/postgis/raw/690cdd624b6d5ad62c497481832bbf0425457687/raster/test/regress/loader/testraster2.tif  -t auto > /dev/nullProcessing 1/1: /vsicurl/https://github.com/postgis/postgis/raw/690cdd624b6d5ad62c497481832bbf0425457687/raster/test/regress/loader/testraster2.tif
INFO: Using computed tile size: 49x49

real    3m12,630s
user    0m12,546s
sys     0m3,494s

```

dbaston:
I guess my thought is that if it's defaulting to the block size, whatever that is, you might end up with pretty bad tile sizes for rasters that are written as scanlines. (By "bad" I mean that they would perform poorly for spatial queries.)

Komzpa:
@dbaston scanline case is guarded by taking image dimension instead of block dimension if it is 1.

PR #409, opened 2019-06-03
mark = as HASHES, MERGES

Komzpa:
Thanks RhodiumToad on IRC.
TODO: upgrades


Komzpa:
```
[RhodiumToad] for upgrading an existing install you might be better off updating the catalog to set oprcanhash/oprcanmerge
```

Algunenano:
Operator's `hashes` and `merges` can't be updated with ALTER OPERATOR:
```
# alter operator = ( geometry, geometry ) set (HASHES = true); 
ERROR:  operator attribute "hashes" cannot be changed

alter operator = ( geometry, geometry ) set (MERGES = true); 
ERROR:  operator attribute "merges" cannot be changed
```

But it can be done with something like:
```sql
update pg_operator set oprcanhash = true, oprcanmerge = true where oprname = '=' and oprcode = 'geometry_eq'::regproc;
```

I haven't investigated what could be the impact of doing this on the fly, just that it is possible and things work after doing it.

PR #410, opened 2019-06-06
Modernize Python 2 code to get ready for Python 3

cclauss:
Use __print()__ function in both Python 2 and Python 3
* Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

Old style exceptions --> new style for Python 3
* Old style exceptions are syntax errors in Python 3 but new style exceptions work as expected in both Python 2 and Python 3.

### DIFF
--- a/raster/scripts/python/genraster.py
+++ b/raster/scripts/python/genraster.py
@@ -26,6 +26,7 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 ###############################################################################
+from __future__ import print_function
 import Image
 import ImageDraw
 import ImageFont
@@ -33,8 +34,8 @@ import random
 import sys
 
 if len(sys.argv) < 5 or len(sys.argv) > 6:
-    print 'Usage: genraster.py <xsize> <ysize> <xsizecell> <ysizecell> <outline colour>'
-    print 'Note: Generated image is saved as out.png'
+    print('Usage: genraster.py <xsize> <ysize> <xsizecell> <ysizecell> <outline colour>')
+    print('Note: Generated image is saved as out.png')
     sys.exit(1)
 
 g_file = "out.png"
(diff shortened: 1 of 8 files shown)


raster/scripts/python/rtrowdump.py
```diff
@@ -130,5 +130,5 @@ def pt2numpy(pt):
         assert band is not None
         band.WriteArray(raster)
 
-except rtreader.RasterError, e:
+except rtreader.RasterError as e:
     print "ERROR - ", e
```
mwtoews: this should be `print("ERROR - ", e)`

raster/scripts/python/genraster.py
```diff
@@ -26,15 +26,16 @@
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
 ###############################################################################
+from __future__ import print_function
```
mwtoews: This import and similar below are not needed, as I don't see any use of `print()` function [as documented](https://docs.python.org/3/library/functions.html#print).

cclauss: This directive will [prevent compatibility regressions](https://docs.python.org/3/howto/pyporting.html#prevent-compatibility-regressions) by changing Python 2 so that it is a syntax error to use legacy __print__ statements.  This prevents Python 2 developers from breaking Python 3 compatibility.

raster/scripts/python/rtrowdump.py
cclauss: Good catch.  Thanks.  Fixed.

raster/scripts/python/genraster.py
mwtoews: these imports are fine and can stay. Python 2 [will not be maintained past 2020](https://pythonclock.org/), so I won't expect too many Python 2 developers down the road.

pramsey:
I'm on expert, but seems fine to me. Going to apply?

PR #411, opened 2019-06-06
Travis CI: Lint Python code for syntax errors

cclauss:
These test will fail until #410 (or similar) is landed.

Also the __sudo:__ tag is now deprecated in Travis CI: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

"_If you currently specify __sudo:__ in your __.travis.yml__, we recommend removing that configuration_"

@mwtoews Your review please?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 services:
   - docker
 
-sudo: true
 language: c
 
 env:
@@ -20,6 +19,11 @@ env:
 matrix:
   allow_failures:
     - env: tag=latest mode=tests
+  include:
+    - language: python
+      script:
+        - pip install flake8  # lint Python code for syntax errors
+        - flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
 
 script:
   - echo "/tmp/logbt-coredumps/core.%p.%E" | sudo tee /proc/sys/kernel/core_pattern
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ matrix:
   allow_failures:
     - env: tag=latest mode=tests
   include:
-    - language: python
+    - name: "Lint: python3 -m flake8 . --select=E9,F63,F72,F82"
+      language: python
       script:
         - pip install flake8  # lint Python code for syntax errors
         - flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -39,7 +39,7 @@ clean-local:
 # TODO: drop 'test' target..
 test: check
 
-check: all docs-check check-no-trailing-blanks
+check: all docs-check check-no-trailing-blanks lint-python
 
 check-no-trailing-blanks:
 	! find . -name '*.c' -o -name '*.h' -o -name '*.proto' | \
@@ -48,6 +48,10 @@ check-no-trailing-blanks:
 		grep -v vector_tile.pb-c | \
 		grep -v postgis/sqldefines.h | \
 		xargs grep -n '[[:space:]]$$'
+
+lint-python:
+	python3 -m pip install --upgrade flake8
+	flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
 
 installcheck: installcheck-base installcheck-upgrade
 
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -39,7 +39,7 @@ clean-local:
 # TODO: drop 'test' target..
 test: check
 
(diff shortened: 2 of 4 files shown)


Komzpa:
Your quote has been edited from original, `If you currently specify sudo: false in your .travis.yml, we recommend removing that configuration soon.`. We're not using sudo:false, we're using sudo:true for logbt to work.

cclauss:
See python/mypy#6675

Komzpa:
Can this please be made not a part of Travis script, but part of normal `make check` invocation?

cclauss:
Where is __make check__ called within this Travis workflow?

Komzpa:
here's example: https://github.com/postgis/postgis/blob/svn-trunk/GNUmakefile.in#L44

call itself is https://github.com/postgis/postgis/blob/svn-trunk/ci/travis/run_tests.sh#L19

cclauss:
I am not such a fan of the new approach because we are now redundantly running the same flake8 tests eight times for each Travis run.  Is that what we want?

cclauss:
Oh.  Plus we are failing __drone__ too because their Python does not have pip installed.

dbaston:
I think it's best left out of `make check`. I don't think having the `flake8` package installed is a reasonable requirement to test PostGIS, and `make check` should certainly not install things into a user's system python library. I don't have any issue with the original approach, though I'm not sure I see the necessity to check python syntax after every commit.

cclauss:
Thanks @dbaston.  I agree that the __make check__ approach is problematic so I have reverted it.

The Travis CI Python run takes 42 seconds while the other runs take >6 minutes.  ```python goes last and it goes fast so given the parallel execution that Travis provides, these tests should rarely extend build times.  In those other runs we are compiling the C code which helps us find “compile-time errors”.  In the Python run we are using the flake8 linter to build an Abstract Syntax Tree (AST) for each Python file.  This is as close as we can get to a compiler for an interpreted language like Python.  Like a compiler, flake8 is going to use that AST to find variables that are used before they are defined, syntax errors, etc.  Just like the C code, we would like to find crashing changes made to Python files before pull requests are reviewed.  We want to remove the syntax errors that are currently in the codebase and we want to be warned if new syntax errors are added if future PRs.

Komzpa:
closed and reopened to check again after #410 merged

Komzpa:
```
1.21s$ flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
./raster/scripts/python/rtreader.py:178:76: F821 undefined name 'x'
            raise RasterError("Falied to query %dx%d of band %d is none" %(x, y, band))
                                                                           ^
./raster/scripts/python/rtreader.py:178:79: F821 undefined name 'y'
            raise RasterError("Falied to query %dx%d of band %d is none" %(x, y, band))
                                                                              ^
./raster/scripts/python/rtreader.py:178:82: F821 undefined name 'band'
            raise RasterError("Falied to query %dx%d of band %d is none" %(x, y, band))
                                                                                 ^
./raster/scripts/python/raster2pgsql.py:673:12: F632 use ==/!= to compare str, bytes, and int literals
    assert len(ulp) == 2 is not None, "Error: invalid upper-left corner"
           ^
1     F632 use ==/!= to compare str, bytes, and int literals
3     F821 undefined name 'x'
4
The command "flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics" exited with 1.
```

cclauss:
The last one is especially a brain twister because __len(x) == 2__ is either __True__ or __False__ but never __None__.

Or is it?

Komzpa:
No idea. What I would try in such cases is to add more parentheses just in case some parser went nuts and tries to do `2 is not None`.

Komzpa:
```
1.08s$ flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics
./raster/scripts/python/rtreader.py:178:76: F821 undefined name 'x'
            raise RasterError("Falied to query %dx%d of band %d is none" %(x, y, band))
                                                                           ^
./raster/scripts/python/rtreader.py:178:79: F821 undefined name 'y'
            raise RasterError("Falied to query %dx%d of band %d is none" %(x, y, band))
                                                                              ^
./raster/scripts/python/rtreader.py:178:82: F821 undefined name 'band'
            raise RasterError("Falied to query %dx%d of band %d is none" %(x, y, band))
                                                                                 ^
3     F821 undefined name 'x'
3
The command "flake8 . --count --select=E9,F63,F72,F82 --show-source --statistics" exited with 1.
```

Komzpa:
Thanks! merged.

PR #420, opened 2019-06-13
Throw on invalid characters when decoding geohash

Algunenano:
Fixes https://trac.osgeo.org/postgis/ticket/4406

### DIFF
--- a/liblwgeom/cunit/cu_algorithm.c
+++ b/liblwgeom/cunit/cu_algorithm.c
@@ -1138,6 +1138,23 @@ static void test_geohash_point_as_int(void)
 	CU_ASSERT_EQUAL(gh, rs);
 }
 
+static void
+test_geohash_bbox(void)
+{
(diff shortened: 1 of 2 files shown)


Algunenano:
> I think rather than duplicate knowledge of valid base32 characters, it is much cleaner to just check the result of strchr:
char* p = strchr(base32, c);
if (!p) { lwerror(...); return; }
cd = p - base32;

Makes sense. It's slower but, as you say, much cleaner. I'll change it.

mentin:
Should not be slower - since the function is calling strchr() anyway. 
Just before using its result it should check whether strchr() returned NULL.
So there should still be only one strchr() call per character, as in original,
and just a single new condition checking whether it returned NULL.

Algunenano:
Changed in https://github.com/postgis/postgis/commit/221bbd932a5ff2ead9d8fbbfeb2bb10c1ffdc9a8

PR #423, opened 2019-06-18
Proj6 impr cache

Algunenano:
Replaces  https://github.com/postgis/postgis/pull/422
Related to https://trac.osgeo.org/postgis/ticket/4372

What this does:
- Cleans up configure.ac when running under PROJ6 (avoids error message that its ignored).
- PROJ 6+ (and only 6+):
  - Caches source_swapped / target_swapped for all PJs. This brings ST_Transform to, more or less, almost the same performance as 5.1 (except the first row, which creates the PJ).
  - Caches source_is_latlong / source_semi_major_metre / source_semi_minor_metre when required (source srid == target srid). This brings a bunch of functions (ST_AsTWKB, geography functions) to, more or less, the same performance as before (except the first row, which creates the PJ).
  - Uses `proj_trans` instead of `proj_trans_generic` for single points (faster and makes it on par of 5.2 performance for that case).
- PROJ (all releases):
  - Allocate auxiliar memory (LWPROJ / PJ) in the parent memory context instead of manually using malloc/free.
  - Avoid double lookup (one to check if the item is in the cache, one for retrieval).
  - Increases PROJ_CACHE_ITEMS from **16** to **128**. PROJ_BACKEND_HASH_SIZE goes from **32** to **256**. This is to minimize the huge impact of creating projections between 5.2 and 6.1, specially for ST_Buffer(geography) which autogenerates srids based on `SRID_RESERVE_OFFSET`. There are over 120+ used SRIDs in that range, so deleting and recreating PJs is a huge performance loss (it isn't as bad for 5.1 since PJ creation is much faster).
  - Removes the call to `proj_context_set_search_paths`. AFAIK, we don't distribute proj files anymore so this is legacy. It's also preventing the default proj installation of finding proj.db and using the fallback method (faster but more imprecise). If it breaks something, I'll consider reading it just for Windows (as it's mentioned in the comments of the function).

@pramsey Any thoughts about this?

### DIFF
--- a/libpgcommon/lwgeom_transform.c
+++ b/libpgcommon/lwgeom_transform.c
@@ -77,7 +77,7 @@ typedef struct {
 typedef struct struct_PJHashEntry
 {
 	MemoryContext ProjectionContext;
-	PJ* projection;
+	LWPROJ *projection;
 }
 PJHashEntry;
 
@@ -87,8 +87,8 @@ uint32 mcxt_ptr_hash(const void *key, Size keysize);
 
 static HTAB *CreatePJHash(void);
 static void DeletePJHashEntry(MemoryContext mcxt);
-static PJ* GetPJHashEntry(MemoryContext mcxt);
-static void AddPJHashEntry(MemoryContext mcxt, PJ* projection);
+static LWPROJ *GetPJHashEntry(MemoryContext mcxt);
+static void AddPJHashEntry(MemoryContext mcxt, LWPROJ *projection);
 
 /* Internal Cache API */
 /* static PROJPortalCache *GetPROJSRSCache(FunctionCallInfo fcinfo) ; */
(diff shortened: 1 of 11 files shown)


libpgcommon/lwgeom_transform.c
```diff
@@ -47,8 +47,7 @@ static char *spatialRefSysSchema = NULL;
  * typically have 2 entries per portal
  * then we shall use a default size of 32)
```
pramsey: update comment?

pramsey:
Seems fine to me. Reading through I'm not sure if we still need the whole hash/mctx dance once we get to a 9.6+ PGSQL baseline, but I guess we're not there yet...

Algunenano:
> . Reading through I'm not sure if we still need the whole hash/mctx dance once we get to a 9.6+ PGSQL baseline, but I guess we're not there yet...

Uhmm you are right; for 9.6+ we could probably set up callbacks in the parent memory context and pass the LWPROJ pointer instead of creating the auxiliar memory context and do the whole PJHash thingy. I'm going to give it a try and see if I can clean the PG9.6+ code and leave the old one ready to be cleaned up when we drop 9.5.

Algunenano:
I've added an extra commit to avoid dropping N elements whenever one eviction is needed, and instead drop the less used projection only.

Algunenano:
Everything has been pushed in multiple commits.

PR #424, opened 2019-06-18
Enable bytebuffer tests

Algunenano:
~~While checking low coverage files I noticed that bytebuffer has tests (well, only one), but it's commented out. Let's enable it and see if something breaks.~~

OTOH... The code that's not covered are unused functions. Let's try removing them (git history it's always there).


Algunenano:
Closed in https://github.com/postgis/postgis/commit/a7edcb38dde1a4d1d90a56973f799deceeacd5e5

PR #427, opened 2019-06-27
[raster] Skip all NODATA tiles

Komzpa:
Skip tiles that don't have anything in them. If you still want them in your db, use existing `-k` option to disable the check.


Komzpa:
https://trac.osgeo.org/postgis/ticket/4442#ticket

PR #432, opened 2019-07-05
LWGEOM_x_point / LWGEOM_y_point avoid deserialization if possible

Algunenano:
This is a WIP, it needs more tests and expanding it to the Z and M coordinates


Algunenano:
Re-implementing this in a different manner.

PR #433, opened 2019-07-05
Fix gserialized_max_header_size

Algunenano:
Fix https://trac.osgeo.org/postgis/ticket/4451

Avoid including extra bytes in the count.

### DIFF
--- a/liblwgeom/gserialized1.c
+++ b/liblwgeom/gserialized1.c
@@ -113,8 +113,9 @@ int gserialized1_is_geodetic(const GSERIALIZED *gser)
 
 uint32_t gserialized1_max_header_size(void)
 {
-	/* read GSERIALIZED size + max bbox according gbox_serialized_size (2 + Z + M) + 1 int for type */
-	return sizeof(GSERIALIZED) + 8 * sizeof(float) + sizeof(int);
+	intptr_t size_of_gserialized_up_to_data = (intptr_t) & ((GSERIALIZED *)NULL)->data;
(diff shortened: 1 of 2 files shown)


Komzpa:
`tag=pg11-clang-geos37-gdal24-proj52 mode=usan_clang` is not happy

Algunenano:
> What I don't quite understand is why logbt didn't decipher the core dump - was there one?

I've run it locally and although it's also crashing and calling 
```
systemd-coredump[25806]: Resource limits disable core dumping for process 25804 (postgres).
```

```
$ su postgres -c "ulimit -a -H"
core file size          (blocks, -c) unlimited
data seg size           (kbytes, -d) unlimited
scheduling priority             (-e) 0
file size               (blocks, -f) unlimited
pending signals                 (-i) 128198
max locked memory       (kbytes, -l) 64
max memory size         (kbytes, -m) unlimited
open files                      (-n) 524288
pipe size            (512 bytes, -p) 8
POSIX message queues     (bytes, -q) 819200
real-time priority              (-r) 0
stack size              (kbytes, -s) unlimited
cpu time               (seconds, -t) unlimited
max user processes              (-u) 128198
virtual memory          (kbytes, -v) unlimited
file locks                      (-x) unlimited
```

Other dumps are working fine, so I don't know what's going on here.

In any case, the issue comes from playing with NULL to get the position of ->data, but since that's constant we can enforce calculating it at compile time and avoid runtime errors.

PR #434, opened 2019-07-05
geometry_geometrytype: Use static strings to avoid calculating the output

Algunenano:
Trac: https://trac.osgeo.org/postgis/ticket/4450

I'll add more tests around ST_GeometryType as there doesn't seem to be many of them.


Algunenano:
It seems `ST_GeometryType` is fully covered (multiple times) by `regress/core/typmod_expected` so pushing the change as is.

PR #435, opened 2019-07-05
ST_X/ST_Y/ST_Z/ST_M: Read coordinates without deserializing

Algunenano:
Pending:
- Separate cleanup / improvement changes from the  new function itself.
- Extra tests around ST_X, ST_Y, ST_Z, ST_M since there aren't too many.

Trac: https://trac.osgeo.org/postgis/ticket/4449

### DIFF
--- a/liblwgeom/gserialized2.c
+++ b/liblwgeom/gserialized2.c
@@ -99,6 +99,20 @@ static inline size_t gserialized2_header_size(const GSERIALIZED *g)
 	return sz;
 }
 
+/* Returns a pointer to the start of the geometry data */
+static inline uint8_t *
+gserialized2_get_geometry_p(const GSERIALIZED *g)
+{
+	uint32_t extra_data_bytes = 0;
+	if (gserialized2_has_extended(g))
+		extra_data_bytes += sizeof(uint64_t);
+
+	if (gserialized2_has_bbox(g))
+		extra_data_bytes += gserialized2_box_size(g);
+
+	return ((uint8_t *)g->data) + extra_data_bytes;
+}
+
 uint8_t lwflags_get_g2flags(lwflags_t lwflags)
 {
 	uint8_t gflags = 0;
@@ -168,8 +182,7 @@ uint32_t gserialized2_max_header_size(void)
 
 uint32_t gserialized2_get_type(const GSERIALIZED *g)
 {
-	uint8_t *ptr = (uint8_t*)g;
-	ptr += gserialized2_header_size(g);
+	uint8_t *ptr = gserialized2_get_geometry_p(g);
 	return *((uint32_t*)(ptr));
 }
 
@@ -237,9 +250,7 @@ static size_t gserialized2_is_empty_recurse(const uint8_t *p, int *isempty)
 int gserialized2_is_empty(const GSERIALIZED *g)
 {
 	int isempty = 0;
-	uint8_t *p = (uint8_t*)g;
-	assert(g);
-	p += gserialized2_header_size(g);
+	uint8_t *p = gserialized2_get_geometry_p(g);
 	gserialized2_is_empty_recurse(p, &isempty);
 	return isempty;
 }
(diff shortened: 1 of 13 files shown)


liblwgeom/gserialized2.c
```diff
@@ -519,6 +526,36 @@ gserialized2_peek_gbox_p(const GSERIALIZED *g, GBOX *gbox)
 	return LW_FAILURE;
 }
 
+int
+gserialized2_peek_first_point(const GSERIALIZED *g, POINT4D *out_point)
```
Komzpa: This is a super useful function!

liblwgeom/gserialized2.c
```diff
@@ -519,6 +526,36 @@ gserialized2_peek_gbox_p(const GSERIALIZED *g, GBOX *gbox)
...
+{
+	uint8_t *data_ptr = gserialized2_get_geometry_p(g);
+
+	uint32_t type = (((uint32_t *)data_ptr)[0]);
+	int isEmpty = (((uint32_t *)data_ptr)[1]) == 0;
+	if (type != POINTTYPE || isEmpty)
```
Komzpa: this does not match function name. what if input is something else?

postgis/lwgeom_ogc.c
```diff
@@ -684,27 +652,17 @@ Datum LWGEOM_z_point(PG_FUNCTION_ARGS)
...
 
-	if ( gserialized_get_type(geom) != POINTTYPE )
+	if (gserialized_get_type(geom) != POINTTYPE)
 		lwpgerror("Argument to ST_M() must be a point");
 
-	lwgeom = lwgeom_from_gserialized(geom);
-	point = lwgeom_as_lwpoint(lwgeom);
-
-	if (lwgeom_is_empty(lwgeom) || !lwgeom_has_m(lwgeom))
+	if (!gserialized_has_m(geom) || (gserialized_peek_first_point(geom, &pt) == LW_FAILURE))
```
Komzpa: is compiler clever enough to inline it and skip unpacking of everything except M?

liblwgeom/gserialized2.c
Algunenano: Do you think `gserialized$_point_peek_first` is better?

postgis/lwgeom_ogc.c
Algunenano: I don't think so, it is going to bring the 4 coordinates if available. Initially I was going to write 4 different functions instead, but the minimal performance gain wasn't worth duplicating the code (4 coordinates x 2 serialization methods).

postgis/lwgeom_ogc.c
```diff
@@ -591,26 +591,17 @@ Datum LWGEOM_pointn_linestring(PG_FUNCTION_ARGS)
...
-	LWGEOM *lwgeom;
-	LWPOINT *point = NULL;
-	POINT2D p;
-
-	geom = PG_GETARG_GSERIALIZED_P(0);
+	GSERIALIZED *geom = PG_GETARG_GSERIALIZED_P(0);
+	POINT4D pt;
 
-	if ( gserialized_get_type(geom) != POINTTYPE )
+	if (gserialized_get_type(geom) != POINTTYPE)
```
Algunenano: Note to self: Maybe this can be moved into the next if condition, and only check it if `gserialized_peek_first_point` fails.

Komzpa: what if someone updates gserialized_peek_first_point to peek first point of anything? (I may, it would be a better sorting key than center of bbox)

liblwgeom/gserialized2.c
Komzpa: "first" means it looks at first if there are many. This one just breaks if there's more than one.
For #436 may be cool to calculate hash on first point instead of going into loops of expanding it into a box and then taking its center, so a function that takes first point of whatever it gets will be useful. It can also speed up ST_StartPoint.

postgis/lwgeom_ogc.c
Komzpa: do we get .bc files for liblwgeom so that postgres's JIT can inline them in runtime?

Algunenano: > do we get .bc files for liblwgeom so that postgres's JIT can inline them in runtime?

AFAIK, no since `liblwgeom` isn't compiled using postgres machinery.

liblwgeom/gserialized2.c
Algunenano: Getting the point of a POINTTYPE is "trivial", but getting the first point of any geometry isn't (since it might involve recursion).  I think I might start throwing an error if something other than a POINT is passed and add other geometry types when needed, but I have to think this through.

Algunenano:
@Komzpa I've changed gserializedX_peek_first_point to make it easier to add other geometry implementations in the near future.

My current plan is:
- Add more tests around the ST_{X,Y,Z,M} functions.
- Push these changes into trunk (probably a couple of commits, one with improvements and one with the changes).
- Study how hard / how much benefit we get from implementing other geometry types to improve other functions (like `ST_StartPoint`).

Does it make sense to you?

PR #438, opened 2019-07-07
spatial_ref_sys.sql: refresh with PROJ 6 and add Geographic3D CRS and ESRI definitions

rouault:
- I've tried to keep the structure as close as possible as previous file
- When possible, WKT1 is used
- For Geographic 3D CRS, WKT2 is used since WKT1 is not possible
- TOWGS84 clauses might have disappear when there are several datum transformations
  possible. This is not a problem when using the modern proj_create_crs_to_crs() with
  the EPSG/ESRI codes.

Was generated with `python swig/python/scripts/epsg_tr.py -postgis > out.sql` from latest GDAL master

### DIFF
--- a/regress/core/regress_proj.sql
+++ b/regress/core/regress_proj.sql
@@ -65,5 +65,5 @@ SELECT 13, count(*) FROM
 (
     SELECT ST_Transform('SRID=4326; POINT(0 0)'::geometry, srid) AS g
     FROM
-        ( SELECT srid FROM spatial_ref_sys LIMIT 150 ) _a
+        ( SELECT srid FROM spatial_ref_sys WHERE srtext LIKE 'GEOGCS%' AND proj4text NOT LIKE '%clrk80ign%' LIMIT 150 ) _a
 ) _b WHERE g IS NOT NULL;


pramsey:
> TOWGS84 clauses might have disappear when there are several datum transformations
possible. This is not a problem when using the modern proj_create_crs_to_crs() with
the EPSG/ESRI codes.

This is interesting... on the one hand it means that PostGIS 3 + Proj 6 wouldn't have any issues, but it also means that PostGIS 3 + Proj 5 would have different (wrong) behaviour. It seems like a non-small issue to loose TOWGS84 for those cases, no?

rouault:
Simple solution: require PROJ 6 :-)

>  It seems like a non-small issue to loose TOWGS84 for those cases, no?

It is lost in the cases where previously the script that imported from the EPSG database made a more or less arbitrary choice in selecting the transformation that had the largest area of use. So for example for Pulkovo(42), this was Poland (or Romania I never remember), which means that the people from the other country (and Albania) where screwed up. So now everybody is screwed up if using this PROJ string without towgs84 :-)

Komzpa:
> Simple solution: require PROJ 6 :-)

Let's go and do it?
https://lists.osgeo.org/pipermail/postgis-devel/2019-July/028009.html

pramsey:
Is we end up at proj 4.9 as our Debian baseline for PostGIS 3 (per the email thread above) that still leaves regressions for those distros. Is there a way to preserve the old TOWGS84 definitions? The idea that having no transform is somehow better than having a less-than-perfect one chosen for you, particularly since the latter is what has already been happening for 15 years, seems misguided to me.

rouault:
> . Is there a way to preserve the old TOWGS84 definitions?

The old logic is no longer in PROJ (actually was never in it, but rather in libgeotiff, but whatever), so not easily. Well, you could explore the different candidate coordinate operations returned by proj_create_operations() and pick up the one with the largest area of use, but I'm not super interested in investing time in this.

Komzpa:
Can we break loudly on these? "Please upgrade Proj to 6 to be able to do this transform correctly"?

pramsey:
Any reason for `-postgis` the IGNF authority is not generated? @rouault ?

rouault:
> Any reason for `-postgis` the IGNF authority is not generated?

This is due to `auth_srid  INTEGER` but the IGNF codes are alphanumeric.

rouault:
Just thinking that one difference I didn't mention w.r.t the previous version is that  the AXIS order in the WKT is the one mandated by the authority, so lat,long for EPSG geographic CRS (and northing, easting for a few projected CRS). This shouldn't impact coordinate transformations as @pramsey actively coded to use the usual GIS friendly axis order even if the CRS definition wants another one, but it would perhaps be worth documenting how axis order is handled in PostGIS if not already done.

pramsey:
@Komzpa I don't think I want this patch as-is, I'd rather just diff to find the net-new EPSG numbers (including the one you wanted) and add those into our current file.

rouault:
> I'd rather just diff to find the net-new EPSG numbers (including the one you wanted) and add those into our current file.

This could be a bit confusing. The patch also includes a refresh from recent EPSG dataset that makes that a number of towgs84 values are refreshed and consistent with proj.db of PROJ 6. So when using the PROJ 6 interface, PostGIS (as far as I understood @pramsey work) will rely on the new values from proj.db, not whatever is in spatial_ref_sys.sql. So, if keeping older values in spatial_ref_sys.sql, users could be confused about why they get slightly different results if they use the SRID or the PROJ string.

In an ideal world with PROJ 6 being required, PostGIS 3 would have an empty proj4text column for all the known SRID from proj.db

One thing I didn't check if if in PROJ 6 mode, PostGIS uses well (auth_name, auth_srid) (or the WKT) to instanciate the PROJ CRS object (and not hard-code ('EPSG:', srid) )

pramsey:
No, it only looks at 'EPSG' auth_name values, so any 'ESRI' values wouldn't be directly looked up, they'd be instantiated via WKT. Also it looks like it might be brittle for NULL values in auth_srid. All in all a good code line to re-examine.

pramsey:
I'm going to saw the baby in half.
I have extracted the net-new entries to add to the existing spatial_ref_sys entries.
Maybe when proj6 is more established we can afford to lose the more detailed proj4text entries.
For people who *do* have proj6, I have ensured that both EPSG and ESRI authority names will get used to instantiate transforms, so it should be a "best case" for them. For the folks on older proj, they'll get the benefit of the new entries and not be any worse off.

pramsey:
Closing this in favour of the half-a-baby in #441

PR #439, opened 2019-07-08
ST_TileEnvelope

pramsey:
Given a tile zoom, y and x, generate an envelope in Web Mercator that can be transformed, etc, and used to build things like MVT tiles. https://trac.osgeo.org/postgis/ticket/4452


postgis/lwgeom_functions_basic.c
```diff
@@ -2064,6 +2064,69 @@ Datum ST_MakeEnvelope(PG_FUNCTION_ARGS)
...
+
+	if (zoom >= 32 || zoom < 0)
+		elog(ERROR, "%s: Invalid tile zoom value, %d", __func__, zoom);
+
+	worldGeoSize = Max(boundsWidth, boundsHeight);
+	worldTileSize = 0x01 << (zoom > 31 ? 31 : zoom);
+
+	if (x >= worldTileSize || x < 0)
+		elog(ERROR, "%s: Invalid tile x value, %d", __func__, x);
+	if (y >= worldTileSize || y < 0)
```
Komzpa: tiling continues to the top and to the bottom, just people rarely render there.

doc/reference_constructor.xml
```diff
@@ -756,4 +756,51 @@ SRID=4326;POLYGON((75 29 1, 77 29 2, 77 29 3, 75 29 1))
 	  </refsection>
 	</refentry>
 
+
+	<refentry id="ST_TileEnvelope">
+		<refnamediv>
+		<refname>ST_TileEnvelope</refname>
+
+<ulink url="https://en.wikipedia.org/wiki/Web_Mercator_projection">Web Mercator</ulink>
```
Komzpa: stray link?

postgis/lwgeom_functions_basic.c
```diff
@@ -2064,6 +2064,69 @@ Datum ST_MakeEnvelope(PG_FUNCTION_ARGS)
...
+	boundsWidth  = bounds->xmax - bounds->xmin;
+	boundsHeight = bounds->ymax - bounds->ymin;
+	if (boundsWidth <= 0 || boundsHeight <= 0)
+		elog(ERROR, "%s: Geometric bounds are too small", __func__);
+
+	if (zoom >= 32 || zoom < 0)
+		elog(ERROR, "%s: Invalid tile zoom value, %d", __func__, zoom);
+
+	worldGeoSize = Max(boundsWidth, boundsHeight);
+	worldTileSize = 0x01 << (zoom > 31 ? 31 : zoom);
```
rouault: 1 << 31 is undefined behaviour for a int32

postgis/lwgeom_functions_basic.c
pramsey: You mean into the polar areas where the stretch gets all crazy? ... then we flirt with overflow...

postgis/lwgeom_functions_basic.c
```diff
@@ -2064,6 +2064,69 @@ Datum ST_MakeEnvelope(PG_FUNCTION_ARGS)
...
+	x = PG_GETARG_INT32(1);
+	y = PG_GETARG_INT32(2);
+	if (PG_NARGS() > 3)
+		srid = PG_GETARG_INT32(3);
+	if (PG_NARGS() > 4)
+		bounds = (GBOX*)(PG_GETARG_POINTER(4));
+	else
+	{
+		gbox_init(&bounds_3857);
+		bounds_3857.ymin = bounds_3857.xmin = -20037508.3427892;
```
Komzpa: do we have any kind of get_box_by_srid function? it's a thing I'd need for #436 too.

postgis/lwgeom_functions_basic.c
rouault: and there's redundancy in the zoom > 31 test with the zoom >= 32 that is just above

postgis/lwgeom_functions_basic.c
Komzpa: yes. it's not invalid, just uncommon :)

postgis/lwgeom_functions_basic.c
```diff
@@ -2064,6 +2064,69 @@ Datum ST_MakeEnvelope(PG_FUNCTION_ARGS)
...
+		bounds_3857.ymax = bounds_3857.xmax =  20037508.3427892;
+		bounds = &bounds_3857;
+	}
+
+	boundsWidth  = bounds->xmax - bounds->xmin;
+	boundsHeight = bounds->ymax - bounds->ymin;
+	if (boundsWidth <= 0 || boundsHeight <= 0)
```
Komzpa: can skip check for Mercator

postgis/lwgeom_functions_basic.c
rouault: Testing a bit with -fsanitize=undefined, you need to have worldTileSize defined as uint32_t *and* qualify  the 0x01 constant with U, so that everything operates on uint32

postgis/lwgeom_functions_basic.c
```diff
@@ -2064,6 +2064,69 @@ Datum ST_MakeEnvelope(PG_FUNCTION_ARGS)
...
+		elog(ERROR, "%s: Geometric bounds are too small", __func__);
+
+	if (zoom >= 32 || zoom < 0)
+		elog(ERROR, "%s: Invalid tile zoom value, %d", __func__, zoom);
+
+	worldGeoSize = Max(boundsWidth, boundsHeight);
```
Komzpa: This is wrong for non-square projections. X is tiled along width, Y is tiled along height.

I've seen two ways to tile 4326: half-tiles (512x256, but scheme is counting from 360x180 being (0,0,0)) and NASA JPL Landsat way, where z0 has two tiles and an empty space, effectively with fancier z0 box and zoom offset: https://github.com/Komzpa/twms/blob/master/twms/projections.py#L38

It's possible to implement both if you untie X and Y tiling.

postgis/lwgeom_functions_basic.c
```diff
@@ -2064,6 +2064,69 @@ Datum ST_MakeEnvelope(PG_FUNCTION_ARGS)
...
+		bounds_3857.ymax = bounds_3857.xmax =  20037508.3427892;
+		bounds = &bounds_3857;
+	}
+
+	boundsWidth  = bounds->xmax - bounds->xmin;
+	boundsHeight = bounds->ymax - bounds->ymin;
+	if (boundsWidth <= 0 || boundsHeight <= 0)
+		elog(ERROR, "%s: Geometric bounds are too small", __func__);
+
+	if (zoom >= 32 || zoom < 0)
```
Algunenano: I'm curious, where does max 32 zoom come from? is it something official or just being enough to pinpoint a grain of salt or something  like this?

Algunenano:
Apart from all the coding changes mentioned (and some required by the CI) this needs tests so I don't break it in the future :smile:

postgis/lwgeom_functions_basic.c
Komzpa: @roualt is there any concept of "projection boundaries" so we can ask these numbers from Proj? I see these values on some websites but don't know how are these collected.

Algunenano: It's part of the projection definition. It comes from the Earth equatorial radius (defined by WGS-84 
 as `6378137`) times 2*PI to get the circumference. But I don't know if we can get it "dinamically", but I'd try to avoid it as creating projections is expensive with PROJ6+

rouault: > is there any concept of "projection boundaries" so we can ask these numbers from Proj? I see these values on some websites but don't know how are these collected.

yes, proj_get_area_of_use() exists. It returns values in lat / long, and in the case of EPSG:3857, they use latitude = 85.06 deg, which translates to northing = 20048966.1040146. So those aren't appropriate for the well-defined GoogleMaps tiling scheme.

postgis/lwgeom_functions_basic.c
pramsey: I'm loath to unconstrain it, as it let's people generate unrepresentable tiles (things we cannot actually do the reprojection for). I also like the basic simplicity of understanding a tile scheme as a set of tiles numbered between 0 and (2**zoom - 1) as the valid area of work.

postgis/lwgeom_functions_basic.c
pramsey: Unfortunately the projection area of validity != area of tiling, even for a "simple" case like web mercator, let alone an arbitrary CRS. I feel like this utility really is constrained to handling the web mercator case.

postgis/lwgeom_functions_basic.c
pramsey: Makes sense on unsigned. I'm OK with the redundancy.

postgis/lwgeom_functions_basic.c
pramsey: Yeah, I don't agree...

postgis/lwgeom_functions_basic.c
pramsey: uint32 for holding zoom, primarily. Also, yeah, you don't need any more for a geographic mapping scheme.  Zoom 32 tile is ~ 1cm on a side, so w/ 256x256 resolution, implies a sub-millimetre pixel (0.000036m)

postgis/lwgeom_functions_basic.c
pramsey: I did the above, so x/y subdivide independently.

strk:
I'm not sure I like the name.
Are you aware of https://github.com/CartoDB/cartodb-postgresql/blob/master/scripts-available/CDB_XYZ.sql#L12-L18 ?

The idea at that time was that "XYZ" was an unambiguous enough name for that way to reference webmercator tiles. I find "Tile" too generic for this...

pramsey:
Yes, am aware...
* `ST_TileXYZEnvelope()`
* `ST_TileTMSEnvelope()`
* `ST_TileBingEnvelope()`
?? I'm actually fine with the generic name because I feel like alternate tiling schemes are disappearing and XYZ is ascendant.

Komzpa:
"Tile" is well known.

I'm using similar plpgsql function under name of TileBBox, https://github.com/mapbox/postgis-vt-util/blob/master/src/TileBBox.sql. I've never heard of ST_XYZ, probably because it cannot be read as words and thus isn't googleable. XYZ also looks unpleasant to me as it's "ХУЙ", russian for "dick".

Our naming uses word "Envelope" for boxes expressed as polygons, so this name fits for this function in PostGIS land.

strk:
> * `ST_TileXYZEnvelope()`
> * `ST_TileTMSEnvelope()`
> * `ST_TileBingEnvelope()`

Yes, I like those better

> ?? I'm actually fine with the generic name because I feel like alternate tiling schemes are disappearing and XYZ is ascendant.

So we'll wait for Penguins to estinguish instead
of allowing their mapping ?

Algunenano:
I like the `ST_TileXYZEnvelope` name too, specially since other places use ZXY and it's easy to forget which order each function uses.

A thing I've just now realized is that this function "looks like" it enables you to create a tile from any projection, but not really, since you don't have an easy way to get projection boundaries. Should we have something like `ST_SRIDBox(int srid) -> Box2D` so we can feed this new function?
I'm not 100% sold on the need of it, but I'm certain that if we provide the ability to use different SRIDs in `ST_TileEnvelope` then we should also make it easier to retrieve the bbox parameter from `spatial_ref_sys` for any compatible CRS (GEOGCS?)

pramsey:
First, WRT to XYZ/ZXY, there's a wee issue in the wild, which is that people *talk* about "XYZ" tiles, but then they make URL schemes like http://host/{z}/{x}/{y}.pbf. That is, ZXY. And, our function signature (now) follows the URL ordering, ST_TileEnvelope(z, x, y). So, potentially we'd be looking at ST_TileXYZEnvelope(z, x, y) as the signature. Super, no? And yet... would I have it any other way?

pramsey:
The "return 3857" but "allow other things" scheme, IMO, provides the best case behaviour for real world use, which is that 99% of users will blindly plug in tile coordinates from web mercator schemes and expect sensible outputs by default. Having SRID/Bounds there optionally allows folks who are doing other weird stuff to parameterize exactly as they wish. There are very few other schemes out there that we might support automatically, and I am loath to add a lot of extra mechanism at this current moment. That said, there's always the possibility of eventually using the srid number to magically generate bounds for projections that have known bounds. I just don't think there will be any, and I think other schemes are effectively dead. I was going to do a ST_TileEnvelope(text) variant that accepted bing format, but... why?

Algunenano:
> `ST_TileXYZEnvelope(z, x, y) as the signature. Super, no?`

Please no, either `ST_TileXYZEnvelope(x ,y, z)` or `ST_TileZXYEnvelope(z, x, y)` but anything in between is going to be extremely confusing.

> Having SRID/Bounds there optionally allows folks who are doing other weird stuff to parameterize exactly as they wish

In that case, wouldn't it be better to join those 2 parameters into one and accept a box with SRID?


In a different topic, I'b remove any 3857 constants from the C code and rely entirely in the function defaults. Can you even receive less than the whole 5 parameters in that case? Aren't parameters declared optional always passed?

pramsey:
"box with an SRID" doesn't seem like something we can instantiate from userland... there's no box literal with an embedded SRID, and calls to ST_SetSRID just end up implicitly casting the box inputs to geometry, so ST_TileEnvelope(z, x, y, geometry bounds) seems like the signature, where the box of the bounds and srid of the bounds are used to define the zoom zero tiling space.

Algunenano:
> so ST_TileEnvelope(z, x, y, geometry bounds) seems like the signature, where the box of the bounds and srid of the bounds are used to define the zoom zero tiling space

That looks fine to me and it's more concise that before.

strk:
> Please no, either `ST_TileXYZEnvelope(x ,y, z)` or `ST_TileZXYEnvelope(z, x, y)` but anything in between is going to be extremely confusing.

Agreed. I'd go `ST_TileXYZEnvelope(x ,y, z)`

strk:
See also http://postgis.net/docs/ST_BoundingDiagonal.html, for generating the last parameter (for the docs)

dr-jts:
I like the simplicity of `ST_TileEnvelope`.  As Paul says, XYZ is the dominant tiling scheme now and for the foreseeable future, so being more verbose for the sake of unknown future requirements seems overly cautious.

cf `ST_Simplify` and `ST_SimplifyVW` - first one in gets the nice simple name!

pramsey:
Merged

vitaligi:
Would like it to support zero level zoom with arbitrary number of tiles in x & y. It could help support 2x1 tiles (level 0) tiling schemes (like in annex d.2 in http://docs.opengeospatial.org/is/17-083r2/17-083r2.html)

PR #440, opened 2019-07-09
Address 32bit warning

Algunenano:
geobuf.c:93:37: warning: absolute value function 'labs' given an argument of type 'int64' (aka 'long long') but has parameter of type 'long' which may cause truncation of value [-Wabsolute-value]
                value->neg_int_value = (uint64_t) labs(intval);
                                                  ^
geobuf.c:93:37: note: use function 'llabs' instead
                value->neg_int_value = (uint64_t) labs(intval);
                                                  ^~~~
                                                  llabs


Algunenano:
Pushed

PR #441, opened 2019-07-11
Generic auth_name support and updated spatial_ref_sys

pramsey:
Respect auth_name for all values of auth_name, add net new lines to spatial_ref_sys table.


pramsey:
Merged at  4d4891e

PR #444, opened 2019-07-12
[WIP] Avoid deserialization in ST_OrderingEquals

Algunenano:
Pending: Re-adding the type and ZM comparison since gserialized_cmp doesn't handle that.


Komzpa:
If we're not using ZM flag in gserialized_cmp what if I add compare for it it to #436?

Komzpa:
I've added ZM to #436.

Komzpa:
close/open to rebuild as rebased

PR #445, opened 2019-07-13
Bump Proj to 4.9 and GDAL to 2.1

Komzpa:


robe2:
I'm okay with this as long as we update the drone docker images first.

robe2:
I decided I'm not really okay with this.  In 3.1 yes.  In 3.0 no cause we already called alpha and I hate changing requirements when we've already called alpha.  Also I don't feel it serves any purpose.  It's not going to sway packagers except maybe from shipping 3.0 and it's going to make people how build their own postgis if they are stuck in a crappy OS -- a lot harder.  

I don't believe we have junk code we can remove by raising the GDAL bar and as for Proj -- maybe.  But I still think we should wait a year.

PR #446, opened 2019-07-16
Add CPPFLAGS where missing

df7cb:
Spotted by Debian's build log hardening checker, blhc.

https://salsa.debian.org/debian-gis-team/postgis/-/jobs/222378


df7cb:
Fwiw, the next test run with that patch confirmed that the issues are fixed: https://salsa.debian.org/debian-gis-team/postgis/-/jobs/222420

Algunenano:
Thanks.

LGTM. Pushing and backporting it to all stable branches.

PR #449, opened 2019-07-19
postgis_extensions_upgrade: Also update on pgupgrade

Algunenano:
Fix for https://trac.osgeo.org/postgis/ticket/4271

I need to test it with pgupgrade to confirm it works


strk:
Sounds good to me. Not sure how pgupgrade matters here

PR #452, opened 2019-08-02
[WIP] Ensure arc linearize has the same start and end as the passed parameters

Algunenano:
Fixes https://trac.osgeo.org/postgis/ticket/4407

I need to run some extra tests


Algunenano:
Closed in favor of https://github.com/postgis/postgis/pull/458

PR #453, opened 2019-08-03
Rewrite ST_GeomFromGeoJSON

Komzpa:
https://trac.osgeo.org/postgis/ticket/4470


pramsey:
This function was a source of a lot of crashers, but I think they are all in regress, so as long as you're still passing I guess I'm OK. It's not clear to me where the bulk of changes are, because there's a lot of reformatting.

PR #458, opened 2019-08-06
Linearize: Iterate over number of segments

Algunenano:
Better alternative to https://github.com/postgis/postgis/pull/452

Trac: https://trac.osgeo.org/postgis/ticket/4407#comment:18

### DIFF
--- a/liblwgeom/lwstroke.c
+++ b/liblwgeom/lwstroke.c
@@ -544,7 +544,7 @@ lwcompound_linearize(const LWCOMPOUND *icompound, double tol,
                       int flags)
 {
 	LWGEOM *geom;
-	POINTARRAY *ptarray = NULL, *ptarray_out = NULL;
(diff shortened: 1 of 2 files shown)


liblwgeom/lwstroke.c
```diff
@@ -409,9 +423,16 @@ lwarc_linearize(POINTARRAY *to,
 	if( is_circle )
 	{
 		increment = fabs(increment);
+		segments = ceil(total_angle / increment);
```
dbaston: Is this kind of repeating the work done around line 342?

Algunenano: Yes; since it might modify the value of increment I added this again here to make sure there is at least 3 segments for circles.

Algunenano:
I merged this in 8442b3c5f42a7cfe5e5bb08741c1c6aeb6c9126d; not sure why it didn't close the PR automagically.

PR #460, opened 2019-08-13
Re-organize protobuf tests

pramsey:
... and add lib/inc specific configuration options. @Algunenano your input, as you touched this area last? I cannot build against stock libprotobuf with things as they stand, see https://trac.osgeo.org/postgis/ticket/4481


pramsey:
Committed

PR #462, opened 2019-08-17
GeoJSON OSS-Fuzz fuzzer

Komzpa:
Includes fix for leaks found in first 20 minutes.


pramsey:
Seems OK reading through. I like the idea of fuzzers but have found tracking down the causes of failures immensely frustrating and hard. If you have a recipe running from "google ticket" to "here are the actual inputs that caused the issue" I would love to see it written down.

Komzpa:
to run fuzzer do something like this
```
git clone https://github.com/google/oss-fuzz
cd oss-fuzz
python infra/helper.py build_image postgis
python infra/helper.py build_fuzzers --sanitizer address postgis /home/kom/proj/postgis/ 
python infra/helper.py run_fuzzer postgis geojson_import_fuzzer
... wait 2 hours ...
==17==ERROR: LeakSanitizer: detected memory leaks
...stacktrace...
artifact_prefix='./'; Test unit written to ./leak-7c9eeff6edcee3fc53f8be8e1981b7538ea90729

cat build/out/postgis/leak-7c9eeff6edcee3fc53f8be8e1981b7538ea90729
*feed into input function as with usual debug process*
```

fuzzers/geojson_import_fuzzer.cpp
```diff
@@ -0,0 +1,75 @@
+/******************************************************************************
+ *
+ * Project:  PostGIS
+ * Purpose:  Fuzzer
+ * Author:   Even Rouault, even.rouault at spatialys.com
```
Algunenano: Late to the party, but I guess this should be 2019 + your name, or the generic Postgis header right?

Algunenano: Same for the other files

PR #463, opened 2019-08-18
Zero tolerance simplification in MVT export

Komzpa:


pramsey:
Definitely do not commit until @Algunenano has had a look. He's put a lot of time into that code line.

Komzpa:
Sure. There's a non-wagyu test failing, area of some polygon becomes out of bounds - that one needs care anyway.

Algunenano:
I'm out on holidays, i won't be able to check this for another week at least. I remember tinkering with disabling this completely and there should be a ticket about that.

Algunenano:
I'm running some performance benchmarks to compare:
- Current Trunk.
- This change (zero tolerance simplification after snapping to grid).
- Removing simplification completely.

I'll let you know as soon as I'm done

Komzpa:
One thing is that zero tolerance simplification can be further optimized into O(N) algorithm, helping this PR and ST_Subdivide (where it's used to clean up extra points along vertical/horizontal lines).

Algunenano:
One thing I've just noticed is that ST_AsMVTGeom in trunk has COST 1 (the default) and we might want to raise that using the cost macros that @pramsey introduced so raise parallelism. In any case, I'll finish running the benchmarks as they are right now first to compare under the same conditions.

Algunenano:
Comparing current trunk vs trunk + this patch:
[cartovl_trunk_20190827_vs_cartovl_zerotol_20190827.pdf](https://github.com/postgis/postgis/files/3546173/cartovl_trunk_20190827_vs_cartovl_zerotol_20190827.pdf)

There isn't any benchmark where this patch is faster than trunk, but there are several where it's noticeable slower.
The most extreme loss is in the `[MVT] Canada [3,1,1 3 −> 3 pgs]` where it gets 3 big polygons and simplifies them quite a bit:
- 6341 points to 49 points
- 141957 to 17986 points
- 1297243 to 44562 points  

There are other benchmarks with small polygons instead (`NYC buildings`) and those show ~10% decrease in performance, so I'd say this patch would need further work to match trunk performance.

Even though the output isn't the same for some geometries, I've also tested removing the simplification steps completely:
[cartovl_trunk_20190827_vs_cartovl_nosimpl_20190827.pdf](https://github.com/postgis/postgis/files/3546236/cartovl_trunk_20190827_vs_cartovl_nosimpl_20190827.pdf)

Here I can see that certain polygon tiles are slower (`[MVT][HIGH] NYC buildings [9,150,192 802977 −> 53929pgs]` is 1x slower, or `[MVT] Canada [3,1,1 3 −> 3 pgs]` which is 2x slower), but some other tiles are faster (`[MVT] Canada [17,32426,21910 1 −> 1 pgs]` is 3x faster).

I think this is related on how much part of the polygon is discarded in the clipping phase.

Although there are still some gains to be made in ST_AsMVTGeom, I think they are likely to depend on both the complexity of the different phases (simplification, clipping, transformation and validation) and some heuristics based on geometry type and size vs the tile extent.

strk:

Komzpa:
@Algunenano can you have a look at performance with this tolerance=0 optimization?

Algunenano:
I understand I can use just https://github.com/postgis/postgis/pull/463/commits/e731d4c846333597736c18dfb99e7f74a8518727 on top of my simplification fix right?

Komzpa:
https://github.com/postgis/postgis/pull/463/commits/e731d4c846333597736c18dfb99e7f74a8518727 and https://github.com/postgis/postgis/pull/463/commits/e6c3a6353245983f62458ea0b50d8a9c716afb3a

Algunenano:
Comparing:
- Base: https://github.com/Algunenano/postgis/tree/simplify_bug
- Changes: https://github.com/Algunenano/postgis/tree/simplify_bug_zerotol

Result: [20191031_base_vs_20191031_zerotol.pdf](https://github.com/postgis/postgis/files/3793633/20191031_base_vs_20191031_zerotol.pdf)

The following tiles got noticeable slower:
- `[MVT][HIGH] NYC buildings [9,150,192 802977 −> 53929pgs]`
- `[MVT] Canada [3,1,1 3 −> 3 pgs]`

These tiles got noticeable faster:
- `[MVT] Canada [7,31,21         1 -> 1 pgs]`
- `[MVT] Canada [11,506,342      1 -> 1 pgs]`
- `[MVT] Canada [15,8106,5477    1 -> 1 pgs]`
- `[MVT] Canada [17,32426,21910  1 -> 1 pgs]`

It probably isn't a coincidence that the tiles that got slower are also the tiles that get noticeable slower when the simplification step is completely removed.

Algunenano:
The query used for [MVT][HIGH] NYC buildings [9,150,192 802977 −> 53929pgs]:
```sql
SET TRANSACTION READ ONLY; SELECT ((SELECT ST_AsMVT(geometries, 'layer0', 2048)
 FROM (
 SELECT ST_AsMVTGeom(
 the_geom_webmercator,
 ST_MakeEnvelope(-8296780.798186171758, 4931105.5687332907617, -8218509.2812221512697, 5009377.08569731125, 3857),
 2048,
 8,
 true
 ) as the_geom_webmercator, "cartodb_id"
 FROM (SELECT cartodb_id, the_geom_webmercator FROM benchmark_af85af967525e69373b7b8678f3bd409a8e78f447671344b7535d) AS cdbq
 WHERE the_geom_webmercator && ST_Expand(ST_MakeEnvelope(-8296780.798186171758, 4931105.5687332907617, -8218509.2812221512697, 5009377.08569731125, 3857), 305.748113140705)
 ) AS geometries
 )) AS st_asmvt
```
Dataset comes from  "https://data.cityofnewyork.us/api/geospatial/nqwf-w8eh?method=export&format=Original" but it's modified by CARTO's importer, so you might want to drop the cartodb_id and use the correct table name and geometry column name.

The query used for `[MVT] Canada [3,1,1 3 −> 3 pgs]`:
```
SET TRANSACTION READ ONLY; SELECT ((SELECT ST_AsMVT(geometries, 'layer0', 2048)
 FROM (
 SELECT ST_AsMVTGeom(
 the_geom_webmercator,
 ST_MakeEnvelope(-15028131.25709193375, 10018754.1713946225, -10018754.1713946225, 15028131.25709193375, 3857),
 2048,
 8,
 true
 ) as the_geom_webmercator, "cartodb_id"
 FROM (SELECT cartodb_id, the_geom_webmercator FROM benchmark_4c7214d90a79aa6760367a084a4d4a2f61fbe1c6cc4f7f9e76020) AS cdbq
 WHERE the_geom_webmercator && ST_Expand(ST_MakeEnvelope(-15028131.25709193375, 10018754.1713946225, -10018754.1713946225, 15028131.25709193375, 3857), 19567.87924100512)
 ) AS geometries
 )) AS st_asmvt
```

Dataset comes from Canada: "http://www12.statcan.gc.ca/census-recensement/2011/geo/bound-limit/files-fichiers/2016/lpr_000b16a_e.zip"
Similar changes required to the adapt the query as the one before (changing table name and geometry column name).

Keep in mind that this is using 2048 extend and just 8 extra buffer instead of the default (4096, 256), which seems excessive for most of our use cases.

Algunenano:
Performance comparison against current status of this branch (no rebase done):
[20191031_base_vs_20191105_ticket_4149_13135.pdf](https://github.com/postgis/postgis/files/3808311/20191031_base_vs_20191105_ticket_4149_13135.pdf)

At a quick glance this last changes haven't made any impact in those 2 slow tests.

Komzpa:
What is we do it differently?

Let ST_AsMVTGeom not do the pixel-size simplification, but instead advise to call it with ST_Simplify(geom,0.5px) in advance. That would also let people call it with Volkswagen's ST_SimplifyVW which is more suited at polygons than Douglas-Peucker. Only issue seems to be to calculate the pixel size at zoom - formula is there but painful.

So ST_PixelSize(zoom, extent=4096, bounds="as_in_tileenvelope"), usage ST_Expand(ST_TileEnvelope(2,1,1), 7*ST_PixelSize(2)), read as "expand tile 1,1 on zoom 2 by 7 pixels of zoom 2" and ST_Simplify(geom, 0.5*ST_PixelSize(2)), read as "simplify geometry to halfpixel at zoom 2".

This also touches #514 which also struggles with need to specify a pixel based expand. 

In days I did rendering I had :pixel_size_at_zoom_6 (1..19) macro.

Mapnik has !pixel_height! and !pixel_width! macros - but it knows the scale factor by itself.

nyurik:
* TBH I would prefer to add a percentage param to `ST_TileEnvelope`. Writing `ST_Expand(ST_TileEnvelope(:zoom, :x, :y), (:extent/2) * ST_PixelSize(:mag: ))` is far less readable and slower than `ST_TileEnvelope(:zoom, :x, :y, :extent/2/100)`.  Also  `ST_TileEnvelope()` by itself has little value at least in [OpenMapTiles](https://github.com/openmaptiles/openmaptiles/tree/master/layers) because it always has to be wrapped with `ST_Expand` -- all layers have a buffer to prevent clipping. Moreover, `ST_AsMVTGeom` itself defaults to 256 - as there are very few cases when the tile envelope does NOT need to be expanded.  Thus I think it really should be a parameter rather than a slower chain of function calls.
* a single pixel size works for Web Mercator, but not for other projections (e.g. [examples](https://www.maptiler.com/coordinate-systems/)), where vertical & horizontal pixel size differ. It seems there is now more and more demand for non-3857. ST_AsMVTGeom removing sub-pixel objects should account for that. `ST_PixelSize()` would make it more difficult.
* If Volkswagen is more suited for polygons than Douglas-Peucker, maybe we should just pass it as a parameter to `ST_AsMVTGeom`?  E.g. `simplifier='dp' / 'vw' / ...` ?
* I don't think we should break backwards compatibility by no longer simplifying geometries by default.

What do you think?

Algunenano:
> WDYT?

I have the same opinion as you. I think getting rid of the simplification step should be the way to go here but, the alternative of asking people to call it themselves isn't good enough. To do it properly we'd turn something like:
```ST_AsMVTGeom(geom, bounds...)```
into
```ST_AsMVTGeom(ST_RemoveRepeatedPoints(ST_Simplify(geom, ST_TileExtent(z := 10) / 4096), ST_TileExtent(z := 10) / 4096), bounds```

Which is hard to accept now (after 3 releases without this monstrous setup).

In the ideal world, I'd like to be able to remove simplification without losing (too much) performance in the corner cases we've already detected.
I haven't had enough time to dive into wagyu to see what's different in those cases but if we could improve them there then we'd be able to drop the simplification step ( or change to tolerance 0 at the end) and not require the external calls.

**Comment by Algunenano** [comment]
> a single pixel size works for Web Mercator, but not for other projections (e.g. examples), where vertical & horizontal pixel size differ. It

MVTs are built based on square extents, at least up to the latest released spec. If you want special support for non square projections you should request it to be added in the [3.0 spec](https://github.com/mapbox/vector-tile-spec/issues/103) first.

Note that this doesn't block you from using it with other SRIDs that aren't square, just that there will be some distortion. [This example](https://bl.ocks.org/makella/ffcd3be98c5354d384008528d0e11fe7) is using Equal Earth with MVTs.

> One open question - if the new algo remove a polygon that shrinks to a single pixel? 

Yes, a collapsed polygon (less than 3 different vertices) is invalid and it should be dropped by ST_AsMVTGeom.

**Comment by peteryangthinkgeo** [comment]
Hi Team,

I found some issues about  ST_AsMVT or ST_AsMVTGeom after upgrading to PostGIS 3.0. I have polygon it shows well in the MVT byta with PostGIS 2.x, but after upgrading to PostGIS 3.0 I found the inner rings are missing. After doing some research, seems that in PostGIS 3.0 the ST_AsMVT calls the simplification process inside. I guess the problem I met is caused by the simplification process, where can I found the tolerance or how can I disable that?

Thanks,
Peter

**Comment by nyurik** [comment]
@peteryangthinkgeo I think you are hitting a known bug that has been fixed in master. I have hit the same issue, and had to build dev-version which solved it.

**Comment by Algunenano** [comment]
I agree, it's highly likely that you are affected by the simplification already fixed in the dev branches. To confirm you can either build from source (stable-3.0 branch), share the query and geometry or wait until 3.0.1 is released.

**Comment by peteryangthinkgeo** [comment]
@nyurik Thanks for your info. So does it mean I need to install PostGIS from source code also including the dependences such as GEOS, Proj, GDAL, LibXML2 and JSON-C, or I can use xxx-devel instead?

**Comment by peteryangthinkgeo** [comment]
Hi guys,

I fixed this issue by install the develop version, but however I met another issues about the buffer.  Below shows the SQL statement I use to generate the vector tile.  Even I set the buffer wehn calling the `st_asmvtgemom` function, but the tile didn't include a buffer as I expected. Anyone knows this?

```
select into water_mvt 
    st_asmvt(mvt_raw.*,'water',tile_size,'shape')
    from 
    (
        select "class", st_asmvtgeom("shape",val_bounding_box,tile_size,water_buffer,false) as "shape" 
        from "{schema}"."osm_water_09_view" 
        where "shape" && val_bounding_box
    ) mvt_raw where "shape" is not null;
```

Algunenano:
The val_bounding_box in the intersection needs to include the extra buffer too, otherwise the points in the buffer are discarded before reaching St_AsMVTGeom. https://github.com/postgis/postgis/pull/514 might give you some more context into the isssue

nyurik:
Peter, you may find the new (master branch only) version of ST_TileEnvelope
useful -- it allows you to increase the tile size by a percentage - e.g. if
your buffer is 256 and extent is 4096, set margin => 256/4096

peteryangthinkgeo:
@Algunenano and @nyurik thanks for the replies. I think that's is the reason causes the issue I met, but I have questions there, I am using develop version, and I don't see ST_TileEnvelop allows to pass a percentage of margin. Is there a workaround for develop version?

nyurik:
Peter, margin has been added to master, but has not yet been released. See
https://postgis.net/docs/manual-dev/ST_TileEnvelope.html

nyurik:
There is also an example in
https://postgis.net/docs/manual-dev/ST_AsMVT.html describing how to use
margin parameter with ST_AsMVT()

peteryangthinkgeo:
@nyurik My understanding is that I need to install PostGIS from master feature branch right? If yes, I have a question want to confirm. I am using the version [postgis-3.0.1dev.tar.gz](https://postgis.net/stuff/postgis-3.0.1dev.tar.gz) which fixed the issue about the MVT simpilifcation issue, does it also included in the master branch?

peteryangthinkgeo:
From the official website, I see there is a package [postgis-3.1.0dev.tar.gz](https://postgis.net/stuff/postgis-3.1.0dev.tar.gz), does it include the new functions?
@nyurik

peteryangthinkgeo:
@Algunenano @nyurik Thanks for your replies, I resolved this issues. It turns out to be an issue with the bounding box in the where clause, which should be buffered with the margin too.

robe2:
@Algunenano  @Komzpa  more to be done here?

Algunenano:
> @Algunenano @Komzpa more to be done here?

There hasn't been any changes since the last review. This issue is tricky since the simplification order impacts performance (and to top it all, I no longer have access to those performance tests).

Komzpa:
I'd say we merge it with NEWS notice that if one experiences performance degradation they can add ST_Simplify under ST_AsMVTGeom. Otherwise we don't have a way to get non-simplifying MVT transform.

Algunenano:
> I'd say we merge it with NEWS notice that if one experiences performance degradation they can add ST_Simplify under ST_AsMVTGeom.

Can you verify that this keeps the same performance as without the change?

> Otherwise we don't have a way to get non-simplifying MVT transform.

There are other ways, like extra flags or extra signatures.

I'm not against a change like this, but seeing that there were some cases that went from 120 ms to 380 ms I'm worried about making it the default. It's true that any of the 2 behaviours could have been the default, but this has been present for 3 releases already, so I'd rather replace it for something that was better most of the time, not something better some times and much worse other times.

Algunenano:
I'm not sure by whom, but this has been merged without improvements, docs or even a reply to the comment I could see. I wonder why I bothered testing or reviewing it at all..

robe2:
> I'm not sure by whom, but this has been merged without improvements, docs or even a reply to the comment I could see. I wonder why I bothered testing or reviewing it at all..

@Algunenano @Komzpa   Which tests were you using to test?  I would like to add them to https://git.osgeo.org/gitea/postgis/postgis-performance  

That said after we do our testing after 3.2.0 alpha1 release, if we agree it's too much of a performance hit, we can take it out as it didn't change any signatures. 

If we want to keep the behavior then we should update the docs to reflect, but given @Algunenano  concerns, I'm not so sure we will be keeping it so am fine with lack of docs at moment.

pramsey:
Don't like to see merges in the middle of convos, and also not to see performance regressions! (Users hate those too)

PR #464, opened 2019-08-19
ST_ConvexHullAgg

pramsey:
As described in https://trac.osgeo.org/postgis/ticket/3770,
Take in a set of geometry and return convex full of full set.

### DIFF
--- a/doc/reference_processing.xml
+++ b/doc/reference_processing.xml
@@ -857,6 +857,42 @@ SELECT ST_ConcaveHull(ST_GeomFromText('MULTILINESTRING((106 164,30 112,74 70,82 
 	  </refsection>
 	</refentry>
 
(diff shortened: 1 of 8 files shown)


pramsey:
☝️ @robe2

pramsey:
OK, ran some tests against this, and the performance is terrible. So the assumptions in the trac ticket
* this might have less memory pressure, yes
* this might be faster, no

robe2:
Should we give up on this as hopeless?  Only real reason for having it is if it could be faster than ST_ConvexHull(ST_Collect(..)).  If it can't then the extra keystrokes is not worth the addition of a new agg.

robe2:
I'm closing this one out as I think we decided it's hopeless

PR #468, opened 2019-08-28
Speed up ST_RemoveRepeatedPoints

Algunenano:
### First query

The table contains 1083175 multipolygons with an average of 9.01 points per geometry.

```sql
explain analyze Select ST_RemoveRepeatedPoints(the_geom_webmercator, 1) from public.benchmark_af85af967525e69373b7b8678f3bd409a8e78f447671344b7535d;
```

Trunk:
```
duration: 20 s
number of transactions actually processed: 66
latency average = 306.126 ms
```

With this PR:
```
number of transactions actually processed: 77
latency average = 263.052 ms
```

### Second query

13 multipolygons with an average of 260843.07 points per geometry.

```sql
explain analyze Select ST_RemoveRepeatedPoints(the_geom_webmercator, 1) from public.benchmark_4c7214d90a79aa6760367a084a4d4a2f61fbe1c6cc4f7f9e76020;
```

Trunk:
```
number of transactions actually processed: 241
latency average = 83.137 ms
```

With this PR:
```
number of transactions actually processed: 375
latency average = 53.409 ms
```

### Explanation

- PG facing (`ST_RemoveRepeatedPoints`)
  - Use `PG_GETARG_GSERIALIZED_P_COPY` and `lwgeom_remove_repeated_points_in_place`: In some cases (like the second query) this avoids an extra geometry clone (one to copy the gserialized and the second one done in lwgeom_remove_repeated_points).
  - Remove the calls to `lwgeom_free`. It will be freed by PG automatically in a faster way. I've kept the call to free the GSERIALIZED copy because the comment in PGs code doesn't give me enough peace of mind (https://github.com/postgres/postgres/blob/bde7493d10898831100a0c6c233a5f3030bfcecd/src/include/fmgr.h#L251-L253)

- liblwgeom (`ptarray_remove_repeated_points_in_place`):
  - Keep a reference to where we want to write and use memcpy instead of `ptarray_copy_point`. With this we avoid calculating the position every time and also the size of the point (which won't change).

  ~~- Move the last_point special case outside the loop. This doesn't have a really big impact since the CPU is pretty good at predicting those branches.~~

### DIFF
--- a/liblwgeom/ptarray.c
+++ b/liblwgeom/ptarray.c
@@ -1464,10 +1464,9 @@ ptarray_remove_repeated_points_in_place(POINTARRAY *pa, double tolerance, uint32
 	if ( n_points <= min_points ) return;
 
 	last = getPoint2d_cp(pa, 0);
-	for (i = 1; i < n_points; i++)
+	void *p_to = ((char *)last) + pt_size;
+	for (i = 1; i < n_points - 1; i++)
 	{
-		int last_point = (i == n_points-1);
-
 		/* Look straight into the abyss */
 		pt = getPoint2d_cp(pa, i);
 
@@ -1479,7 +1478,7 @@ ptarray_remove_repeated_points_in_place(POINTARRAY *pa, double tolerance, uint32
 				/* Only drop points that are within our tolerance */
 				dsq = distance2d_sqr_pt_pt(last, pt);
 				/* Allow any point but the last one to be dropped */
-				if (!last_point && dsq <= tolsq)
+				if (dsq <= tolsq)
 				{
 					continue;
 				}
@@ -1490,21 +1489,36 @@ ptarray_remove_repeated_points_in_place(POINTARRAY *pa, double tolerance, uint32
 				if (memcmp((char*)pt, (char*)last, pt_size) == 0)
 					continue;
 			}
+		}
 
+		/* Compact all remaining values to front of array */
+		memcpy(p_to, pt, pt_size);
+		n_points_out++;
+		p_to += pt_size;
+		last = pt;
+	}
+
+	/* Last point has a different behaviour */
+	{
+		pt = getPoint2d_cp(pa, n_points - 1);
+		if ((n_points + n_points_out > min_points + i) && (n_points_out > 1) && (tolerance > 0.0))
+		{
 			/* Got to last point, and it's not very different from */
 			/* the point that preceded it. We want to keep the last */
 			/* point, not the second-to-last one, so we pull our write */
 			/* index back one value */
-			if (last_point && n_points_out > 1 && tolerance > 0.0 && dsq <= tolsq)
+			dsq = distance2d_sqr_pt_pt(last, pt);
+			if (dsq <= tolsq)
 			{
 				n_points_out--;
+				p_to -= pt_size;
 			}
 		}
 
-		/* Compact all remaining values to front of array */
-		ptarray_copy_point(pa, i, n_points_out++);
-		last = pt;
+		memcpy(p_to, pt, pt_size);
+		n_points_out++;
 	}
+
 	/* Adjust array length */
 	pa->npoints = n_points_out;
 	return;
--- a/liblwgeom/ptarray.c
+++ b/liblwgeom/ptarray.c
@@ -1465,8 +1465,10 @@ ptarray_remove_repeated_points_in_place(POINTARRAY *pa, double tolerance, uint32
 
 	last = getPoint2d_cp(pa, 0);
 	void *p_to = ((char *)last) + pt_size;
-	for (i = 1; i < n_points - 1; i++)
+	for (i = 1; i < n_points; i++)
 	{
+		int last_point = (i == n_points - 1);
+
 		/* Look straight into the abyss */
 		pt = getPoint2d_cp(pa, i);
 
@@ -1478,7 +1480,7 @@ ptarray_remove_repeated_points_in_place(POINTARRAY *pa, double tolerance, uint32
 				/* Only drop points that are within our tolerance */
 				dsq = distance2d_sqr_pt_pt(last, pt);
 				/* Allow any point but the last one to be dropped */
-				if (dsq <= tolsq)
+				if (!last_point && dsq <= tolsq)
 				{
 					continue;
 				}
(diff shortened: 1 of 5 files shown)


liblwgeom/ptarray.c
```diff
@@ -1498,11 +1499,14 @@ ptarray_remove_repeated_points_in_place(POINTARRAY *pa, double tolerance, uint32
 			if (last_point && n_points_out > 1 && tolerance > 0.0 && dsq <= tolsq)
 			{
 				n_points_out--;
+				p_to -= pt_size;
 			}
 		}
 
 		/* Compact all remaining values to front of array */
-		ptarray_copy_point(pa, i, n_points_out++);
```
Komzpa: can this be done by, say, moving the ptarray_copy_point into lwinline.c? should render to similar assembly if inlined, and worth investigating if not.

Algunenano: > can this be done by, say, moving the ptarray_copy_point into lwinline.c? should render to similar assembly if inlined, and worth investigating if not.

I've tested it with https://github.com/Algunenano/postgis/commit/9508c2d8e870ee895996bf05de91f1187e8996a0 and it's slower than the approach in this PR:

```
number of transactions actually processed: 73
latency average = 275.065 ms
```
```
number of transactions actually processed: 307
latency average = 65.338 ms
```

But it might be useful for other cases.

Algunenano:
> I'm not sure this is a good behavior to set on a function-by-function basis. Can you set lwfree to a no-op if you want to let PG handle cleanup?

AFAIK, yes you can with some exceptions. In fact I created an issue related to this: https://trac.osgeo.org/postgis/ticket/4264 but I haven't had time to look into this in detail.

Basically the idea is that PG resets the memory context after every transaction, so you don't need to delete intermediate objects unless they are going to amount to a lot (big allocations, loops, recursive functions) or you are not using PG allocation (GEOS objects). 

I believe this is common practise in Postgres codebase, but I can't find any comment to reference it now, just examples:
- The ASCII functions: https://github.com/postgres/postgres/blob/8255c7a5eeba8f1a38b7a431c04909bde4f5e67d/src/backend/utils/adt/ascii.c#L120

And counter examples:
- The varlena functions (Used for sorting thus the garbage is kept longer?): https://github.com/postgres/postgres/blob/8255c7a5eeba8f1a38b7a431c04909bde4f5e67d/src/backend/utils/adt/varlena.c#L2135

Algunenano:
I totally get what you mean and, ideally, I'd like to just make lwfree a no-op; but to do that I'd need to review the hundreds of calls to lwalloc (and it's callers) to make their deletion can be deferred until the end of the transaction.

One thing to note is that I removed the call to lwfree in the extension code (`postgis/`), not in the library (`liblwgeom`) which should still be memory safe, it's just that I've chosen not to manually delete the memory I've gotten from the library because I know it's going to be recovered as soon as Postgres removes the memory context (which is done immediately after and it's faster as it doesn't need to traverse structures).

There are other places (also in the extension code) that also make use of this, for example `mvt_geom` doesn't care if you either modify or create a new geometry in `mvt_clip_and_validate`; it will return just that and leaves Postgres the task of deleting everything at the end. Looking through the code I see there are other functions (`ST_DistanceRectTree`, `geography_point_outside`) functions don't call `lwgeom_free` before returning but most of them do, although not always in error handling.

Algunenano:
I realized that changing the gserialized read from the normal (PG_GETARG_GSERIALIZED_P) to a copy (PG_GETARG_GSERIALIZED_P_COPY) might have a impact on points so I decided to test it too with a 1M rows dataset:

```
explain analyze Select ST_RemoveRepeatedPoints(the_geom_webmercator, 1) from public.yellow_1m;
```

Trunk:
```
number of transactions actually processed: 144
latency average = 139.095 ms
```

This PR:
```
number of transactions actually processed: 143
latency average = 140.610 ms
```

So there might be a slight penalty for that case (a no-op) but it's not clear.

Algunenano:
After rebasing on top of trunk and adding a shortcut to optimize the no-op case (return the input when no points were removed):
* Trunk
- Tolerance: 0.00001 (no-op): ~220 its / 30 secs
- Tolerance: 1 (~1% points removed): ~230 its / 30 secs
- Tolerance: 1000000 (~98% of points removed): ~360 its / 30 secs

* This PR
- Tolerance: 0.00001 (no-op): ~615 its / 30 secs --> **2.8x as fast**
- Tolerance: 1 (~1% points removed): ~350 its / 30 secs --> **1.5x as fast**
- Tolerance: 1000000 (~98% of points removed): ~720 its / 30 secs --> **2x as fast**

Dataset: 13 multipolygons with an average of 260843.07 points per geometry

dr-jts:
> I totally get what you mean and, ideally, I'd like to just make lwfree a no-op; but to do that I'd need to review the hundreds of calls to lwalloc (and it's callers) to make their deletion can be deferred until the end of the transaction.

How about making a second function `lwfree_noop` (or something) which is a no-op, and has a comment indicating that this only works if running in the context of a memory pool manager?

This captures the fact that memory has been allocated and would need to be freed if not running under the security of Postgres?

Algunenano:
> How about making a second function lwfree_noop (or something) which is a no-op, and has a comment indicating that this only works if running in the context of a memory pool manager?

That's a nice way to distinguish between them, but again it requires revisiting all lwfree calls to distinguish between those where it's safe to not free them under Postgresql context and those where it isn't. In the current case in particular, the lwfree that was removed is always under postgres context (it's not inside the lwgeom library) so removing it it's safe.

Anyhow, ideally I'd rather not allocate memory so we don't need to deallocate it :smile:, but that's again a lot of work that i'd rather do in a case by case situation.

dr-jts:
> > How about making a second function lwfree_noop (or something) which is a no-op, and has a comment indicating that this only works if running in the context of a memory pool manager?
> 
> That's a nice way to distinguish between them, but again it requires revisiting all lwfree calls to distinguish between those where it's safe to not free them under Postgresql context and those where it isn't. 

> In the current case in particular, the lwfree that was removed is always under postgres context (it's not inside the lwgeom library) so removing it it's safe.

I wasn't thinking that work would need to be done right away.  Just that it would establish a new pattern for future cleanup, and make this particular case visible. But if this `lwfree` isn't needed ever, then it makes sense to just remove it.

strk:
On Fri, Sep 13, 2019 at 09:50:57AM -0700, Martin Davis wrote:
> > I totally get what you mean and, ideally, I'd like to just make lwfree a no-op; but to do that I'd need to review the hundreds of calls to lwalloc (and it's callers) to make their deletion can be deferred until the end of the transaction.
> 
> How about making a second function `lwfree_noop` (or something) which is a no-op, and has a comment indicating that this only works if running in the context of a memory pool manager?
> 
> This captures the fact that memory has been allocated and would need to be freed if not running under the security of Postgres?

`lwfree` is supposed to be definable by user. PostgreSQL user defines it
to `pfree`, while cunit user defines it to `free`.

Please beware of raises in memory usage when postponing release, test
with big tables.

PR #475, opened 2019-09-06
ptarray_calculate_gbox_cartesian speed up

Algunenano:
Warning: For now this is exploratory, but comments are welcome.

After fixing several functions not updating the bbox of the output geometry (ST_RemoveRepeatedPoints, ST_Simplify and ST_SnapToGrid) those functions have received an important performance impact; to the point where they might spend more time calculating the bbox than doing the actual work; e.g.:
```
   - 97.47% ST_RemoveRepeatedPoints                                                                                                                                                                                                         ▒
      + 32.26% lwgeom_calculate_gbox_cartesian                                                                                                                                                                                              ▒
      + 23.72% lwgeom_remove_repeated_points_in_place                                                                                                                                                                                       ▒
      + 22.67% heap_tuple_untoast_attr                                                                                                                                                                                                      ▒
      + 16.47% gserialized2_from_lwgeom                                                                                                                                                                                                     ▒
      + 2.31% lwgeom_from_gserialized2
```

So I felt like I had to do something to win that time back (:D) and that's where this PR comes in:
* Calculates the cartesian bbox of a ptarray without copying points.
* Uses 3 different functions depending on the coordinate number (2, 3, 4) to avoid the 2 if calls inside the loop.

Since bbox calculation is used pretty much every time there is a line or polygon generation, I'll do some extra benchmarks to see if I can squeeze this even further.

### DIFF
--- a/liblwgeom/gbox.c
+++ b/liblwgeom/gbox.c
@@ -533,46 +533,114 @@ static int lw_arc_calculate_gbox_cartesian(const POINT4D *p1, const POINT4D *p2,
 	return rv;
 }
 
-int ptarray_calculate_gbox_cartesian(const POINTARRAY *pa, GBOX *gbox )
+static void
+ptarray_calculate_gbox_cartesian_2d(const POINTARRAY *pa, GBOX *gbox)
 {
-	uint32_t i;
-	POINT4D p;
-	int has_z, has_m;
+	const POINT2D *p = getPoint2d_cp(pa, 0);
 
-	if ( ! pa ) return LW_FAILURE;
-	if ( ! gbox ) return LW_FAILURE;
-	if ( pa->npoints < 1 ) return LW_FAILURE;
+	gbox->xmax = gbox->xmin = p->x;
+	gbox->ymax = gbox->ymin = p->y;
 
-	has_z = FLAGS_GET_Z(pa->flags);
-	has_m = FLAGS_GET_M(pa->flags);
+	for (uint32_t i = 1; i < pa->npoints; i++)
+	{
+		p = getPoint2d_cp(pa, i);
+		gbox->xmin = FP_MIN(gbox->xmin, p->x);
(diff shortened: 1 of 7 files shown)


Algunenano:
Note: Rebased with just these change (I've done the coding on top of https://github.com/postgis/postgis/pull/468 but they should be treated separately).

pramsey:
Just as a general note regarding re-calculating bounding boxes after altering the point list: it might be better just to free and null the existing box and let re-calculation wait until the object hits some function that actually needs the box. By explicitly re-filling it, you're potentially getting into a place where (if the coordinates are altered several times in a processing chain) the box gets re-calculated multiple times. 

(This would be even more true if we were using expanded-object-headers, but since we're serializing at the end of every function call, it won't happen often.) 

But `asmvtgeom` would seem to be one of those rare cases where there's multiple operations in a row that mutate coordinates, and where bounding box calculation might be deferred until the very end).

Making the process faster in general, as you're doing here, is awesome regardless.

Algunenano:
> Just as a general note regarding re-calculating bounding boxes after altering the point list: it might be better just to free and null the existing box and let re-calculation wait until the object hits some function that actually needs the box. By explicitly re-filling it, you're potentially getting into a place where (if the coordinates are altered several times in a processing chain) the box gets re-calculated multiple times.

I agree. I was talking with Darafei the other way about this and we could "simply" call `lwgeom_drop_bbox` when we modify any ptarray. I didn't do it in those tickets since "simply" is not all that simple if you want to do it lazily, that is, only drop the box when  (on point modification, not on function called).

I intend to revisit that just for 3.0 with an extra twist. The idea I have now is to have lwgeom_remove_repeated_points_in_place (for example) return a boolean indicating if there was any change in coordinates, then if there was a change drop the bbox, but it there wasn't then return the gserialized input argument, thus avoiding serializing the geometry.

PR #478, opened 2019-09-19
[EXPERIMENTAL] Speed up LWGEOM_noop

Algunenano:
To continue with the streak of speeding up things, I decided to look up the serialization code and see where I can squeeze some extra performance and to do that there is nothing better than looking at `postgis_noop`:

This is a WIP and might change at any time, but right now I have:
* da7c51e: Pass the SRID during the geometry creation to avoid needing to update it later. This is only _slightly_ noticeable with big multigeometries, since those set its SRID in all subgeometries, so doing it during creation avoids all the memory jumps from iterating over it.
* d3676bf: This needs madness glasses. I've introduced a new function (`geometry_serialize_reuse`) that allows you to pass an already allocated buffer to serialize on top of it. For this to works it needs:
  * The gserialized buffer needs to a copy (`PG_GETARG_GSERIALIZED_P_COPY`), not a pointer that might point to the disk (`PG_GETARG_GSERIALIZED_P`).
  * The input buffer needs to be bigger or equal than the space required to store the geometry. I've added a safeguard for that to fallback to the old allocation if it isn't.
  * The tricky part: The geometry must either not contain no reference to the gserialized buffer (for example if it was cloned, passed to geos and back, etc) or, if it does, it's references keep the order that was initially given during the deserialization; i.e. when you read the geometry you don't find pointers to address in the buffer previous to address already pointed at. We are memmove'ing the data based on that knowledge.
  * When these conditions are met (which depends on the process done in liblwgeom and the call to geometry_serialize_reuse) this drops the time to gserialize a geometry masively; for example with big geometries, the run time of postgis_noop (deserialization + serialization) drops by over 50% (from 39ms to 18ms).

I want to see where this might take me since the impact of serialization is really important in already highly optimized functions.

### DIFF
--- a/liblwgeom/gserialized.c
+++ b/liblwgeom/gserialized.c
@@ -221,7 +221,13 @@ int gserialized_ndims(const GSERIALIZED *g)
 */
 GSERIALIZED* gserialized_from_lwgeom(LWGEOM *geom, size_t *size)
 {
-	return gserialized2_from_lwgeom(geom, size);
+	return gserialized2_from_lwgeom_reuse(geom, size, NULL);
+}
+
+GSERIALIZED *
+gserialized_from_lwgeom_reuse(LWGEOM *geom, size_t *size, GSERIALIZED *g_in)
+{
+	return gserialized2_from_lwgeom_reuse(geom, size, g_in);
 }
 
 /**
--- a/liblwgeom/gserialized2.c
+++ b/liblwgeom/gserialized2.c
@@ -815,16 +815,16 @@ static size_t gserialized2_from_lwpoint(const LWPOINT *point, uint8_t *buf)
 	loc = buf;
 
 	/* Write in the type. */
-	memcpy(loc, &type, sizeof(uint32_t));
+	memmove(loc, &type, sizeof(uint32_t));
 	loc += sizeof(uint32_t);
 	/* Write in the number of points (0 => empty). */
-	memcpy(loc, &(point->point->npoints), sizeof(uint32_t));
+	memmove(loc, &(point->point->npoints), sizeof(uint32_t));
 	loc += sizeof(uint32_t);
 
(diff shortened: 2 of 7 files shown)


pramsey:
I think that if you're bumping up against serialization as a major overhead it might make sense to visit expanded object headers again... that way in function(function(function())) chains we can avoid serialization overhead entirely. The downside of EOH though is that for ordinary function() calls we potentially incur a spare copy step that otherwise wouldn't happen. However, if you haven't spent some mental cycles exploring EOH, this might be the time to start.

Algunenano:
> However, if you haven't spent some mental cycles exploring EOH, this might be the time to start.

It makes sense. Apart from PG code, do you know where to start looking into this? By chance, did you try / mock an implementation at any point?

pramsey:
Relevant Pg interfaces are described here: https://github.com/postgres/postgres/blob/master/src/include/utils/expandeddatum.h
An implementation from a few years ago: https://github.com/pramsey/postgis/tree/expanded-object/postgis

pramsey:
I gave up early on EOH, potentially from a mis-conception of how our implementation would work. My concern was that if you commit to expanding to an EOH in deserialization you end up always having to create a copy, which for a use case like ST_AsGeoJSON(geom) is stupid. But maybe (and here my recollection gets fuzzy) what's really needed is for deserialization to be sensitive to both fully serialized and EOH inputs, for all functions, but for only geometry-returning functions to return EOH. The main "problem" with EOH, in my estimation was just that the overhead in our deserialization was so low that avoiding it wasn't worth the extra machinery and the potential extra copying involved in setting up an EOH in the first place. For comparison, Tom invented EOH in order to allow the in-memory format for arrays to persist through operations, usually in PL/PgSQL, so that things like array appends in loops and so on would not be terribly performers.

strk:

PR #482, opened 2019-09-24
[WIP] Adjust function costs

Algunenano:
I'm opening this as as WIP while I'm trying to find a more standard way to setup costs for functions, or at least update the costs that might have become obsolete due to improvements in the C code.

As we don't have cost per geometry size, I think the best philosophy for now would be to err on the higher side of things; getting parallelism when you don't need it has less impact than not getting it when you do. 

Steps taken (WIP):
* Explicitly set up as _COST_DEFAULT those functions that work with raw serialized geometries. These functions are expected to read the gserialized input just once (at most), so they don't allocate memory or chase collection pointers. 

Other steps I want to take before 3.0 final release:
* Review the costs of the functions using `LANGUAGE 'sql'` since by default they'll have cost 100, which might not be what we want. Better to be explicit.
* Review the costs of some of the recently modified functions (ST_RemoveRepeatedPoints, ST_Simplify...).
* Review the costs of the MVT family functions.


Algunenano:
I've tried updating `ST_AsMVTGeom` costs to 10000 but I only see it making a difference in 2 benchmarks out of 50, and it's for the worse (they are slower as they get parallelized); all other cases remain the same.

I think it might make sense to raise the cost with also a raise in parallel_setup_cost, as several small queries (both mvt and not) are slower due to switching to a parallel plan when the sequential one is faster.

strk:

robe2:
@Algunenano  so it sounds like you decided not to merge this?  because better to increase parallel setup costs? Can we close this out then?

Algunenano:
> so it sounds like you decided not to merge this? because better to increase parallel setup costs?

No, because I didn't have time to test this properly. I wanted to modify the costs of some functions whose costs seems way higher or lower than others with equivalent performance, but I gathered this should be done with some kind of performance tests.

> Can we close this out then?

Sure, I'll reopen it in the future if necessary.

PR #483, opened 2019-09-27
Apply ordering to the results of geos UNION 

pramsey:
to avoid version-dependent edge ordering in topologies that break
regression


pramsey:
Going another way

PR #484, opened 2019-09-30
Change from 4326 to 4269 as target SRS for reproject test

pramsey:
... to avoid accidentally firing up grid-shift code in Proj, which
may or may not be installed for any given Proj install,
depending on whether the packager/installer chose
some/all/none of the grid shift files.


pramsey:
This has been applied.

PR #485, opened 2019-09-30
Remove schema qualification from magic upgrade-if-only-needed function

pramsey:


Algunenano:
LGTM. I've never run postgis using the scripts (no extension) so I didn't know this could be an issue :sweat:

pramsey:
This has been applied.

PR #486, opened 2019-10-01
Allow non-SRID geometries to insert into SRID-specific columns.

pramsey:
```
CREATE TABLE t (g GEOMETRY(POINT, 4326));
INSERT INTO t VALUES ('POINT(0 0)');
```


robe2:
Why do we want this?  Seems like a bad idea to me.

Makes sense for geography - which we already have doing this, but not geometry

Komzpa:
I like it.

This remaps "zero srid" from "is zero" to "not defined/any". Makes sense in terms of what typmod really does for all the other data types - like you can cast any numeric into numeric(2,5) and get it rounded/clipped correctly.

If we go this path to the end I see:

 - ST_Union(0 rows) => 'GEOMETRYCOLLECTION EMPTY', ST_Difference(some_geom_with_srid, ST_Union(0 rows)) -> some_geom_with_srid; this pulls out a lot of coalesce's in spots unexpected for newbies.

 - ST_SetSRID can be dropped on import in many software, if it can only output WKB and not EWKB.

dr-jts:
> I like it too. I'd even like SRID=0 types accepting non-0 SRID'ed geometry values, on that same line (not sure if this proposed change does it).

Wouldn't that risk ending up with different SRIDs into a single column?

strk:
> Wouldn't that risk ending up with different SRIDs into a single column?

Yes, but that'd not be a problem for me (if I don't constraint I
clearly don't care about ending up with different SRIDs into the
column).

We actually already do accept them, btw, I just realized :)

pramsey:
It more closely matches what other typmod types do when presented with input that has no preference regarding the typmod restrictions -- the incoming value inherits the typmod of the column. I think it makes sense, and as an added bonus it makes example SQL statements simpler because you don't have to constantly set the SRID on things as you insert them.

strk:

robe2:
I guess I can live with this as long as something like this still fails:

```
CREATE TABLE t (g GEOMETRY(POINT, 4326));
INSERT INTO t (g)  VALUES ( ST_GeomFromText('POINT(0 0)', 4269 )  );
```

robe2:
So when are we going to do this? or are we going to keep this open forever?

pramsey:
Committed at f95dfc7b4

https://trac.osgeo.org/postgis/ticket/4569

PR #490, opened 2019-10-07
Trunk warnings

Algunenano:
Fix for https://trac.osgeo.org/postgis/ticket/4527

### DIFF
--- a/postgis/lwgeom_functions_basic.c
+++ b/postgis/lwgeom_functions_basic.c
@@ -1546,12 +1546,12 @@ Datum LWGEOM_makepoly(PG_FUNCTION_ARGS)
 		holes = lwalloc(sizeof(LWLINE *) * nholes);
 		for (i = 0; i < nholes; i++)
 		{
-#if GCC_VERSION >= 40604
+#if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6)
 #pragma GCC diagnostic push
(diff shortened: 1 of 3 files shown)


strk:
How about fixing the sign-compare warning rather than hiding it ?

Algunenano:
> How about fixing the sign-compare warning rather than hiding it ?

It's completely under Postgresql code so we can't fix it from Postgis. The coding standards between the 2 projects are different, and AFAIK PG doesn't care about these kind of errors.

postgis/lwgeom_functions_basic.c
```diff
@@ -1546,12 +1546,12 @@ Datum LWGEOM_makepoly(PG_FUNCTION_ARGS)
 		holes = lwalloc(sizeof(LWLINE *) * nholes);
 		for (i = 0; i < nholes; i++)
 		{
-#if GCC_VERSION >= 40604
```
pramsey: If you're stripping out use of `GCC_VERSION` (which is fine, I just am too dim to do the major/minor/patch logic), can you also remove the definition from postgis_config.h.in?

Algunenano: I don't see any `GCC_VERSION` in my `postgis_config.h.in` or anywhere in the tree. Did you add it and forgot to push it?

pramsey:
I was wondering why you were opposed enough w/ my commit to strip it back out, but that explains it. Yeah, I didn’t commit that with the #defines. Sorry. Thanks for the change.

PR #495, opened 2019-10-10
WKB: Avoid buffer overflow

Algunenano:
This only happens when not running under PG context, as
lwerror continues execution and that means that even after
detecting there isn't enough bytes still try to read from the buffer

### DIFF
--- a/liblwgeom/lwin_wkb.c
+++ b/liblwgeom/lwin_wkb.c
@@ -36,14 +36,15 @@
 typedef struct
 {
 	const uint8_t *wkb; /* Points to start of WKB */
-	size_t wkb_size; /* Expected size of WKB */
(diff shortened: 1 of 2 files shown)


liblwgeom/lwin_wkb.c
```diff
@@ -519,12 +539,14 @@ static LWPOLY* lwpoly_from_wkb_state(wkb_parse_state *s)
 	for( i = 0; i < nrings; i++ )
 	{
 		POINTARRAY *pa = ptarray_from_wkb_state(s);
-		if( pa == NULL )
-			continue;
+		if (pa == NULL)
+			return NULL;
+		;
```
Komzpa:
```suggestion

```

PR #498, opened 2019-10-17
Remove set search_path when possible

Algunenano:


robe2:
one minor thing.  can you lowercase GEOMETRY.  I know it was like that and it just annoys me noticing it now that I had that like that all these years.

Algunenano:
> one minor thing. can you lowercase GEOMETRY. I know it was like that and it just annoys me noticing it now that I had that like that all these years.

Will do

Rebased and superseeded by https://github.com/postgis/postgis/pull/498

PR #501, opened 2019-10-18
Add tests around schema qualification

Algunenano:


regress/core/dump.sql
```diff
@@ -178,3 +178,14 @@ GEOMETRYCOLLECTION (
...
+-- Check that it works without the extension schema being available
+SET search_path TO pg_catalog;
+WITH data AS
+(
+    SELECT 'POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::public.geometry as geom
+)
+SELECT	't12',
+	public.ST_AsText((public.ST_Dump(geom)).geom),
+	public.ST_AsText((public.ST_DumpRings(geom)).geom) FROM data;
+RESET search_path;
```
strk: Please add newline at end of file

regress/core/dump.sql
```diff
@@ -178,3 +178,14 @@ GEOMETRYCOLLECTION (
 	)
 )
 ');
+
+-- Check that it works without the extension schema being available
+SET search_path TO pg_catalog;
+WITH data AS
+(
+    SELECT 'POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::public.geometry as geom
```
strk: I don't fully like the assumption that PostGIS is installed in the "public" schema.
Especially since we were talking about adding tests for it being installed in a foreign schema.
Any chance to use psql variables to define a schema ?

strk: Seems to work:
```
strk=# \set schema public.
strk=# select 'POINT(0 0)':: :schema geometry;
                  geometry                  
--------------------------------------------
 010100000000000000000000000000000000000000
(1 row)
```

Algunenano: I think it's doable. I could add it in the `run_simple_test` and set it to an option or `public` if not set (OPT_POSTGIS_SCHEMA or something like that).

Algunenano:
Pushed

PR #504, opened 2019-10-31
Fix oversimplification of polygon inner rings

Algunenano:
Reported in https://github.com/postgis/postgis/commit/86057e2e46a272838a54eff9e6ebb5e56f33fab7#r35741467


Komzpa:
is it also https://trac.osgeo.org/postgis/ticket/4556 ?

Algunenano:
I've tested it and it fixes that too

PR #506, opened 2019-11-07
Fix bug of mapping long pg field name to dbf field name

fallish:
pg field name should be passed to the second argument of function colmap_dbf_by_pg, but dbffieldname is truncated to 10 chars and will fail to match pg field name in column mapping file.


strk:
@fallish can you please provide an automated testcase for this ? Support for passing switches to loader/dumper should be already available, check out regress/loader for examples

PR #514, opened 2019-11-14
Add ST_TileEnvelope margin argument

nyurik:
Increase the `ST_TileEnvelope()` bbox  by a given margin percentage.

This solves a very common use case of generating MVT buffer data around a tile,
e.g. including nearby labels (points) that are outside of the tile, but the text is long
enough to spill over to the current tile. Without such filtering, `ST_TileEnvelope()`
is not as useful, because its result has to be grown by a certain percentage.

See [email thread](https://lists.osgeo.org/pipermail/postgis-devel/2019-November/028322.html).

The result is clipped along the vertical (Y) axis according to bounds argument,
and is allowed to spill over the antimeridian along the horizontal (X) axis.
If the total width is more than the entire X-bounds, the width is also clipped to the bounds.
Negative value of up to `-0.5` (-50%) are also allowed to shrink the bbox rather than grow it.


nyurik:
I tried to figure out what magic generates this code, but alas... a PostGIS guru is needed :(

```

psql:/drone/src/regress/00-regress-install/share/contrib/postgis/postgis_proc_set_search_path.sql:155: ERROR:  syntax error at or near "="
--
LINE 1: ...om integer, x integer, y integer, bounds geometry =3857;LINE...
                                                             ^
QUERY:  ALTER FUNCTION  ST_TileEnvelope(zoom integer, x integer, y integer, bounds geometry =3857;LINESTRING(-20037508.342789 -20037508.342789, 20037508.342789 20037508.342789 ) SET search_path=public,pg_catalog;
CONTEXT:  PL/pgSQL function inline_code_block line 23 at EXECUTE
```

nyurik:
Digging a bit further - it seems the regex-based [SQL parser](https://github.com/postgis/postgis/blob/master/utils/postgis_proc_set_search_path.pl#L147) (sigh) does not handle the default values properly when it sees a complex default value like `DEFAULT 'SRID=3857;LINESTRING(-20037508.342789 -20037508.342789, 20037508.342789 20037508.342789)'::geometry` ...   I tried to replace that regex with `s/(default\s+((\'[^\']*\')?[A-Za-z\.\+\-0-9\[\]\:\s]*))//ig;`, but it seems the closing  quote and parenthesis  get stripped before it gets to that regex :(

Komzpa:
How about ST_DWithin(geom, ST_TileEnvelope(), 50px)?

nyurik:
@Komzpa not exactly the same, esp because how difficult it is to use it: 
 `ST_DWithin(geom, ST_TileEnvelope(:zoom, :x, :y), 20037508.34 * 2 / (2 ^ :zoom) / 4096 * (1 + 50%))`.  Compare it with `geom && ST_TileEnvelope(:zoom, :x, :y, 50%)`.

Also, I might be mistaken, but `ST_DWithin` would work in its geometry variant (distance in the SRID of the geometry), but it won't work with geographies (e.g. when we need to find something across the antimeredian) because in that case the distance has to be in meters, and that would be very different for web mercator (esp because width and height are very different there, depending on the location)

Lastly, not sure if this applies, but MapTiler recently demoed [ESPG:4326](https://cloud.maptiler.com/maps/basic-4326/) MVT tiles (using OpenLayers UI), and it would be great to have a better support for such usages. Percentage increase might do it better than meter distances?

golubev:
It would be a nice feature :+1:

lseelenbinder:
We're hoping to use `ST_TileEnvelope` to help with rendering vector tiles directly in PostGIS, but without the bounds this gets super tricky, so we're currently using our own buffering logic in application code to make it happen. Having it baked-in is less code for us worry about. 😄

klokan:
This is a good idea - practical for OpenMapTiles.org project. +1 for merging.

robe2:
I see the tests are all in place and a lot of people evidentally want this.  Seems fine to me.

postgis/lwgeom_functions_basic.c
```diff
@@ -2065,7 +2066,11 @@ Datum ST_TileEnvelope(PG_FUNCTION_ARGS)
 	x = PG_GETARG_INT32(1);
 	y = PG_GETARG_INT32(2);
 
-	bounds = PG_GETARG_GSERIALIZED_P(3);
+	margin = PG_GETARG_FLOAT8(3);
+	if (margin < -0.5)  // shrinking by more than 50% would eliminate the tile outright
```
Algunenano: Please use `/* */` all added comments.

postgis/lwgeom_functions_basic.c
```diff
@@ -2088,10 +2093,27 @@ Datum ST_TileEnvelope(PG_FUNCTION_ARGS)
...
 	tileGeoSizeY = boundsHeight / worldTileSize;
-	x1 = bbox.xmin + tileGeoSizeX * (x);
-	x2 = bbox.xmin + tileGeoSizeX * (x+1);
-	y1 = bbox.ymax - tileGeoSizeY * (y+1);
-	y2 = bbox.ymax - tileGeoSizeY * (y);
+
+	// 1 margin (100%) is the same as a single tile width
+	// if the size of the tile with margins span more than the total number of tiles,
+	// reset x1/x2 to the bounds
+	if (1 + margin * 2 > worldTileSize)
```
Algunenano: Can you add parenthesis here to clarify the order of operations?

nyurik:
@Algunenano thanks for reviewing.  I fixed the inline comments.  WRT your first point - I was trying to do that too, but I do not think it is possible without breaking backwards compatibility.  The existing function signature is

`ST_TileEnvelope(integer tileZoom, integer tileX, integer tileY, geometry bounds=SRID=3857;LINESTRING(-20037508.342789 -20037508.342789,20037508.342789 20037508.342789));`

so adding a new parameter as last would require any user who want to use this function in a very common default way to copy that entire long default `bounds` parameter.  Inserting `margin` as the fourth would break backwards compatibility. Thus the only real option is to introduce two different SQL functions, while maintaining just one C function.

WRT tests - could you clarify what kind of a test you would like to see?
Thanks!

Algunenano:
> so adding a new parameter as last would require any user who want to use this function in a very common default way to copy that entire long default bounds parameter. Inserting margin as the fourth would break backwards compatibility. Thus the only real option is to introduce two different SQL functions, while maintaining just one C function.

You can always use named parameters in your postgresql function, so calling it with something like `ST_TileEnvelope(0,0,0, margin=> 0.1)` or the more verbose `ST_TileEnvelope(tileZoom=>0, tileX=>0, tileY=>0, margin=>0.1) is an option, isn't it?

> WRT tests - could you clarify what kind of a test you would like to see?

I'd expect ST_TileEnvelope(0,0,0) to be `POLYGON((-20037508.342789 -20037508.342789,-20037508.342789 20037508.342789, 20037508.342789 20037508.342789, 20037508.342789 -20037508.342789, -20037508.342789 -20037508.342789))`

Also, maybe showing some other tile, like `10,300,387` with multiple margins confirming they are ok and using floating point precision in its output.

nyurik:
@Algunenano thanks, I wasn't aware of the `=>` syntax. I updated the code to keep `ST_TileEnvelope` as a single function with a new default.

> I'd expect ST_TileEnvelope(0,0,0) to be POLYGON((-20037508.342789 -20037508.342789,-20037508.342789 20037508.342789, 20037508.342789 20037508.342789, 20037508.342789 -20037508.342789, -20037508.342789 -20037508.342789))

This is not what I see when using the current implementation of the `ST_TileEnvelope`:
```
# select ST_AsText(ST_TileEnvelope(0, 0, 0));
POLYGON((-20037510 -20037510,-20037510 20037510,20037510 20037510,20037510 -20037510,-20037510 -20037510))
```

This might be a bug in the current implementation, but if so, I don't think the fix should be separate from my patch.

nyurik:
@Algunenano I added a few more tests for NYC z10 tile.  The dark green is the tile with 0 margin, the other margins are: 0.1, 0.5, 2, -0.3 (red)
image

Algunenano:
I've created an issue to trac the imprecision in https://trac.osgeo.org/postgis/ticket/4600, it comes from using a 32b float to serialize bboxes.

I think this looks good now. There are 2 things that would improve this PR. If you don't do it yourself I'll add them when pushing the new feature:

- Change the ST_AsMVT documentation example to include this buffer using it correctly:
```
WITH mvtgeom AS
(
  SELECT ST_AsMVTGeom(geom, ST_TileEnvelope(12,513,412), extent => 4096, buffer => 256) AS geom, name, description
  FROM points_of_interest
  WHERE geom && ST_TileEnvelope(12,513,412, margin => (256 / 4096))
)
SELECT ST_AsMVT(mvtgeom.*)
FROM mvtgeom;
```
I might be using the parameters incorrectly, so please confirm them. Also ex

- Drop the old signature. This is an internal requirement so we don't leave old functions dangling, which is an issue with repeated migrations. You can do it with this:
```
diff --git a/postgis/postgis_before_upgrade.sql b/postgis/postgis_before_upgrade.sql
index c709b05d6..6e585c995 100644
--- a/postgis/postgis_before_upgrade.sql
+++ b/postgis/postgis_before_upgrade.sql
@@ -187,6 +187,13 @@ SELECT _postgis_drop_function_if_needed
     'GeometryA geometry, GeometryB geometry'
     );
 
+-- FUNCTION st_tileenvelope added a new default argument in 3.1
+SELECT _postgis_drop_function_if_needed
+    (
+    'st_tileenvelope',
+    'zoom integer, x integer, y integer, bounds geometry DEFAULT ''0102000020110F00000200000052107C45F81B73C152107C45F81B73C152107C45F81B734152107C45F81B7341''::geometry'
+    );
+
 
 -- FUNCTION st_askml changed to add defaults in 3.0 / r17357
 -- These signatures were superseeded
```

The binary geometry is needed as detailed there, which might give us some issues depending on the PROJ release and precision, so we might have to keep the old signature in place anyway :sweat_smile: . But don't worry, we'll readd it if necessary once we start releasing alphas / betas / etc.


To trac the merge I've created a trac issue in https://trac.osgeo.org/postgis/ticket/4601. Please let us know how would you like to be attributed (name / alias / whatever).

nyurik:
Thanks again @Algunenano , updated docs and the drop func script

nyurik:
P.S. Full name creds is fine, i.e. `Yuri Astrakhan (https://github.com/nyurik)`

Algunenano:
Pushed as e18a0b84dfa0995f3f78d52227aceb3b3e494a52. Thanks a lot @nyurik

PR #515, opened 2019-11-14
Fix postgis_proc_set_search_path.pl to handle complex defaults quotes

nyurik:
As discovered in https://github.com/postgis/postgis/pull/514,
an argument default with parenthesis instide the string breaks
regression sql generator (evil regexes).  Fixed with more evil regexes:
now it has a special handling for a '...' -- ignoring all chars inside.

Escaping is not handled (doesn't seem like PostGIS has any of them)

Example:

The script can now convert this:

CREATE OR REPLACE FUNCTION ST_TileEnvelope(zoom integer, x integer, y integer, margin float8, bounds geometry DEFAULT 'SRID=3857;LINESTRING(-20037508.342789 -20037508.342789, 20037508.342789 20037508.342789)'::geometry)

into this:

ST_TileEnvelope(zoom integer, x integer, y integer, bounds geometry)

### DIFF
--- a/utils/postgis_proc_set_search_path.pl
+++ b/utils/postgis_proc_set_search_path.pl
@@ -93,14 +93,14 @@ open( INPUT, $sql_file ) || die "Couldn't open file: $sql_file\n";
 while(<INPUT>)
 {
 
-	if ( /^create or replace function([^\)]+)([\)]{0,1})/i )
+	if ( /^create or replace function([^(]+([^)']+'[^']+')*[^)]+)([)]{0,1})/i )
 	{
 		my $funchead = $1; # contains function header except the end )
 		my $endhead = 0;
-		my $endfunchead = $2;
+		my $endfunchead = $3;
 		my $search_path_safe = -1; # we can put a search path on it without disrupting spatial index use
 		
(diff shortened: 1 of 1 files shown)


robe2:
Hmm we were thinking of getting rid of postgis_proc_set_search_path.pl.  It shouldn't be used for anything anymore -- you see it used somewhere.  That was put in 2.2/2.3 days before we realized setting search path on functions kills performance.

nyurik:
Regina, I'm all for making the build process more streamlined, but this is orthogonal to the bug at hand. At the moment, any change to `ST_TileEnvelope()` signature would cause this bug to surface - that's why #514 wouldn't build until I fixed this issue and rebased on it.

Algunenano:
> Hmm we were thinking of getting rid of postgis_proc_set_search_path.pl. It shouldn't be used for anything anymore -- you see it used somewhere. That was put in 2.2/2.3 days before we realized setting search path on functions kills performance

I've created an issue to drop it: https://trac.osgeo.org/postgis/ticket/4579 I'll try to do it during the day so this PR will no longer be necessary.

Algunenano:
postgis_proc_set_search_path.pl removed

PR #516, opened 2019-11-16
Silence yacc warnings

Algunenano:
I was getting tired of them.

Fixes https://trac.osgeo.org/postgis/ticket/4473 but it

### DIFF
--- a/liblwgeom/lwin_wkt_parse.h
+++ b/liblwgeom/lwin_wkt_parse.h
@@ -1,4 +1,4 @@
-/* A Bison parser, made by GNU Bison 3.4.  */
+/* A Bison parser, made by GNU Bison 3.4.2.  */
 
 /* Bison interface for Yacc-like parsers in C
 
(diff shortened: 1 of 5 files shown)


liblwgeom/lwin_wkt_parse.h
```diff
@@ -1,4 +1,4 @@
-/* A Bison parser, made by GNU Bison 3.4.  */
+/* A Bison parser, made by GNU Bison 3.4.2.  */
```
Komzpa: Can we do something about not committing this file into repo? It looks completely generated, and seems to be going to be built anyway during build process.

strk: Unexpectedly I've been asking for this for a long time now. Contrary argument has always been: not everyone can easily get a parser generator nowadays ... Please please let's get rid of it now!

Algunenano: I'd be ok with it. We just need to make sure it can also be generated with the tarball releases.

strk:
> I'd be ok with it. We just need to make sure it can also be generated with the tarball releases.

Yes, make_dist.sh would need a tweak to fix that.
The "distclean" rule should be already NOT removing that file.

Algunenano:
I've created a different issue to handle the file removal (https://trac.osgeo.org/postgis/ticket/4584)

PR #527, opened 2020-01-03
pgsql2shp quiet mode

kthy:
Fixes [#2972](https://trac.osgeo.org/postgis/ticket/2972).


kthy:
Fixed the formatting now. If you know a good C linter for VS Code…

Algunenano:
> Fixed the formatting now. If you know a good C linter for VS Code…

I use clang-format. I have a git pre-hook that checks it before commit (and rejects it unless it's fixed or forced with `git commit --no-verify`), and you can call `git clang-format` to fix your stashed changes. I don't have to pre-hook on hand right now, but I can share it with you on Monday if you want it.

Algunenano:
Pushed, thanks a lot!

My pre-commit hook: https://raw.githubusercontent.com/Algunenano/myScripts/master/pre-commit (place it as `.git/hooks/pre-commit`). Keep in mind that it might be slow for big diffs, so you might want to ignore it (`--no-verify`) if it takes too long.

PR #528, opened 2020-01-03
Fix several bugs around index support functions

Algunenano:
Fix for https://trac.osgeo.org/postgis/ticket/4608

- ST_Covers and ST_CoveredBy in geography were using the wrong strategy (which didn't exist).
- st_containsproperly in geometry was using the wrong strategy.
- st_3dintersects was duplicated so I removed the bad one.


pramsey:
Cute. Glad someone else has now read this code. +1

PR #531, opened 2020-01-08
Add optional value parameter(s) to the Force3D*, Force4D functions

kthy:
Fixes [#3057](https://trac.osgeo.org/postgis/ticket/3057).

Second attempt, previously discussed in PR #525. Compared to that I've made changes as requested by @Algunenano and @robe2.

Includes documentation update, CUnit tests and regression tests.


kthy:
I'll do it and push changes tonight. My own preference is for the first method (warning like in #530), so unless you have a strong preference for the second, that's what I'll do. :wink:

kthy:
I didn't completely like either option, so I ended up choosing the third way. If you don't like it, please fix and push. :smile:

Algunenano:
Pushed. Thanks a lot and sorry for the delay.

PR #535, opened 2020-01-16
Remove potential double-include of liblwgeom.h

pramsey:


Algunenano:
Did this come from any kind of error? `liblwgeom.h`, and hopefully any other header, has a guard that prevents any re-definition when they are included twice.

pramsey:
Yes, for some reason I could not build the doc generator.c utility until I got rid of the liblwgeom header, and then it worked. Everything *else* worked, but not the doc generator.c utility.

PR #537, opened 2020-01-17
[WIP] Speed up geojson output

Algunenano:
Trac issue: https://trac.osgeo.org/postgis/ticket/4615

Printing the doubles directly into the output buffer doubles the tps with big inputs:

Before 02e62ee:
```
number of clients: 1
number of threads: 1
duration: 30 s
number of transactions actually processed: 31
latency average = 975.931 ms
tps = 1.024663 (including connections establishing)
tps = 1.024788 (excluding connections establishing)
```

After 02e62ee:
```
number of clients: 1
number of threads: 1
duration: 30 s
number of transactions actually processed: 58
latency average = 518.217 ms
tps = 1.929693 (including connections establishing)
tps = 1.929935 (excluding connections establishing)
```

After the introduction of Ryu (https://trac.osgeo.org/postgis/ticket/4543) there are new rough corners, so I plan to keep polishing for a while before merging.

### DIFF
--- a/liblwgeom/lwout_geojson.c
+++ b/liblwgeom/lwout_geojson.c
@@ -722,49 +722,56 @@ asgeojson_geom_buf(const LWGEOM *geom, char *output, GBOX *bbox, int precision)
 static size_t
 pointArray_to_geojson(POINTARRAY *pa, char *output, int precision)
 {
-	uint32_t i;
-	char *ptr;
-	char x[OUT_DOUBLE_BUFFER_SIZE];
-	char y[OUT_DOUBLE_BUFFER_SIZE];
-	char z[OUT_DOUBLE_BUFFER_SIZE];
+	char *ptr = output;
 
 	assert ( precision <= OUT_MAX_DOUBLE_PRECISION );
-	ptr = output;
 
-	/* TODO: rewrite this loop to be simpler and possibly quicker */
 	if (!FLAGS_GET_Z(pa->flags))
 	{
-		for (i=0; i<pa->npoints; i++)
+		for (uint32_t i = 0; i < pa->npoints; i++)
 		{
-			const POINT2D *pt;
-			pt = getPoint2d_cp(pa, i);
+			if (i)
+			{
+				*ptr = ',';
+				ptr++;
+			}
+			const POINT2D *pt = getPoint2d_cp(pa, i);
 
-			lwprint_double(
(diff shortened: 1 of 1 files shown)


Komzpa:
:heart:

Algunenano:
Some notes about other things I've tested:
- I've removed the need to calculate the length of the final string (aka, use `cstring_to_text_with_len` instead of `cstring_to_text`) by passing the size to `lwgeom_to_geojson` as an output value: Initially I introduced a bug and the size returned was the max calculated size, not the actual output size which made 6% slower and left me surprised; after fixing it: 1% faster. Not worth the trouble.

- As mentioned in https://trac.osgeo.org/postgis/ticket/4623, I've tried to create a new liblwgeom type that contains both the size and the data, that is, a `varlena` compatible type so there is no need to transform the data from liblwgeom to Postgresql: 7% improvement, might be worth the trouble even if it's just for geojson (but I expect the impact on other functions to be much higher since the time spent on the function bodies is smaller, e.g St_AsBinary): https://github.com/Algunenano/postgis/commit/6f86f18335694d97e9c48ceaa93073ae82977293


After the previous change (the varlena type) the current process with big geometries show an ussage like this:
```
+   98.53%     0.01%             2  postgres  postgis-3.so            [.] LWGEOM_asGeoJson                                                                                                                                                                  ▒
+   94.78%     0.02%            45  postgres  postgis-3.so            [.] lwgeom_to_geojson                                                                                                                                                                 ▒
+   94.63%     0.04%            65  postgres  postgis-3.so            [.] asgeojson_multipolygon_buf                                                                                                                                                        ▒
+   94.41%     4.05%          4720  postgres  postgis-3.so            [.] pointArray_to_geojson                                                                                                                                                             ▒
+   90.51%     7.85%          8992  postgres  postgis-3.so            [.] lwprint_double                                                                                                                                                                    ▒
+   82.84%    80.38%         90489  postgres  postgis-3.so            [.] d2fixed_buffered_n
```

So technically, between the 98.53% for LWGEOM_asGeoJson and the time spent printing doubles (which should be the high as possible), that is 82.84%, there is only 16% of time being spent, which is low. I might still have a look to see if anything can be improved, but I might end up pushing only this commit here and the varlena change as part of a different PR implementing it for other functions.

pramsey:
Great work @Algunenano 👍

PR #539, opened 2020-01-24
lwprint: Avoid using snprintf

Algunenano:
It can be simplified with simple pointer writes

Looking at the disassembly I noticed the call to `snprintf` which we don't really want to keep.

Related to https://trac.osgeo.org/postgis/ticket/4543


pramsey:
👍

PR #542, opened 2020-01-29
Revert bad search and replace

fbonzon:
From 97ef9ac9d523b4b96328f78f5d9d51189f9476a4


Algunenano:
Pushed. Thanks for contributing!

PR #544, opened 2020-01-31
ST_Hexagon: Improve precision

Algunenano:
Will depend on architecture, but on x86 long doubles should be
80bit precision (vs 64b). Might not work on Windows

Trac: https://trac.osgeo.org/postgis/ticket/4627

### DIFF
--- a/postgis/lwgeom_generate_grid.c
+++ b/postgis/lwgeom_generate_grid.c
@@ -68,24 +68,20 @@ GeometryGridState;
 
 /* Build origin hexagon centered around origin point */
 static const double hex_x[] = {-1.0, -0.5,  0.5, 1.0, 0.5, -0.5, -1.0};
-static const double hex_y[] = { 0.0, -1*H, -1*H, 0.0,   H,    H,  0.0};
+static const double hex_y[] = {0.0, -0.5, -0.5, 0.0, 0.5, 0.5, 0.0};
 
 static LWGEOM *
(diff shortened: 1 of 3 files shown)


Algunenano:
Come to think of it, it might be better to change the C api / calculations to do sh + h*(n - 1) and sh + h*(n). That way the result should always be consistent with previous tiles.

I'll try to have a look next week

dbaston:
Yes, there should only be one way to calculate each coordinate. Maybe need to check the square grid code for the same issue.

Algunenano:
I think this should be stable across tiles. The square tiles are calculated as `origin + (size * (cell + 1))` so they should already be consistent.

Algunenano:
Using these changes, I've made a brute test and it looks ok:
```
WITH geoms AS
(
    SELECT * FROM  (SELECT (ST_HexagonGrid(5, '01030000000100000005000000fdffffffff7f66c0fcffffffff7f56c0fdffffffff7f66c07cc0e71a95e8544000000000008066407cc0e71a95e854400000000000806640fcffffffff7f56c0fdffffffff7f66c0fcffffffff7f56c0')).*) sq
)
SELECT '316',
  count(*) as c,
   ST_GeometryType(ST_Intersection(j0.geom, j1.geom)) t
FROM geoms j0, geoms j1 where NOT ST_Equals(j0.geom, j1.geom ) AND  ST_Intersects(j0.geom, j1.geom) GROUP BY t;
```

```316|5896|ST_LineString```

So all the intersections in the generated tileset are lines, which is what we would expect.

PR #545, opened 2020-02-21
Simplify zigzag

Algunenano:
This only has an impact on GCC, as clang optimizes it already when using -O3: https://godbolt.org/z/XWYBki


Algunenano:
I'm not sure why clang complains about undefined behaviour under clang 8 but not in clang 9 and I don't care enough to solve it now.

PR #548, opened 2020-03-12
ST_Simplify: Don't copy if nothing is removed

Algunenano:
Corner case optimization: When nothing has been removed, there is no need to move points around.

Trac: https://trac.osgeo.org/postgis/ticket/4651

### DIFF
--- a/liblwgeom/ptarray.c
+++ b/liblwgeom/ptarray.c
@@ -1692,7 +1692,7 @@ ptarray_simplify_in_place(POINTARRAY *pa, double tolerance, uint32_t minpts)
 		       pa->serialized_pointlist + pt_size * (pa->npoints - 1),
 		       pt_size);
 	}
-	else
+	else if (pa->npoints != keptn) /* We don't need to move any points if we are keeping them all */
 	{
 		for (uint32_t i = 1; i < pa->npoints; i++)
 		{


liblwgeom/ptarray.c
```diff
@@ -1692,7 +1692,7 @@ ptarray_simplify_in_place(POINTARRAY *pa, double tolerance, uint32_t minpts)
 		       pa->serialized_pointlist + pt_size * (pa->npoints - 1),
 		       pt_size);
 	}
-	else
+	else if (pa->npoints != keptn)
```
Komzpa: please leave a comment.

Algunenano:
Comparison (big polygons):
```sql
explain analyze Select ST_Simplify(the_geom_webmercator, 0.000000001) from benchmark_4c7214d90a79aa6760367a084a4d4a2f61fbe1c6cc4f7f9e76020;
```

Pre-patch:
```
number of transactions actually processed: 162
latency average = 185.678 ms
```

After-patch:
```
number of transactions actually processed: 169
latency average = 177.686 ms
```

liblwgeom/ptarray.c
Algunenano: You are right. Will do

PR #552, opened 2020-04-07
Master toast cache

pramsey:
Put a simple cache in front of the de-toasting in functions that use an in-memory cached index scheme. This does not replace in in-memory scheme, so there's some inefficiency, but it makes the patch much simpler than a complete re-write to join the two schemes together while not sacrificing performance in some common cases, like point-in-polygon.

### DIFF
--- a/libpgcommon/lwgeom_cache.c
+++ b/libpgcommon/lwgeom_cache.c
@@ -13,7 +13,12 @@
 #include "postgres.h"
 #include "fmgr.h"
 #include "access/tuptoaster.h"
+
+#if POSTGIS_PGSQL_VERSION >= 120
 #include "utils/hashutils.h"
+#else
+#include "access/hash.h"
+#endif
 
 #include "../postgis_config.h"
 #include "lwgeom_cache.h"
--- a/libpgcommon/lwgeom_cache.c
+++ b/libpgcommon/lwgeom_cache.c
@@ -14,12 +14,6 @@
 #include "fmgr.h"
 #include "access/tuptoaster.h"
 
-#if POSTGIS_PGSQL_VERSION >= 120
-#include "utils/hashutils.h"
-#else
-#include "access/hash.h"
-#endif
-
 #include "../postgis_config.h"
 #include "lwgeom_cache.h"
 
@@ -262,17 +256,6 @@ ToastCacheGet(FunctionCallInfo fcinfo)
 	return cache;
 }
 
-inline static uint64_t
-ToastCacheHashVarlena(const struct varlena *attr)
-{
-	struct varatt_external ve;
-	VARATT_EXTERNAL_GET_POINTER(ve, attr);
-	uint64_t a = DatumGetUInt64(hash_uint32(ve.va_valueid));
-	uint64_t b = DatumGetUInt64(hash_uint32(ve.va_toastrelid));
-	/* Merge hashes */
-	return a ^ (b + UINT64CONST(0x49a0f4dd15e5a8e3) + (a << 54) + (a >> 7));
-}
-
 GSERIALIZED*
 ToastCacheGetGeometry(FunctionCallInfo fcinfo, uint32_t argnum)
 {
@@ -287,10 +270,14 @@ ToastCacheGetGeometry(FunctionCallInfo fcinfo, uint32_t argnum)
 	if (!VARATT_IS_EXTERNAL_ONDISK(attr))
 		return (GSERIALIZED*)PG_DETOAST_DATUM(datum);
 
-	/* Argument is TOASTed, check if we already have a copy handy. */
-	uint64_t hash = ToastCacheHashVarlena(attr);
+	/* Retrieve the unique keys */
+	struct varatt_external ve;
+	VARATT_EXTERNAL_GET_POINTER(ve, attr);
+	Oid valueid = ve.va_valueid;
+	Oid toastrelid = ve.va_toastrelid;
+
 	/* We've seen this object before? */
-	if (arg->hash == hash)
+	if (arg->valueid == valueid && arg->toastrelid == toastrelid)
 	{
 		if (arg->geom)
 			return arg->geom;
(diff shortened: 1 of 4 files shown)


libpgcommon/lwgeom_cache.c
```diff
@@ -236,4 +245,70 @@ GetGeomCache(FunctionCallInfo fcinfo,
...
+{
+	struct varatt_external ve;
+	VARATT_EXTERNAL_GET_POINTER(ve, attr);
+	uint64_t a = DatumGetUInt64(hash_uint32(ve.va_valueid));
+	uint64_t b = DatumGetUInt64(hash_uint32(ve.va_toastrelid));
+	/* Merge hashes */
+	return a ^ (b + UINT64CONST(0x49a0f4dd15e5a8e3) + (a << 54) + (a >> 7));
+}
 
+GSERIALIZED*
```
Algunenano, line 276: I assume this is should be a `const GSERIALIZED*` as in `thou shalt not modify this nor the geometry that you build on top of it`, right?

pramsey: Maybe so, but doing so cascade down into piles of other legacy code, so if we want to start making gserialized const over the whole code base that's a different task.

pramsey:
Please do!

libpgcommon/lwgeom_cache.c
```diff
@@ -236,4 +239,64 @@ GetGeomCache(FunctionCallInfo fcinfo,
...
+{
+	Assert(argnum < ToastCacheSize);
+	ToastCache* cache = ToastCacheGet(fcinfo);
+	ToastCacheArgument* arg = &(cache->arg[argnum]);
+
+	Datum datum = PG_GETARG_DATUM(argnum);
+	struct varlena *attr = (struct varlena *) DatumGetPointer(datum);
+
+	/* Argument is not TOASTED, access as normal. */
+	if (!VARATT_IS_EXTERNAL_ONDISK(attr))
```
Algunenano, line 270: PG's `pg_detoast_datum` will do work with anything coming from disk (`VARATT_IS_EXTERNAL_ONDISK`), but also indirect access (`VARATT_IS_EXTERNAL_INDIRECT`), expanded (`VARATT_IS_EXTERNAL_EXPANDED`) and so on; so I'm curious about why just limit the cache to data coming from disk and not to all extended objects (matching detoast).

libpgcommon/lwgeom_cache.c
```diff
@@ -236,4 +239,64 @@ GetGeomCache(FunctionCallInfo fcinfo,
...
+		arg->geom = (GSERIALIZED*)PG_DETOAST_DATUM_COPY(datum);
+		MemoryContextSwitchTo(old_context);
+		return arg->geom;
+	}
+	/* New object, clear our old copies and see if it */
+	/* shows up a second time before taking a copy */
+	else
+	{
+		if (arg->geom)
+			pfree(arg->geom);
```
Algunenano, line 296: Is it necessary to switch context to call free here? I'm not sure if pfree works across contexts.

libpgcommon/lwgeom_cache.c
```diff
@@ -236,4 +239,64 @@ GetGeomCache(FunctionCallInfo fcinfo,
...
+		if (arg->geom)
+			return arg->geom;
+
+		/* Take a copy into the upper context */
+		MemoryContext old_context = MemoryContextSwitchTo(FIContext(fcinfo));
+		arg->geom = (GSERIALIZED*)PG_DETOAST_DATUM_COPY(datum);
+		MemoryContextSwitchTo(old_context);
+		return arg->geom;
+	}
+	/* New object, clear our old copies and see if it */
```
Algunenano, line 291: Is it worth avoiding the extra memcpy in the first hit for the case where all values are new?

libpgcommon/lwgeom_cache.c
pramsey: This is following the Word of Tom, and probably he could elabourate, but the key is that when toasted on disk, an object has a relid and a varid that are unique. Before then it'll have a pointer into a memory buffer that *might* change underneath it. So only things that have been stored on disk can be identified uniquely without having to first read in the contents of the object. Fortunately, that's most of the things we care about.

libpgcommon/lwgeom_cache.c
pramsey: Nope, not necessary (to my knowledge).

libpgcommon/lwgeom_cache.c
pramsey: I go back and forth on that, but I think that (a) in the case where the objects are repeating, we'll get more than 2-3 repetitions and the extra cost will wash out and (b) in the case where the objects are not repeating, we don't want to incur the extra copy and free cost of putting a copy into the upper context.

libpgcommon/lwgeom_cache.h
```diff
@@ -110,4 +111,24 @@ GeomCache *GetGeomCache(FunctionCallInfo fcinfo,
...
+typedef struct
+{
+	Oid valueid;
+	Oid toastrelid;
+	GSERIALIZED *geom;
+} ToastCacheArgument;
+
+typedef struct
+{
+	int type;
```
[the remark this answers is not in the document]
pramsey: Yeah, I never set up any checks that the type you where receiving was the same as the type that was stored... I kind of want to leave it lying around :)

libpgcommon/lwgeom_cache.c
Algunenano: Can you leave a comment about it please? I'm bound to think about this next time I look at the code a few months from now.

libpgcommon/lwgeom_cache.c
Algunenano: You are right, pfree does look for the pointer context before freeing it, so it's all good.

pramsey:
Ran a few more performance tests, with a data set with about 10% toasted values (2x faster) and a data set with 0% toasted values (same performance), so still feeling good.

pramsey:
Merged at 8b548a469

PR #554, opened 2020-04-10
Enforce outputting the requested number of decimal digits

Algunenano:


liblwgeom/cunit/cu_clip_by_rect.c
```diff
@@ -29,32 +29,28 @@ static void test_lwgeom_clip_by_rect(void)
...
+	ASSERT_STRING_EQUAL(wkt, tmp)
 	lwfree(tmp); lwgeom_free(out); lwgeom_free(in);
 
 	wkt = "MULTIPOINT(0 0, 6 6, 7 5)";
 	in = lwgeom_from_wkt(wkt, LW_PARSER_CHECK_NONE);
 	out = lwgeom_clip_by_rect(in, 5, 0, 10, 10);
 	tmp = lwgeom_to_ewkt(out);
-	/* printf("%s\n", tmp); */
-	CU_ASSERT_STRING_EQUAL("MULTIPOINT(6 6,7 5)", tmp)
+	ASSERT_STRING_EQUAL("MULTIPOINT(6 6,7 5)", tmp)
```
dr-jts, line 53: Why this change?  It seems like there are a lot of these, and they are not related to the output precision issue?

Algunenano: It's true. Due to how I ended up dividing the PRs I squashed multiple commits into one and I lost quite a bit of information that I should recover before merging.

`CU_ASSERT_STRING_EQUAL` compares strings and returns an error, but it doesn't output the values (so you are blind as you what the difference was). OTOH, we also have `ASSERT_STRING_EQUAL` which does the same but outputs the strings when an error happens. Since there were multiple broken tests that didn't output the diffs, I decided to replace all calls to the one with the output and removed any manual printf that some tests had to do that.

dr-jts: Makes sense.  But does make the diff seem larger than it actually is.

dr-jts:
It seems like this PR causes a lot more decimal places to be output than were previously.  Is this because the stated "default max decimal places = 15" is being interpreted literally? If so, I don't think this is the semantic that was intended, even though the manual appears to read that way.  It makes more sense if the semantic is understood as "if no precision is specified, a maximum of 15 *significant digits* is output".

Algunenano:
> It seems like this PR causes a lot more decimal places to be output than were previously. Is this because the stated "default max decimal places = 15" is being interpreted literally? If so, I don't think this is the semantic that was intended, even though the manual appears to read that way. It makes more sense if the semantic is understood as "if no precision is specified, a maximum of 15 significant digits is output".

The thing is, we can't have a different meaning for the value depending on whether is the default value or not (mainly because we don't know if the default value was used or not since Postgres handles that). So the meaning of 15 needs to be stable, either it means 15 decimal digits or it means 15 significant digits, but I don't see how it could mean both.

liblwgeom/cunit/cu_clip_by_rect.c
Algunenano: You are right. I'll try to break down the PR into different commits or even remove that change and leave only the digit changes so it's more clear what's truly changing in the PR.

dr-jts:
> The thing is, we can't have a different meaning for the value depending on whether is the default value or not (mainly because we don't know if the default value was used or not since Postgres handles that). So the meaning of 15 needs to be stable, either it means 15 decimal digits or it means 15 significant digits, but I don't see how it could mean both.

Agreed.  The way to handle this is to change the default value to a "sentry value" (something non-meaningful) that will indicate that no value was provided by the user.  We could use say 999 or -999 for this.

The manual would have to change, of course. It should **no longer say** that there is a default value, but will specify that if the precision value is omitted then numbers are output with their "inherent" floating-point precision.

Algunenano:
> Agreed. The way to handle this is to change the default value to a "sentry value" (something non-meaningful) that will indicate that no value was provided by the user. We could use say 999 or -999 for this.

Uhmm, that could actually be an option. Do you think that upon passing the sentry value we should do the old behaviour (print 20 digits tops, no trailing zeros) or would it be preferable to output `as little digits as possible guaranteeing round trip safety` (aka floating point precision)?

liblwgeom/cunit/cu_clip_by_rect.c
Algunenano: I'm going to force-push removing these changes and leaving only the minimum changes reflecting the `return always N amount of decimal digits` change.

dr-jts:
> > Agreed. The way to handle this is to change the default value to a "sentry value" (something non-meaningful) that will indicate that no value was provided by the user. We could use say 999 or -999 for this.
> 
> Uhmm, that could actually be an option. Do you think that upon passing the sentry value we should do the old behaviour (print 20 digits tops, no trailing zeros) or would it be preferable to output `as little digits as possible guaranteeing round trip safety` (aka floating point precision)?

In the interests of backwards compatibility, for now, I think we should stick to the old behaviour.   What we don't want is a lot of extra digits showing up unexpectedly.

Algunenano:
> In the interests of backwards compatibility, for now, I think we should stick to the old behaviour. What we don't want is a lot of extra digits showing up unexpectedly.

At that point, there is not a big difference to changing the definition of precision to number of significant digits and keeping the old implementation (which is what the other PR does, #555).

dr-jts:
> > In the interests of backwards compatibility, for now, I think we should stick to the old behaviour. What we don't want is a lot of extra digits showing up unexpectedly.
> 
> At that point, there is not a big difference to changing the definition of precision to number of significant digits and keeping the old implementation (which is what the other PR does, #555).

I don't see how this is the case.  This proposal does not provide any mechanism to specify the number of significant digits.  

It provides:
* if the precision value is *omitted*, then the number is output with the minimum number of digits required to fully represent the binary value 
* if the precision is specified, then that number of *decimal* places is output

Algunenano:
> if the precision value is omitted, then the number is output with the minimum number of digits required to fully represent the binary value

I understood from your previous comment that we wanted to stick with the old behaviour for the default (not declaring a precision), but that isn't the old behaviour.
The old default was essentially to print 15 total digits; and this also applied to non default precision; for example if you request 16 precision in ST_AsText precision you get 16 digits: 1234567890.0123456789 gives you `1234567890.012346`.

So changing both `if precision is omitted` and `if precision is specified` is a breaking change in both cases. The first gives a completely new output (for the better in my opinion, since it only prints as many digits as needed, not as many as requested), and the second one too (for the better too, since it does what the documentation says it does).

That's why, if we don't want breaking changes, changing the documentation to say it will always print N significant digits is the simplest way to go.

Also, one important distinction: I've been talking about significant digits but they are really `not zero digits`. Those extra digits many be unnecessary. Ideally we would always use the shortest up to N digits, but we've never had that.
An example using ryu's own example:
```
Select ST_AsText(ST_MakePoint(0, 0.300000011920928955078125), 16);
         st_astext          
----------------------------
 POINT(0 0.300000011920929)
```

It doesn't make any sense to print that many digits. Printing `0.3` leads back to the same double, so all the digits after the 3 are unnecessary (but significant*).
* Using my made up definition of significant and necessary.

Maybe I could try to take ryu and use the `shortest` function instead of the printf replacement (which I did to keep compatibility gaining 5x speed) instead?

So what we could do is to print the shortest representation, then if the use happened to request a max amount of decimal digits, go back and trim that carefully. If that's something you consider a good alternative I'll investigate what we need to adapt from ryu's code to get it (in my exploratory tests I was always getting some kind of exponential notation like 3E1 with that approach so I discarded it without much investigation).

pramsey:
> On Apr 17, 2020, at 10:40 AM, Raúl Marín <notifications@github.com> wrote:
> 
> Also, one important distinction: I've been talking about significant digits but they are really not zero digits. Those extra digits many be unnecessary. Ideally we would always use the shortest up to N digits, but we've never had that.
> An example using ryu's own example:
> 
> Select ST_AsText(ST_MakePoint(0, 0.300000011920928955078125), 16);
>          st_astext          
> ----------------------------
>  POINT(0 0.300000011920929)
> 
> It doesn't make any sense to print that many digits. Printing 0.3 leads back to the same double, so all the digits after the 3 are unnecessary (but significant*).
> 

This seems fruitful. I didn't realize such a magic mode existed. So ryu will produce a "reasonable" number of digits that still allow a round-trip? That seems like the best of all possible worlds. Seems like a huge win for the default.

P

dr-jts:
> So changing both `if precision is omitted` and `if precision is specified` is a breaking change in both cases. The first gives a completely new output (for the better in my opinion, since it only prints as many digits as needed, not as many as requested), and the second one too (for the better too, since it does what the documentation says it does).
> 
> That's why, if we don't want breaking changes, changing the documentation to say it will always print N significant digits is the simplest way to go.

So the documentation and the current behaviour are inconsistent?   The docs say "max decimal digits", but the behaviour is "max total digits"?  That is unfortunate IMO...  As I previously mentioned, specifying the max decimal digits is more useful than specifying max total digits (since the user doesn't really know the magnitude of the numbers he is outputting, and the magnitude can vary across a given dataset.  Also, I think people want to be able to say "output with millimetre precision" and have that work).

What are the implications of using the semantics I stated:

* if the precision value is omitted, then the number is output with the minimum number of digits required to fully represent the binary value
* if the precision is specified, then that number of decimal places is output

Algunenano:
Closing this. I'll open a new one once I have finished implementing and testing the shortest (up to precision decimal digits) output.

PR #559, opened 2020-05-01
Share gserialized cache between cache types

Algunenano:
This enables sharing pointers between the toast cache and the geom caches (multiple types). This is done to enable the fast path where pointer comparison is enough to find the geometry in `GetGeomCache`, which avoids the costly memcmp call.

Using @pramsey infamous dataset for a quick comparison:
```
explain analyze SELECT count(*), c.name    FROM countries c   JOIN places p   ON ST_Intersects(c.geom, p.geom)   GROUP BY c.name;
-- Before latency average = 626.083 ms
-- After latency average = 347.727 ms
-- Perf: 1.80x
```

Almost a 2x.

This needs a more in depth review (it's my third big PR today) to make sure memory bookkeeping is correct.

### DIFF
--- a/libpgcommon/lwgeom_cache.c
+++ b/libpgcommon/lwgeom_cache.c
@@ -160,25 +160,20 @@ GetGeomCache(FunctionCallInfo fcinfo,
 	}
 
 	/* Cache hit on the first argument */
-	if ( g1 &&
-	     cache->argnum != 2 && (
-	     (g1 == cache->geom1) ||
-	     (cache->geom1_size == VARSIZE(g1) && memcmp(cache->geom1, g1, cache->geom1_size) == 0)
-	     ))
+	if (g1 && cache->argnum != 2 &&
+	    ((g1 == cache->shared_geom1->geom) ||
+	     (cache->geom1_size == VARSIZE(g1) && memcmp(cache->shared_geom1->geom, g1, cache->geom1_size) == 0)))
 	{
 		cache_hit = 1;
-		geom = cache->geom1;
-
+		geom = cache->shared_geom1->geom;
 	}
 	/* Cache hit on second argument */
-	else if ( g2 &&
-	          cache->argnum != 1 && (
-	          (g2 == cache->geom2) ||
-	          (cache->geom2_size == VARSIZE(g2) && memcmp(cache->geom2, g2, cache->geom2_size) == 0)
-	          ))
+	else if (g2 && cache->argnum != 1 &&
+		 ((g2 == cache->shared_geom2->geom) ||
+		  (cache->geom2_size == VARSIZE(g2) && memcmp(cache->shared_geom2->geom, g2, cache->geom2_size) == 0)))
 	{
 		cache_hit = 2;
-		geom = cache->geom2;
+		geom = cache->shared_geom2->geom;
 	}
 	/* No cache hit. If we have a tree, free it. */
 	else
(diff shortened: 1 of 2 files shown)


pramsey:
I'm kind of concerned about the prior PR that has moved the cache into CurTransactionContext and started using a process-wide global to hold the generic cache. This will certainly cause collisions in statements that use multiple caching functions at once, and churn in cases where the multiple functions use multiple geometries. 
```
WHERE st_contains(g1, g2) and st_intersectsS(g3, g4)
```
WRT this particular PR, I had a much less invasive scheme in mind. It has the idea of a cached gserialized as well, which is what the toast cache would return, something like 
```
{
uint32_t hash;
GSERIALIZED *geom
}
```
Basically when the toast cache read in a large object it would also calculate the hash. 
The later caches could continue to manage their own copies of the geometry, so no reference counting required. They'd just be able to skip the memcmp as an identity check, since they could use the hash instead.

So still a little copying (once into the toast cache, once into the index cache) but no memcmp'ing for large objects, and the cache implementations remain independent instead of being wired precariously together.

Algunenano:
> This will certainly cause collisions in statements that use multiple caching functions at once, and churn in cases where the multiple functions use multiple geometries.

This case has changed the first row from a 'cache doesn't exist' into a 'cache miss'. Since functions are being executes sequentially whithin the worker, it's impossible for them to conflict with each other, right?

> Basically when the toast cache read in a large object it would also calculate the hash.

It's an interested approach I hadn't considered, but doesn't it require hashing all geometries (toasted or not). I would expect the impact on non toasted geometries to be significant.

pramsey:
Regarding hashing: no, when the toastcache reads in small objects it just passes them along with a zero'ed hash value. Only large objects would get hashed. So the extra costs are just on the large objects.

Algunenano:
Another thought: hashes to int aren't unique, so to guarantee uniqueness you need to confirm that both the hash and the gserialized object are the same. This would mean that you can't remove the memcmp, so no performance gain for the case improves here (equal large geometries).

pramsey:
Your cache is around for the life of the transaction. I expect the executor is going to be working row-by-row, and each row can involve multiple calls of cache-enabled functions, which are now all going to be looking into the **same** cache, the global one at `internal_cache`. Before they were looking at a function-level context in `flinfo->fl_memctx` so they almost never conflicted (I think the exceptions where places where the same function might either call the rtree or the preparegeom code line, due to variable input types, which is why the genericcache construction came to exist at all)

Algunenano:
> Your cache is around for the life of the transaction. I expect the executor is going to be working row-by-row, and each row can involve multiple calls of cache-enabled functions, which are now all going to be looking into the same cache, the global one at internal_cache

I was under the impression that the executor always worked function by function, that is it executes functionA for all rows, then functionB for all rows and so on. If that's not the case (and it executes functionA, functionB... for first row, functionA, functionB for second row and so on) then you are right and it's going to be throublesome. The tests I did locally aparently confirmed my believes, but now I would like to have a confirmation one way or the other.

Komzpa:
> it executes functionA, functionB... for first row, functionA, functionB for second row and so on

Postgres is not columnar db. It has to spit out everything row-by-row tuple-by-tuple. There can be fancy plans for fancy queries where there are two different intermediate tuples and functions are thus executed function-by-function, but that has to be written manually.

Algunenano:
Well, it's likely that I got it wrong then. I'll try to revisit it next week and come up with a few scenarios to see how the cache behaves (to see where it's better per function or better per process). Backing out that change is a one-liner, although getting back the performance in ST_AsGML might require replacing the SQL inline with an extra C function (or some tweaking to it, not sure).

pramsey:
Understood. I'd rather have a slow AsGML (who uses that?!) than what feels somewhat dangerous changes at the cache system. Also, a tighter C bind on AsGML would be nice (death to inlining! :)

pramsey:
Regarding using hashes to proxy for identity checks, how much does using uint64_t hashes make you feel better? uint128_t? 😄

Komzpa:
Hashes exist to quickly prove inequality, not to quickly prove equality. You have to do a full memcmp if you can't compare pointers on equality.

Algunenano:
I agree with @Komzpa. A (good) checksum could be another option, but I'm pretty sure that's going to be much slower than the memcmp call and require quite a lot of code (or another external library).

pramsey:
Well, checksum would only have to be calculated once. It's just a way to see if the copy you have in the toast cache is the same as the copy you have in the index cache. It just avoids having to basically unify the two caches.

Algunenano:
What's exactly your concern here? Is it the geom cache looking into the toast cache? The shared pointers?

Can we find a way to address that concern that doesn't involve a costly approach?

pramsey:
WRT the toast cache and index cache, yes, it's the shared pointers which effectively make them one big-ass cache, and a good deal more complicated to reason about.

WRT the changes in general, the move up into the TransactionContext is also a huge red flag for me.

P

Algunenano:
> WRT the toast cache and index cache, yes, it's the shared pointers which effectively make them one big-ass cache, and a good deal more complicated to reason about.

Would it be better if GetGeomCache used `SHARED_GSERIALIZED` as arguments and I added methods to create it, compare them, add a reference and remove a reference so that it never access the `ToastCache` directly? 

> WRT the changes in general, the move up into the TransactionContext is also a huge red flag for me.

I have to ask you to please stop mixing unrelated changes. That change was done in a different commit / PR that has nothing to do with this proposal, and I've already offerered to revert it to the older behaviour. I don't know what to do when you bring it up here.

Algunenano:
Rebasing this on top of master. I'm planning to do an alternative API with a similar spirit (keeping the counter, but handling it through calls) in the hopes that looks better than the current proposal

Algunenano:
Alternative changes in https://github.com/postgis/postgis/pull/564

Algunenano:
Closing in favor of #564

PR #561, opened 2020-05-07
Inline gml

Algunenano:
Related to https://trac.osgeo.org/postgis/ticket/4674

Always use a C function directly in ST_AsGML(geometry) so the `getSRSbySRID` cache is used.

```
-- 683788 Points (1 vertex on avg)
-- EXPLAIN ANALYZE Select ST_AsGML(the_geom, 8) from benchmark_7773a711c8441d4b494a51fd9feebeac7a9b9c734619398620293;
-- Latency avg before: 8286.371 ms
-- Latency avg with cache: 223.431 ms
-- Perf: 37.08x
```

Follows the same style as st_asgml(geography), which was already doing this. I don't know why we have 2 implementations that are almost equal to be honest.

### DIFF
--- a/postgis/geography_inout.c
+++ b/postgis/geography_inout.c
@@ -36,7 +36,7 @@
 #include "utils/array.h"
 #include "utils/builtins.h"  /* for pg_atoi */
 #include "lib/stringinfo.h"  /* For binary input */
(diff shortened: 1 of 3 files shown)


Algunenano:
Merged

PR #564, opened 2020-05-08
Share gserialized cache between cache types alternative

Algunenano:
Alternative to https://github.com/postgis/postgis/pull/559/

It's cleaner (I think/hope) but requires more changes.

```
explain analyze SELECT count(*), c.name    FROM countries c   JOIN places p   ON ST_Intersects(c.geom, p.geom)   GROUP BY c.name;
-- Before latency average = 627.031 ms
-- After latency average = 316.351 ms
-- Perf: 1.98x
```

It's based on the same principle, ToastCacheGetGeometry will return a SHARED_GSERIALIZED created either in the current context (if not toasted) or in the cache context (if toasted). Then GetGeomCache, will add a reference to that SHARED_GSERIALIZED (which will copy it into the cache context if needed).

Even though it's unused I've also applied the full modifications to ST_DistanceRectTreeCached.

Things look fine in the local environment, but the tests were already broken so not everything is executed.

### DIFF
--- a/libpgcommon/lwgeom_cache.c
+++ b/libpgcommon/lwgeom_cache.c
@@ -292,11 +292,6 @@ ToastCacheGetGeometry(FunctionCallInfo fcinfo, uint32_t argnum)
 	/* We've seen this object before? */
 	if (arg->valueid == valueid && arg->toastrelid == toastrelid)
 	{
-		if (arg->geom)
-			return arg->geom;
-
-		/* Take a copy into the upper context */
-		arg->geom = shared_gserialized_new_cached(fcinfo, datum);
 		return arg->geom;
 	}
 	/* New object, clear our old copies and see if it */
@@ -307,8 +302,8 @@ ToastCacheGetGeometry(FunctionCallInfo fcinfo, uint32_t argnum)
 			shared_gserialized_unref(fcinfo, arg->geom);
 		arg->valueid = valueid;
 		arg->toastrelid = toastrelid;
-		arg->geom = NULL;
-		return shared_gserialized_new_nocache(datum);
+		arg->geom = shared_gserialized_new_cached(fcinfo, datum);
+		return arg->geom;
 	}
 }
 
--- a/libpgcommon/lwgeom_cache.c
+++ b/libpgcommon/lwgeom_cache.c
@@ -133,8 +133,8 @@ GetPROJSRSCache(FunctionCallInfo fcinfo)
 GeomCache *
 GetGeomCache(FunctionCallInfo fcinfo,
 	     const GeomCacheMethods *cache_methods,
-	     const GSERIALIZED *g1,
-	     const GSERIALIZED *g2)
+	     SHARED_GSERIALIZED *g1,
+	     SHARED_GSERIALIZED *g2)
 {
 	GeomCache* cache;
 	int cache_hit = 0;
@@ -147,7 +147,6 @@ GetGeomCache(FunctionCallInfo fcinfo,
 	Assert(entry_number < NUM_CACHE_ENTRIES);
 
 	cache = (GeomCache*)(generic_cache->entry[entry_number]);
-
 	if ( ! cache )
 	{
 		old_context = MemoryContextSwitchTo(PostgisCacheContext(fcinfo));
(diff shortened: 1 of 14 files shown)


libpgcommon/lwgeom_cache.c
```diff
@@ -303,25 +292,18 @@ ToastCacheGetGeometry(FunctionCallInfo fcinfo, uint32_t argnum)
 	/* We've seen this object before? */
 	if (arg->valueid == valueid && arg->toastrelid == toastrelid)
 	{
-		if (arg->geom)
```
pramsey, line 306: The point of only loading on a cache hit (counterintuitively) is so that we only load when we've see the object *twice*. That way we "know" (loosely) that we're in a repeating pattern. Loading on every cache miss can mean taking a copy every time, in that case where we aren't seeing a repeating pattern. I figure the trade off in the coherent case (one extra load at the start) was offset by the incoherent case (every load is a wasted load).

libpgcommon/lwgeom_cache.h
```diff
@@ -35,6 +35,18 @@
 /* Returns the MemoryContext used to store the caches */
 MemoryContext PostgisCacheContext(FunctionCallInfo fcinfo);
 
+typedef struct {
+	uint32_t count;
+	GSERIALIZED *geom;
+} SHARED_GSERIALIZED;
```
pramsey, line 41: Reverse struct entries to avoid padding between `count` and `geom`? (You're the expert on such optimizations though 😄

libpgcommon/lwgeom_cache.h
```diff
@@ -43,10 +55,8 @@ MemoryContext PostgisCacheContext(FunctionCallInfo fcinfo);
 */
 typedef struct {
 	int                         type;
-	GSERIALIZED*                geom1;
-	GSERIALIZED*                geom2;
-	size_t                      geom1_size;
-	size_t                      geom2_size;
+	SHARED_GSERIALIZED *geom1;
+	SHARED_GSERIALIZED *geom2;
 	LWGEOM*                     lwgeom1;
```
pramsey, line 60: Not really your Problem, but I'm pretty sure in one of my rewrites I stripped the `lgeom` entries from this struct as they aren't ever actually used in any of the cache lines.

libpgcommon/shared_gserialized.c
```diff
@@ -0,0 +1,98 @@
...
+ **********************************************************************
+ *
+ * Copyright 2020 Raúl Marín
+ *
+ **********************************************************************/
+
+#include "lwgeom_cache.h"
+
+SHARED_GSERIALIZED *
+shared_gserialized_new_nocache(Datum d)
```
pramsey, line 28: I don't have my head entirely around the lifecycle of this, but is there a place for a `void shared_gserialized_new_nocache(Datum d, SHARED_GSERIALIZED *s)` and passing a statically allocated `SHARED_GSERIALIZED` in? Probably the world's tiniest optimization, given that the next step is copying in a whole `GSERIALIZED`, but it saves the `palloc`

libpgcommon/shared_gserialized.c
```diff
@@ -0,0 +1,98 @@
...
+	s->geom = (GSERIALIZED *)PG_DETOAST_DATUM_COPY(d);
+	MemoryContextSwitchTo(old_context);
+	s->count = 1;
+	return s;
+}
+
+SHARED_GSERIALIZED *
+shared_gserialized_ref(FunctionCallInfo fcinfo, SHARED_GSERIALIZED *ref)
+{
+	if (MemoryContextContains(PostgisCacheContext(fcinfo), ref))
```
pramsey, line 50: Do we have any take on the cost of `MemoryContextContains()`?

Algunenano: Should be constant value. I (when doing this PR) learnt that pointers returned by palloc contain a header that precedes them, so all this does is check the header.

pramsey: 👍

libpgcommon/lwgeom_cache.h
Algunenano: Not really, it should be the other way around although in x86-64 it doesn't matter (it's 16 bytes no matter the order) and since it's not placed in any array or something like that it would really be a micro-tiny-optimization, but a good thing to keep in mind on principle. I'll change it.

libpgcommon/lwgeom_cache.c
Algunenano: The truth is, I did this because with the old behaviour when ToastCacheGetGeometry sees the toasted geometry for the first time it would return a temporal shared shared_gserialized (same as the non toasted value). When this reaches the GetGeomCache, it wants to cache it, so it adds a reference (which in turn will do a copy in the cache context, since it needs it to live). 
The second time the geometry appears, the toastcache will create a "perpetual" shared_gserialized, but when this reaches the getgeocache the geometry is already in cache, so you have 2 "perpetual" shared_gserialized containing different copies of the geometry but they pass the equality check so they won't "merge" (and perform the memcpy everytime).
There are several ways to handle this, like updating the cache if the gserialized objects are equal in value but not in pointer, but removing the possibility altogether seemed like a simpler solution.

> Loading on every cache miss can mean taking a copy every time, in that case where we aren't seeing a repeating pattern.

Keep in mind that when we don't "persist" on the cache context here, we need to copy it in the GetGeomCach. Also, by using this shared cache we are only keeping a single copy of gserialized object, and not 2 like before.

Talking from memory, I think that PG_DETOAST_DATUM and PG_DETOAST_DATUM_COPY will act the same for toasted datums (you never get a non copied value since it doesn't exist on disk), so there shouldn't be any change performance wise if that worries you.

libpgcommon/lwgeom_cache.h
Algunenano: I'll have a look. I did remove the size since that was already included in the gserialized struct itself (just behind the VARSIZE macros), so I'll try to clean the lwgeom pointers too if they are unused.

libpgcommon/shared_gserialized.c
Algunenano: This is only used in `ToastCacheGetGeometry` which will return the pointer (which is good since we want to return a pointer to the local context when not caching, and a pointer to the cache context when caching) so the palloc will be needed either here or there.

Algunenano:
Thanks a lot for the review, @pramsey

PR #565, opened 2020-06-05
Running optipng to reduces the size of png files

lhirlimann:
Reduce the size of the png files in the project.

I ran find . -name "*.png" | xargs optipng -o7

size of all files before 0.00201134 , after 0.0017761.
sizes obtained using
```
find . -name "*.png" -print0 | du -sb --files0-from=-  | awk '{ total += $1} END { print total/1024/1024/1024 }' 
```


Algunenano:
Pushed. Thanks a lot.

PR #566, opened 2020-06-08
Fix ST_locate_between_measures

Kaplas80:
Fixes a bug in legacy function "ST_locate_between_measures"


pramsey:
Thanks, applied via https://trac.osgeo.org/postgis/ticket/4702

PR #568, opened 2020-06-26
Travis: Use PG13 as the default target

Algunenano:


Algunenano:
Merged in https://github.com/postgis/postgis/commit/17e3e8fd60319f67ce071df3c290e96550c818ed

PR #579, opened 2020-09-22
Vendor in libprotobuf-c

Algunenano:


gdt:
This is contrary to packaging norms.  The system version should be used, and an included version should be treated as a workaround with a warning.

Does postgis have a plan to have timely micro releases every time there is a CVE fix in protobuf?

gdt:
Also, the commit message does not really describe the change and discuss the consequences.  But I was able to figure that out from the diff.

Algunenano:
Closing this. I'm planning on creating a different one keeping the system protobuf-c and throwing an error if the dependency isn't met (or opted out).

PR #582, opened 2020-10-02
Enforce protobuf-c dependency

Algunenano:
Unless explicitly disabled, enforce the dependency on libprotobuf-c >= 1.1.0.

Previously it would fail silently and disabled MVT support, which translated into end user complaints.


Komzpa:
:tada:

PR #583, opened 2020-10-05
FlatGeobuf format input/output

bjornharrtell:
See https://github.com/bjornharrtell/flatgeobuf for more on the format.

See for https://lists.osgeo.org/pipermail/postgis-devel/2020-September/028610.html for a discussion in relation to PostGIS.

Proposed input functions:

* ST_TableFromFlatGeobuf(text, text, bytea) RETURNS void
* ST_FromFlatGeobuf(anyelement, bytea) RETURNS setof anyelement

The design is derived from the proposal on how to handle "Various complete table types" at https://stackoverflow.com/questions/11740256/refactor-a-pl-pgsql-function-to-return-the-output-of-various-select-queries.

Proposed output functions is ST_AsFlatGeobuf with similar signature and behaviour as ST_AsGeobuf.

Considering supporting Large Object in addition to bytea. Update: Initially seemed like a good fit but not sure anymore after reading up on it again.


postgis/flatgeobuf.h
```diff
@@ -0,0 +1,61 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with PostGIS.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ **********************************************************************
+ *
+ * Copyright (C) 2021 Björn Harrtell <bjorn@wololo.org>
```
Komzpa, line 21: 2020?

bjornharrtell: Fixed. Wasn't intentional to mean I won't try to finish this until 2021. :)

bjornharrtell:
I can compile this locally but it requires two deps, my own flatgeobuf includes  (6 headers only) and flatcc which is the native c implementation of flatbuffers. To depend on flatcc (https://github.com/dvidelabs/flatcc) it needs to build an link to static lib `libflatcc.a` from it's source, so it would mean a rather big code drop in `/deps`. I wonder if it would not make more sense to pull this in from (hopefully) a reliable CDN URL the build process?

Komzpa:
Hi, the build should have all the dependencies documented and checked on build stage. I see `flatbuffers-compiler-dev` in debian, if that is it - get it installed on CI and in documented build instructions.

bjornharrtell:
@Komzpa I don't need the schema compiler, that's part of the flatgeobuf dependencies. But I need the flatbuffers C implementation and it's not packaged for any distribution AFAIK and should be compiled in. (note that Debian/Ubuntu package libflatbuffers1 is C++ specific)

Algunenano:
> But I need the flatbuffers C implementation and it's not packaged for any distribution AFAIK and should be compiled in. (note that Debian/Ubuntu package libflatbuffers1 is C++ specific)

Are you talking about the libraries? I see this in my flatcc package:
```
$ pacman -Ql flatcc | grep so$
flatcc /usr/lib/libflatcc.so
flatcc /usr/lib/libflatccrt.so
```

Ideally I would depend on the system libraries if they exist. We could discuss if forcing packagers to also package flatcc would be ok or we could provide a secondary dependency where we build it in-tree if it's not available, but keep in mind that this would require keeping it up to date and so on.

bjornharrtell:
@Algunenano yes, libflatccrt.* is what I'm after, so it appears I was wrong to assume it's not packaged for any dist. It also appears nothing flatbuffers related is in Debian, the compiler and C++ runtime is Ubuntu only. Surprising and concerning. But then again, it might be more suitable to link statically for various reasons anyway.

Algunenano:
> @Algunenano yes, libflatccrt.* is what I'm after, so it appears I was wrong to assume it's not packaged for any dist. It also appears nothing flatbuffers related is in Debian, the compiler and C++ runtime is Ubuntu only. Surprising and concerning. But then again, it might be more suitable to link statically for various reasons anyway.

The runtime is fairly small (6 C files, https://github.com/dvidelabs/flatcc/tree/master/src/runtime) so you could bring only the necessary files, build a static library from it and link it. The wagyu dependency is doing pretty much that (except it is a C++ header only library so we build a single file that exposes a C API).

bjornharrtell:
@Algunenano agreed, I will attempt something like that.

bjornharrtell:
I've now reached a fully functional working implementation with what I would view as alpha-ish code. I do not know why the test run fails at CI, it passes for me locally.

I would highly appreciate feedback from any interested parties, both on design choices as detailed in the issue description and the code itself. It would be good to establish if I'm on right track before applying polish.

PS. @pramsey I still feel bytea is not suited for this even though it perhaps will work well enough for many cases. COPY semantics is what I would really want. :) Or a PostgreSQL feature exposing stdin/stdout to functions.

postgis/flatgeobuf.c
```diff
@@ -0,0 +1,1039 @@
...
+	flatcc_builder_init(B);
+
+	// inspect columns
+	// NOTE: last element will be unused if geom attr is found
+	for (int i = 0; i < natts; i++) {
+		Oid typoid = getBaseType(TupleDescAttr(tupdesc, i)->atttypid);
+		const char *key = TupleDescAttr(tupdesc, i)->attname.data;
+		POSTGIS_DEBUGF(3, "flatgeobuf: inspecting column definition for %s with oid %d", key, typoid);
+		if (ctx->geom_name == NULL) {
+			if (!geom_found && typoid == postgis_oid(GEOMETRYOID)) {
```
Algunenano, line 130: If you use postgis_oid, you need to ensure the caller with access to fcinfo calls `postgis_initialize_cache(fcinfo);`. In this case I believe (and I'm using Github as GUI so I'm likely to be wrong) that would be `pgis_asflatgeobuf_finalfn`. See a recent commit in https://github.com/postgis/postgis/commit/678f526670b6b1805495431e62f12072bb9485cf for examples that require this same fix.

I'll have a look at the whole PR once I can fetch the branch locally and see it in my local environment.

robe2:
Hey @bjornharrtell   you still working on this?

bjornharrtell:
@robe2 I lost momentum but would like to pick it up again sooner or later.

robe2:
@bjornharrtell  We will be releasing a PostGIS 3.2.0 alpha1 in about 1.5 weeks.  Any chance you will revisit this before then?

bjornharrtell:
@robe2 I will certainly try.

bjornharrtell:
Merge was easy.. but the code does not exactly seem to be in a stable mood:

Died at /src/postgis/regress/run_test.pl line 744.
 ./regress/core/flatgeobuf .. failed (psql exited with an error: /tmp/pgis_reg/test_17_out)
psql:flatgeobuf.sql:2: server closed the connection unexpectedly

bjornharrtell:
Tried replicating environment settings as in https://github.com/postgis/postgis/blob/master/ci/travis/run_tests.sh but still not able to get it to crash and stop fail as in CI.

Algunenano:
> Tried replicating environment settings as in https://github.com/postgis/postgis/blob/master/ci/travis/run_tests.sh but still not able to get it to crash and stop fail as in CI.

Keep in mind that the CI is detecting the failure when running under the Undefined Behaviour Sanitizer. In https://github.com/postgis/postgis/runs/3441796911 for example you can see that it's complaining about:
```sql
#0  0x00007f6919fbb3d8 in __flatbuffers_uoffset_read_from_pe (p=0x55f868abd361) at ../deps/flatcc/include/flatcc/flatcc_endian.h:87
No locals.
#1  0x00007f6919fbb7f2 in flatbuffers_read_size_prefix (b=0x55f868abd361, size_out=0x7fff00fa8d80) at ../deps/flatgeobuf/flatbuffers_common_reader.h:545
No locals.
#2  0x00007f6919fc54e7 in flatgeobuf_decode_header (ctx=0x55f868abdfc8) at flatgeobuf.c:571
        header = 0x868abc718
        columns = <optimized out>
        size = 94525396341593
#3  0x00007f6919fcdf68 in pgis_tablefromflatgeobuf (fcinfo=0x55f868abc800) at 
```

The error message from the sanitizer is likely in the Postgresql logs but it doesn't seem that we output those (or the last N rows)  in the CI. You can either use any of the 2 docker images (it complains both in gcc ubsan and clang ubsan) or build postgresql with those flags and then build postgis with the appropiate flags (https://github.com/postgis/postgis/blob/f6def67654c25d812446239036cee44812613748/ci/travis/run_usan_clang.sh)

bjornharrtell:
Problem is I must not use the Ubuntu bundles postgresql locally but rather compile that myself too with the right incantations to get full debug and memory checks in place.

Algunenano:
You could probably modify the CI so that it outputs that log on error instead

bjornharrtell:
@Algunenano hmm but what I feel I lack is a proper "local" development environment that reproduces the error. I think I need to move to a setup using the docker builds or compile all the things myself.. not sure what will be best/easiest for me yet.

bjornharrtell:
Looks like something like this is possibly what I'm looking for:

> docker run -it --mount type=bind,source=/home/bjorn/code/postgis,target=/src/postgis postgis/postgis-build-env:pg14-clang-geosmain-gdal31-proj71 /bin/bash
> source ./ci/travis/run_usan_clang_bjorn.sh
> ./regress/run_test.pl regress/core/flatgeobuf --verbose

I customized run_usan_clang.sh a little bit. Dies as expected and allows me to quickly change the code locally and have the container running for quick recompile and test runs.

If this is a sound setup I'm considering writing a small section about it in an appropriate place.. unless I've managed to miss an existing documentation snippet about how to get this kind of setup in place.

bjornharrtell:
Unfortunately I have been unable to track down what is causing the crasch:

```
Program terminated with signal SIGILL, Illegal instruction.
#0  0x00007feab0c0c438 in __flatbuffers_uoffset_read_from_pe (p=0x56155f08d331) at ../deps/flatcc/include/flatcc/flatcc_endian.h:87
87	 __flatcc_define_integer_accessors(__flatbuffers_uoffset, flatbuffers_uoffset_t,

Thread 1 (Thread 0x7feab9c5f740 (LWP 8657)):
#0  0x00007feab0c0c438 in __flatbuffers_uoffset_read_from_pe (p=0x56155f08d331) at ../deps/flatcc/include/flatcc/flatcc_endian.h:87
No locals.
#1  0x00007feab0c0c852 in flatbuffers_read_size_prefix (b=0x56155f08d331, size_out=0x7ffecd3b9ee0) at ../deps/flatgeobuf/flatbuffers_common_reader.h:545
No locals.
#2  0x00007feab0c16547 in flatgeobuf_decode_header (ctx=0x56155f08df98) at flatgeobuf.c:571
```

bjornharrtell:
@Algunenano I was having trouble getting debug level logs in the setup.. but with that solved (https://github.com/postgis/postgis/pull/637) I can see this in the logs:

2021-09-02 14:43:20.571 UTC [21651] DEBUG:  [lwgeom_in_flatgeobuf.c:pgis_tablefromflatgeobuf:101] flatgeobuf: pgis_tablefromflatgeobuf bytea data size is 88
2021-09-02 14:43:20.571 UTC [21651] DEBUG:  [flatgeobuf.c:flatgeobuf_decode_header:570] flatgeobuf: reading header size prefix at 8
2021-09-02 14:43:20.753 UTC [14709] DEBUG:  reaping dead processes
2021-09-02 14:43:20.753 UTC [14709] DEBUG:  server process (PID 21651) was terminated by signal 4: Illegal instruction

And I now realize that the size variable garbage content is a red herring - it doesn't get that far because it crashes when actually trying to read that value from the buffer, here https://github.com/bjornharrtell/postgis/blob/747bf51385bb37e654f080c1125c5cc9955565d5/postgis/flatgeobuf.c#L571. But how can that be when it looks fine.. bytea data size 88 bytes and read attempt at offset 8 bytes. I suppose `ctx->buf` must be wrong/corrupt for some reason and not represent the 88 byte array at all.

bjornharrtell:
Getting nowhere so I've reached out to flacc project as the core dump happens in a support function from there, https://github.com/dvidelabs/flatcc/issues/209.

Algunenano:
Please note it's not a normal coredump. It is the undefined sanitizer detecting there was a call to a function that is an undefined behaviour according to the C standard.

bjornharrtell:
@Algunenano I'm having difficulties understanding what that means. Can you clarify?

Algunenano:
Have a look at the doc page for it: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

bjornharrtell:
Getting somewhere. I've determined that it is specifically crashing with -fsanitize=alignment and that it can be fixed by doing memcpy of the VARDATA_ANY data into the buffer instead of using it directly and there is the "caution: this will return a possibly unaligned pointer" on VARDATA_ANY so that makes sense. However it does not resolve the test failureand later in the logic it spits out "WARNING:  problem in alloc set ExprContext: detected write past chunk" so I must have some broken logic somewhere.

robe2:
@bjornharrtell I'm rooting for you.  I'll test out what you have so far on my system tomorrow to see how it goes.  If it's just the alignment compile switch thing and it works, I'm fine with accepting it for PostGIS 3.2.0alpha1 with plan to clean it up before PostGIS 3.2.0 release time.  Of course would require consensus in the PSC.

I'm holding off on releasing PostGIS 3.2.0alpha1 until sometime next week so I can do a bit more testing with pg_upgrade etc and hopes that we can get this in there.

bjornharrtell:
Thanks @robe2 :) I think I really need to figure out that problem though... I've been able to reproduce it outside of this implementation in a basic testcase for flatcc (the c implementation of flatbuffers) and I'm trying to figure it out / get help in this issue - https://github.com/dvidelabs/flatcc/issues/210.

robe2:
> 
> 
> Thanks @robe2 :) I think I really need to figure out that problem though... I've been able to reproduce it outside of this implementation in a basic testcase for flatcc (the c implementation of flatbuffers) and I'm trying to figure it out / get help in this issue - [dvidelabs/flatcc#210](https://github.com/dvidelabs/flatcc/issues/210).

FWIW just tested under mingw64 gcc 8.1.0 and mys 2 and get similar to the CIs. At least it compiles :)

```
PostgreSQL 14beta3 on x86_64-w64-mingw32, compiled by gcc.exe (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 8.1.0, 64-bit
  Postgis 3.2.0dev - (3.1.0rc1-500-g5c7006633) - 2021-09-05 17:39:41
  scripts 3.2.0dev 3.1.0rc1-500-g5c7006633
  raster scripts 3.2.0dev 3.1.0rc1-500-g5c7006633
  GEOS: 3.10.0dev-CAPI-1.15.0
  PROJ: 7.1.1
  SFCGAL: 1.3.10
  GDAL: GDAL 3.2.3, released 2021/04/27


 ./regress/core/flatgeobuf .. failed (diff expected obtained: /projects/postgis/tmp/3.2.0dev_pg14_geos3.10_gdal3.2.0w64/test_17_diff)
--- ./regress/core/flatgeobuf_expected  2021-09-05 13:37:09.605547600 -0400
+++ /projects/postgis/tmp/3.2.0dev_pg14_geos3.10_gdal3.2.0w64/test_17_out       2021-09-05 13:46:42.325757900 -0400
@@ -1,21 +1,30 @@
 --- Null geometry ---
 T1|0|
 --- Geometry roundtrips ---
-P1|0|POINT(1.1 2.1)
-P1|0|POINT(1.1 2.1)
-P2|0|POINT Z (1.1 2.11 3.2)
-P3|0|POINT ZM (1.1 2.12 3.2 4.3)
-L1|0|LINESTRING(1 2,3 4)
-L2|0|LINESTRING Z (1 2 3,3 4 5)
-L3|0|LINESTRING ZM (1 2 3 5,3 4 5 6)
-P1|0|POLYGON((75 29,77 29,77 29,75 29))
-P2|0|POLYGON((35 10,45 45,15 40,10 20,35 10),(20 30,35 35,30 20,20 30))
-MP1|0|MULTIPOINT(10 40,40 30,20 20,30 10)
-ML1|0|MULTILINESTRING((10 10,20 20,10 40),(10 10,20 20,10 40,40 40,30 30,40 20,30 10))
-MP1|0|MULTIPOLYGON(((40 40,20 45,45 30,40 40)),((20 35,10 30,10 10,30 5,45 20,20 35),(30 20,20 15,20 25,30 20)))
-GC1|0|GEOMETRYCOLLECTION(POINT(40 10),LINESTRING(10 10,20 20,10 40),POLYGON((40 40,20 45,45 30,40 40)))
+P1|0|
+P1|0|
+P2|0|
+P3|0|
+L1|0|
+L2|0|
+L3|0|
+P1|0|
+P2|0|
+MP1|0|
+ML1|0|
+MP1|0|
+GC1|0|
 --- Multiple rows ---
-P1|0|POINT(1 2)
-P1|1|POINT(3 4)
+P1|0|
+P1|1|
 --- Attribute roundtrips ---
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 0000000008042048
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 0000000008042070
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 0000000008042098
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 00000000080420C0
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 00000000080420F8
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 0000000008042120
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 0000000008042148
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 0000000008042180
+WARNING:  problem in alloc set ExprContext: detected write past chunk end in block 0000000008041CD0, chunk 00000000080422C0
 A1|0||t|1|2|3|4|1.2|1.3|hello

```

bjornharrtell:
Finally getting somewhere with concluding that https://github.com/postgis/postgis/pull/583/commits/2bbe3bc20b14c6f9df98a2e757c36bfe344dfb96 caused much of bad stuff.

Only out functions tested for now.. They pass with sanitizer for me locally, but at CI it still crashes hard on some flatcc stuff.. and the behaviour of flatcc is still being investigated.

bjornharrtell:
Strange.. I cannot reproduce the failure locally when running with the "pg13-geos39-gdal31-proj71, usan_gcc" setup... however I do get a crash when using the "pg13-clang-geos39-gdal31-proj71, usan_clang" setup but that is not fun because it gives no core dump.

robe2:
My mingw64/msys2 gcc 8.1 on PG14 regresses cleanly after removal of null geometry test at 75cde2c
I guess that ain't much help in troubleshooting the issues :(

robe2:
@bjornharrtell also don't forget the documentation for the manual.  Assuming you are saving that for last.

bjornharrtell:
I still suspect the remaining issues I have are related to the flatcc implementation. I'm considering abandoning it and instead use the original c++ flatbuffers implementation via c api wrapper.

robe2:
> 
> 
> I still suspect the remaining issues I have are related to the flatcc implementation. I'm considering abandoning it and instead use the original c++ flatbuffers implementation via c api wrapper.

Okay I still plan to release tomorrow, but we can still sneak this in for next alpha/beta release if you are not done by EOD tomorrow.

robe2:
Okay I pushed this to my gitea repo to have dronie test and doesn't compile under the trisquel2 docker image we have for testing on dronie.

https://dronie.osgeo.org/robe/postgis/37/1/2

robe2:
On closer inspection of the dronie failures, they all are caused by defined but not used which (compliments of @strk ) is any warning is considered a no-no.

```
In file included from src/runtime/json_printer.c:25:0:
include/flatcc/portable/pprintint.h:465:27: error: 'print_int' defined but not used [-Werror=unused-function]
 __define_print_int_simple(print_int, print_uint, int, unsigned int)
                           ^
include/flatcc/portable/pprintint.h:454:12: note: in definition of macro '__define_print_int_simple'
 static int NAME(T n, char *buf)                                             \
            ^
include/flatcc/portable/pprintint.h:466:27: error: 'print_long' defined but not used [-Werror=unused-function]
 __define_print_int_simple(print_long, print_ulong, long, unsigned long)
                           ^
include/flatcc/portable/pprintint.h:454:12: note: in definition of macro '__define_print_int_simple'
 static int NAME(T n, char *buf)         

cc1: all warnings being treated as errors  
```

bjornharrtell:
Closing as #641 has effectively replaced this.

PR #589, opened 2020-11-27
ST_GeomFromGML(): fix axis order issue with urn:ogc:def:crs:EPSG:: syntax on projected CRS with northing,easting axis order

rouault:
It was assumed that coordinate swapping was only necessary on geographic CRS
when srsName='urn:ogc:def:crs:EPSG::XXXX', but there are projected CRS such
as EPSG:2393 that have a northing,easting axis order too.

This is very similar to a TinyOWS fix I just did in
https://github.com/MapServer/tinyows/pull/94


rouault:
The 2 failures of https://travis-ci.org/github/postgis/postgis/builds/746605298 on tag=pg10-geos36-gdal23-proj49 mode=tests on regress/core/regress_selectivity are IMHO completely unrelated to that work.

pramsey:
Thanks!

PR #597, opened 2021-01-26
Use pkg-config to get PROJ version if available

fidomax:
In case of cross-compile AC_PROJ_VERSION([POSTGIS_PROJ_VERSION]) fails.
So try to get version by pkg-config first

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -864,6 +864,7 @@ elif test ! -z "$PKG_CONFIG"; then
 		[
 			PROJ_CPPFLAGS="$PROJ_CFLAGS"
 			PROJ_LDFLAGS="$PROJ_LIBS"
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -886,7 +886,13 @@ AC_CHECK_HEADER([proj_api.h],
 	)
 
 dnl Return the PROJ.4 version number
-AC_PROJ_VERSION([POSTGIS_PROJ_VERSION])
+if test ! -z "$PKG_CONFIG"; then
```
pramsey, line 889: There's a check for PKG_CONFIG earlier in the script where this could go much more sensibly. Then AC_PROJ_VERSION can just be conditional on POSTGIS_PROJ_VERSION not already being set.

PR #602, opened 2021-03-02
Fix formatting errors in reference_accessor.xml

sharkAndshark:
affect doc only


dr-jts:
Fixed this in a separate commit.

PR #603, opened 2021-03-21
ST_Boundary(TIN) should return linework of TIN

kalenikaliaksandr:
https://trac.osgeo.org/postgis/ticket/4401
https://trac.osgeo.org/postgis/ticket/3779


liblwgeom/lwgeom_geos.c
```diff
@@ -517,7 +517,7 @@ LWGEOM2GEOS(const LWGEOM* lwgeom, uint8_t autofix)
 			geostype = GEOS_MULTIPOINT;
 		else if (lwgeom->type == MULTILINETYPE)
 			geostype = GEOS_MULTILINESTRING;
-		else if (lwgeom->type == MULTIPOLYGONTYPE)
+		else if (lwgeom->type == MULTIPOLYGONTYPE || lwgeom->type == TINTYPE)
```
Komzpa, line 520: can you also add TRIANGLE into POLYGON?

Komzpa: oh it is there already, disregard.

postgis/lwgeom_geos.c
```diff
@@ -846,51 +846,105 @@ Datum boundary(PG_FUNCTION_ARGS)
...
-		GEOSGeom_destroy(g1);
-		HANDLE_GEOS_ERROR("GEOSBoundary");
-	}
+			break;
+		}
+		case MULTILINETYPE:
+		{
+			LWGEOM *lwmpoint = (LWGEOM *)lwmpoint_from_lwgeom(lwgeom);
+			// it is not going to work
+			lwgeom_remove_repeated_points_in_place(lwmpoint, 0.0001);
```
kalenikaliaksandr, line 879: points that occur even number of times should be left and points that occur odd number of times should be removed. am I right?

Komzpa: I believe it was about endpoints only, not any points.

postgis/lwgeom_geos.c
```diff
@@ -846,51 +846,105 @@ Datum boundary(PG_FUNCTION_ARGS)
...
-
-	g1 = LWGEOM2GEOS(lwgeom, 0);
-	lwgeom_free(lwgeom);
-
-	if (!g1)
-		HANDLE_GEOS_ERROR("First argument geometry could not be converted to GEOS");
+		case POINTTYPE:
+		{
+			LWPOINT *lwpoint = (LWPOINT *)lwgeom;
+			lwpoint->point->npoints = 0;
```
Komzpa, line 854: feels nasty. can you lwgeom_construct_empty instead?

postgis/lwgeom_geos.c
kalenikaliaksandr: oh, yes, you are right. so only endpoints that occur odd number of times should be included in result.

postgis/lwgeom_geos.c
```diff
@@ -846,51 +846,105 @@ Datum boundary(PG_FUNCTION_ARGS)
...
+			// it is not going to work
+			lwgeom_remove_repeated_points_in_place(lwmpoint, 0.0001);
+			lwgeom_free(lwgeom);
+			lwgeom = lwmpoint;
 
-	POSTGIS_DEBUGF(3, "result: %s", GEOSGeomToWKT(g3));
+			break;
+		}
+		case TRIANGLETYPE:
+			lwgeom->type = LINETYPE;
```
Komzpa, line 886: this works by luck that TIN and LINE mem layout is the same. either assert it here (at least that sizeof() of both structures is the same) or rebuild from scratch.

postgis/lwgeom_geos.c
```diff
@@ -846,105 +847,140 @@ Datum boundary(PG_FUNCTION_ARGS)
...
+
+			POINT4D *out = lwalloc(sizeof(POINT4D) * lwmline->ngeoms * 2);
+			uint32_t *counts = lwalloc(sizeof(uint32_t) * lwmline->ngeoms * 2);
+			uint32_t n = 0;
+			for (i = 0; i < lwmline->ngeoms * 2; i++) {
+				int seen = 0;
+				for (j = 0; j < n; j++) {
+					if (memcmp(&(endpoints[i]), &(out[j]), sizeof(POINT4D)) == 0)
+					{
+						seen = 1;
```
Komzpa, line 896: 1 is LW_TRUE

Komzpa:
Please add docs:
 - NEWS one-liner
 - `Changed: 3.2.0 support for TIN, does not use geos` in the docs xml

liblwgeom/lwgeom.c
```diff
@@ -2534,3 +2534,181 @@ void lwgeom_trim_bits_in_place(LWGEOM* geom, int32_t prec_x, int32_t prec_y, int
...
+					POINTARRAY *points = ptarray_clone_deep(lwtin->geoms[i]->points);
+					lwmline_add_lwline(lwmline, lwline_construct(srid, lwtin->geoms[i]->bbox, points));
+				}
+
+				lwresult = (LWGEOM *)lwmline;
+			}
+
+			break;
+		}
+		case COLLECTIONTYPE:
```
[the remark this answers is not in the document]
kalenikaliaksandr: From OGC:
> The boundary of an arbitrary collection of geometric objects whose interiors are disjoint consists of geometric objects drawn from the boundaries of the element geometric objects by application of the “mod 2” union rule (Bibliographic Reference [1], section 3.12.3.2).

So my assumption that boundary of geometry collection is collection of elements boundaries must to be wrong. I want to make it to meet the specification but not sure how boundary is supposed to be calculate for collection whose interiors are _not_ disjoint.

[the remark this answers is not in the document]
kalenikaliaksandr: From OGC:
> The subtypes of GeometryCollection, to be presented next, shall ensure the following constraints, which are not automatically ensured by aggregates of the Spatial schema. These constraints are summarized as follows.
> a) For every element in a GeometryCollection, its interior shall be disjoint to the interior of every other, but distinct element of the same GeometryCollection.
> b) For every boundary of an element in a GeometryCollection, it may only intersect with a boundary of another, but distinct element at most in a finite number of points.

Does that mean that it is acceptable to output wrong boundary for geometry collection that doesn't meet these constraints?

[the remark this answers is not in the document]
Komzpa: My reading is "If you do UnaryUnion on your collection before, you can expect a boundary to be then edge between meat and exterior of collection. If you don't, you get back something you may not like."

liblwgeom/lwgeom.c
```diff
@@ -2534,3 +2534,181 @@ void lwgeom_trim_bits_in_place(LWGEOM* geom, int32_t prec_x, int32_t prec_y, int
...
+			}
+
+			lwresult = (LWGEOM *)lwmline;
+			break;
+		}
+		case TINTYPE:
+		{
+			LWTIN *lwtin = lwgeom_as_lwtin(lwgeom);
+
+			if (lwtin->ngeoms == 1)
```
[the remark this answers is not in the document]
kalenikaliaksandr: GEOS returns the geometry itself if it is the only one in a collection so I tried to replicate that behaviour. Likely I would prefer to have more consistent output too.

strk:
This PR opened a can of worms.
Just a couple are reported here:
https://trac.osgeo.org/postgis/ticket/4956
https://trac.osgeo.org/postgis/ticket/4960

How about still using GEOS for everything supported there ?

strk:
@kalenikaliaksandr are you willing to fix all the issues and complete work on unit testing ?
I have it started here: https://gitlab.com/postgis/postgis/-/merge_requests/50

strk:
BTW, once Boundary is touched it may be good to add support for a switch to specify the BoundaryNodeRule, as is if accepted in ST_Relate (https://postgis.net/docs/ST_Relate.html)

kalenikaliaksandr:
> @kalenikaliaksandr are you willing to fix all the issues and complete work on unit testing ?
> I have it started here: https://gitlab.com/postgis/postgis/-/merge_requests/50

yes, I will try to work through the issues. Thanks for findings.

strk:
I think a geometry component iterator will be useful, for nested collections...

PR #604, opened 2021-04-02
GDAL interpolation and gridding bindings

pramsey:
Adds
```
CREATE OR REPLACE FUNCTION ST_GDALContour(
		rast raster,
		bandnumber integer DEFAULT 1,
		level_interval float8 DEFAULT 100.0,
		level_base float8 DEFAULT 0.0,
		fixed_levels float8[] DEFAULT ARRAY[]::float8[],
		polygonize boolean DEFAULT false
		)
	RETURNS table(geom geometry, id integer, value float8)
	AS 'MODULE_PATHNAME', 'RASTER_GDALContour'
	LANGUAGE 'c' IMMUTABLE STRICT PARALLEL SAFE;

CREATE OR REPLACE FUNCTION ST_GDALGrid(
		geom geometry,
		options text,
		rast raster,
		bandnumber integer DEFAULT 1
		)
	RETURNS raster
	AS 'MODULE_PATHNAME', 'RASTER_GDALGrid'
	LANGUAGE 'c' IMMUTABLE STRICT PARALLEL SAFE;
```


Komzpa:
Can it not have GDAL in the name please?

dbaston:
> 		level_interval float8 DEFAULT 100.0,
>		level_base float8 DEFAULT 0.0,

What do you think about defaulting to a fixed number of intervals, based on the range of the data?

> 		options text,

I wonder if there is some way to avoid leaking the GDAL config string format to the PostGIS caller. I guess it's not possible for a PostgreSQL function to accept arbitrary keyword arguments, but maybe it's worth at least expanding it to `interpolation_method text DEFAULT 'invdist', interpolation_options text`  ?

raster/rt_core/rt_gdal.c
```diff
@@ -0,0 +1,217 @@
...
+		return _rti_contour_arg_destroy(&arg);
+
+	// GDALContourGenerate( GDALRasterBandH hBand,
+	//   double dfContourInterval, double dfContourBase,
+	//   int nFixedLevelCount, double *padfFixedLevels,
+	//   int bUseNoData, double dfNoDataValue,
+	//   void *hLayer, int iIDField, int iElevField,
+	//   GDALProgressFunc pfnProgress, void *pProgressArg );
+
+	static int use_no_data = 0;
```
rouault, line 159:
```suggestion
	int use_no_data = 0;
```

raster/rt_core/rt_gdal.c
```diff
@@ -0,0 +1,217 @@
...
+	static double no_data_value = 0.0;
```
rouault, line 160:
```suggestion
	double no_data_value = GDALGetRasterNoDataValue(hBand, &use_no_data);
```

raster/rt_core/rt_gdal.c
```diff
@@ -0,0 +1,217 @@
...
+	if (cplerr != CE_None)
+		return _rti_contour_arg_destroy(&arg);
+
+	/* Convert the OGR layer into PostGIS geometries */
+	nfeatures = OGR_L_GetFeatureCount(arg.dst.lyr, TRUE);
+	if (nfeatures < 0)
+		return _rti_contour_arg_destroy(&arg);
+
+	*contours = rtalloc(sizeof(struct rt_contour_t) * nfeatures);
+	for (i = 0; i < nfeatures; i++) {
```
rouault, line 182: iterating over features with OGR_L_GetFeature() is strongly discouraged as you make assumption on the values of the FID. Use rather OGR_L_GetNextFeature() and stop when NULL is returned

raster/rt_core/rt_gdal.c
```diff
@@ -0,0 +1,217 @@
...
+		struct rt_contour_t contour;
+		OGRGeometryH hGeom;
+		OGRFeatureH hFeat;
+		LWGEOM *geom;
+		if (!(hFeat = OGR_L_GetFeature(arg.dst.lyr, i))) continue;
+		/* Store elevation/id */
+		contour.id = OGR_F_GetFieldAsInteger(hFeat, 0);
+		contour.elevation = OGR_F_GetFieldAsDouble(hFeat, 1);
+		/* Convert OGR geometry to LWGEOM via WKB */
+		if (!(hGeom = OGR_F_GetGeometryRef(hFeat))) continue;
```
rouault, line 194:
```suggestion
		if (!(hGeom = OGR_F_GetGeometryRef(hFeat)))
		{
		    OGR_F_Destroy(hFeat);
		    continue;
		}
```

pramsey:
> Can it not have GDAL in the name please?

ST_InterpolateRaster() and ST_Contour() ?... note that ST_FromGDALRaster and ST_AsGDALRaster() exist, but most importantly, ST_GDALWarp() is an exact analogue. Should be ST_Warp() perhaps, but it's not.

pramsey:
> What do you think about defaulting to a fixed number of intervals, based on the range of the data?

Pretty -0 on that one, I rather prefer the default base/interval scheme to a "you always get 10 contours" scheme. For me it fits the principle of least surprise better.
> > ```
> >   options text,
> > ```

> I wonder if there is some way to avoid leaking the GDAL config string format to the PostGIS caller. I guess it's not possible for a PostgreSQL function to accept arbitrary keyword arguments, but maybe it's worth at least expanding it to `interpolation_method text DEFAULT 'invdist', interpolation_options text` ?

The GDAL string has the advantage of being fully documented already. I initially was looking at having a text array of options, and an algorithm string, and I was just going to have to have all the strings in my code so I could validate inputs, and a bunch of code to create the rights options struct, and I was resigned to that, but then I saw GDAL has exposed the config string parser so I could outsource all that, and it seemed just way smarter, particularly since I also got doco for free.

pramsey:
So is a change to ST_InterpolateRaster() and ST_Contour() generally agreed?

pramsey:
Also, a name for a function that takes a raster and a geometry and copies the raster values into the Z or M ordinate of the vector as necessary? 

* ST_Drape(geometry, ordinate, raster, bandnum) => geometry? would be Z-specific in connotation. 
* ST_CopyZ and ST_CopyM?
* Something else?
* Also, does this already exist?

robe2:
> 
> 
> Can it not have GDAL in the name please?

Agree.  Those extra characters add more to typing and expose an implementation detail that is not required to be exposed.  In the case of ST_AsGDALRaster we had to put something there to distinquish from our internal raster format, but even there I question if GDAL was the right term to use since the end result is just a particular type of raster like PNG / JPEG /TIFF.

robe2:
> 
> 
> > Can it not have GDAL in the name please?
> 
> ST_InterpolateRaster() and ST_Contour() ?... note that ST_FromGDALRaster and ST_AsGDALRaster() exist, but most importantly, ST_GDALWarp() is an exact analogue. Should be ST_Warp() perhaps, but it's not.

In hind site I have no idea why we put GDAL on those aside from to distinguish from our own internal format.
Yes ST_GDALWarp should have just been ST_Warp().  Note that ST_Transform, ST_Rescale and friends I think all internally use GDALWarp but don't have that on the name.

robe2:
> ```
> CREATE OR REPLACE FUNCTION ST_GDALContour(
> 		rast raster,
> 		bandnumber integer DEFAULT 1,
> 		level_interval float8 DEFAULT 100.0,
> 		level_base float8 DEFAULT 0.0,
> 		fixed_levels float8[] DEFAULT ARRAY[]::float8[],
> 		polygonize boolean DEFAULT false
> 		)
> 	RETURNS table(geom geometry, id integer, value float8)
> 	AS 'MODULE_PATHNAME', 'RASTER_GDALContour'
> 	LANGUAGE 'c' IMMUTABLE STRICT PARALLEL SAFE;
> 
> CREATE OR REPLACE FUNCTION ST_GDALGrid(
> 		geom geometry,
> 		options text,
> 		rast raster,
> 		bandnumber integer DEFAULT 1
> 		)
> 	RETURNS raster
> 	AS 'MODULE_PATHNAME', 'RASTER_GDALGrid'
> 	LANGUAGE 'c' IMMUTABLE STRICT PARALLEL SAFE;
> ```

Don't forget the costing

Developer66:
Hey @pramsey,
thank you very much for the work!

Is the `polygonize` feature for `ST_Contour`  planed or possible to implement in the future?

Thanks for your answer
Tim

pramsey:
I wasn't super planning on it because I it added more parameter combos to deal with, and it felt a little superfluous since PostGIS supports other ways of polygonizing things.

pramsey:
What parameters do you use to call ST_Contour?

Developer66:
`SELECT (ST_Contour(rast, 1, level_interval := 0.25, level_base := 1)).* FROM raster`

or this using `ST_Polygonize`
```
WITH x as (SELECT (ST_Contour(rast, 1, level_interval := 0.25, level_base := 1)).* FROM raster)
SELECT st_astext((st_dump(st_polygonize(geom))).geom), value
FROM x
GROUP BY value

pramsey:
Check #737.

pramsey:
This fix is committed for upcoming 3.4 release.

Developer66:
Thanks you very much!

PR #605, opened 2021-04-11
configure: add explicit checking for postgres.h availability

rouault:
On Ubuntu, one needs to install the postgresql-server-dev-XX package
to get postgres.h


rouault:
Dropping the ball. No idea why this fails on the CI setup

PR #608, opened 2021-04-17
configure.ac: fix cross-compilation with protobuf-c

ffontaine:
Try to retrieve the version from pkg-config to avoid the following error:

```
checking protobuf-c version... configure: error: in `/home/fabrice/buildroot/output/build/postgis-3.1.1':
configure: error: cannot run test program while cross compiling
```

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


pramsey:
Thanks!

PR #609, opened 2021-04-19
improve performance of removerepeatedpoints(multipoint)

kalenikaliaksandr:


liblwgeom/lwgeom.c
```diff
@@ -1606,55 +1621,29 @@ lwgeom_remove_repeated_points_in_place(LWGEOM *geom, double tolerance)
...
+			for (uint32_t j = 1; j < mpt->ngeoms; j++)
 			{
-				int seen = 0;
-				LWPOINT *p1 = mpt->geoms[i];
-				const POINT2D *pt1 = getPoint2d_cp(p1->point, 0);
-				for (j = 0; j < n; j++)
+				POINT2D ptj, pti;
+				getPoint2d_p(mpt->geoms[i]->point, 0, &pti);
+				getPoint2d_p(mpt->geoms[j]->point, 0, &ptj);
+				if (distance2d_sqr_pt_pt(&ptj, &pti) > tolsq)
```
Komzpa, line 1634: This is not enough - for instance, in the square in the center of this image there can possibly be three points within the distance.
image

Your method can still be a special case for tolerance=0, just plug back in old one :)

I see three big options how to implement this:

I. via double grid snap:
 - snap to grid of size = tolerance
 - check all the ones in the same bucket if they are the same
 - snap to grid of size = tolerance with a shift of 0.5 
 - check all the ones in the same bucket if they are the same
 
II. double-sorting with window tracking:
 - order by X
 - iterate through list, look for duplicates back in the list until X is within tolerance
 - order by Y
 - iterate through list, look for duplicates back in the list until Y is within tolerance
 
III. tree
 - build a tree
 - while building a tree insertions, skip inserting points that have a point in tree within radius.

liblwgeom/lwgeom.c
```diff
@@ -1549,6 +1549,21 @@ void lwgeom_set_srid(LWGEOM *geom, int32_t srid)
...
+{
+	LWPOINT *p1 = *((LWPOINT **)pn1);
+	LWPOINT *p2 = *((LWPOINT **)pn2);
+
+	GBOX b1, b2;
+	lwgeom_calculate_gbox((LWGEOM *)p1, &b1);
+	lwgeom_calculate_gbox((LWGEOM *)p2, &b2);
+
+	uint64_t h1, h2;
+	h1 = gbox_get_sortable_hash(&b1, 0);
```
Komzpa, line 1563: will probably be neater to refactor it to also allow points (it casts box back into a point internally and that piece can be isolated), and pass original srid

liblwgeom/lwgeom.c
```diff
@@ -1606,55 +1621,29 @@ lwgeom_remove_repeated_points_in_place(LWGEOM *geom, double tolerance)
...
-			for (i = 0; i < mpt->ngeoms; i++)
+			uint32_t i = 0;
+			for (uint32_t j = 1; j < mpt->ngeoms; j++)
 			{
-				int seen = 0;
-				LWPOINT *p1 = mpt->geoms[i];
-				const POINT2D *pt1 = getPoint2d_cp(p1->point, 0);
-				for (j = 0; j < n; j++)
+				POINT2D ptj, pti;
+				getPoint2d_p(mpt->geoms[i]->point, 0, &pti);
```
Komzpa, line 1632: can you get a _cp ("const pointer") here? it will point directly to memory in pointarray and skip copying

liblwgeom/lwgeom.c
```diff
@@ -1606,55 +1629,46 @@ lwgeom_remove_repeated_points_in_place(LWGEOM *geom, double tolerance)
...
+					const POINT2D *pti = getPoint2d_cp(mpt->geoms[i]->point, 0);
+					for (uint32_t j = i + 1; j < mpt->ngeoms; j++)
 					{
-						seen = 1;
-						break;
+						if (mpt->geoms[j] == NULL)
+							continue;
+
+						const POINT2D *ptj = getPoint2d_cp(mpt->geoms[j]->point, 0);
+						if ((dim == 0 ? ptj->x - pti->x : ptj->y - pti->y) > tolerance)
```
Komzpa, line 1650: Sounds like this should be in the outer loop condition and as a whole it should be a while loop - no need to init j if all points are farther away than dedup threshold

kalenikaliaksandr: hm, I don't think I can move this in outer loop since array is sparse and not null item in points array need to be found first

liblwgeom/lwgeom.c
```diff
@@ -1606,55 +1629,46 @@ lwgeom_remove_repeated_points_in_place(LWGEOM *geom, double tolerance)
...
-			if (use_heap)
-				out = lwalloc(sizeof(LWMPOINT *) * mpt->ngeoms);
-			else
-				out = out_stack;
+			double tolsq = tolerance * tolerance;
+			LWMPOINT *mpt = (LWMPOINT *)geom;
 
-			/* Inefficient O(n^2) implementation */
-			for (i = 0; i < mpt->ngeoms; i++)
+			for (uint32_t dim = 0; dim < 2; dim++)
```
Komzpa, line 1635:
```suggestion
			for (uint8_t dim = 0; dim < 2; dim++)
```

Komzpa:
merged https://github.com/postgis/postgis/commit/423b519e1851eb630c57ffa242bf370edfb365ad

PR #610, opened 2021-04-30
added -Z option to prevent tables from being  analyzed

wsdookadr:
Adding a new flag called -Z . When used, this flag will prevent the generation of ANALYZE statements. Without the -Z flag (default behaviour), the ANALYZE statements will be generated.


robe2:
Noticed one more thing missing.  Can you update the docs - 

https://github.com/postgis/postgis/blob/master/doc/using_postgis_dataman.xml#L1435

robe2 (approved):
I tested locally and works as expected. Will commit in a bit.

PR #611, opened 2021-05-04
added tests for -Z switch for shp2pgsql

wsdookadr:


robe2:
It looks good to me but what do I know :)

I'll give @strk one day to comment before I commit.

regress/run_test.pl
```diff
@@ -970,6 +970,14 @@ sub run_loader_and_check_output
 			}
 		}
 
+		if ( $loader_options =~ m{^-Z } ) {
```
strk, line 973: Does this not only match `-Z` if given as the very first option ?

strk (commented):
Looks good to me, but there's no actual test checking for -Z ?

regress/run_test.pl
wsdookadr: Yes. I propose we replace that with `m{^-Z|\s-Z\s}` to cover all the valid locations:

```bash
echo -e "-Z a b\na -Z b\na b -Z\na b c\na-Z b c\na b c-Z" | perl -ne 'print if m{^-Z|\s-Z\s};'
```

output:
```
-Z a b
a -Z b
a b -Z
```

wsdookadr:
Closing in favor of #612

PR #612, opened 2021-05-05
test if -Z works properly by checking pg system catalogs

wsdookadr:


regress/run_test.pl
```diff
@@ -1267,7 +1267,9 @@ sub count_postgis_objects
 sub create_db
 {
 	my $createcmd = "createdb --encoding=UTF-8 --template=template0 --lc-collate=C $DB > $REGRESS_LOG";
-	return system($createcmd);
+    my $rv = system($createcmd);
+    system("psql -d $DB -f disable_autovacuum.sql ");
```
strk, line 1271: You can use --after-create-script switch to run arbitrary scripts. I suggest placing the script in regress/hooks/hook-after-create.sql and add the appropriate switch in core/tests.mk.in to the RUNTESTFLAGS_INTERNAL variable

regress/loader/TestSkipANALYZE-pre.sql
```diff
@@ -0,0 +1 @@
+CREATE SCHEMA pgreg;
```
strk, line 1: Just to reduce changes, how about loading the shapefile in the default schema rather than forcing `pgreg` ? It would let you drop both the pre and the post sql scripts

regress/loader/TestSkipANALYZE.select.sql
```diff
@@ -0,0 +1,5 @@
+-- Checks if the table pgreg.loadedshp was analyzed by
```
strk, line 1: Maybe add a similar check in another test NOT using -Z to verify ANALYZE *is* called ?

wsdookadr: Yes, this is actually a very good idea and make the test suite more complete. Just added a negative test case to check what happens in the absence of -Z.

regress/loader/TestSkipANALYZE-pre.sql
wsdookadr: Yes, I'm definitely for simplification whenever possible. Now the test table is in the default schema (public).

strk:
Looks good to me

robe2:
Remind me again why we aren't just checking the .sql output file or did that prove more difficult than putting in an event trigger?

PR #616, opened 2021-05-18
fix ST_DumpPoints

kalenikaliaksandr:


mwtoews:
Nice find! This bug is from PostGIS 2.1, so worth backporting to maintained branches.

dr-jts:
Ideally this should have an issue on Trac, so it is tracked.

kalenikaliaksandr:
https://trac.osgeo.org/postgis/ticket/4925#ticket

PR #618, opened 2021-06-14
GUC variable postgis.gdal_vsi_options

pramsey:
To set the options used by GDAL connections, instead of setting globals using the ST_GDALOptions() function (now removed). Hopefully should be more flexible and fit better into the PgSQL gestalt.


raster/rt_core/rt_util.c
```diff
@@ -376,13 +377,58 @@ rt_util_gdal_driver_registered(const char *drv) {
...
+				char *key = olist[2*i];
+				char *val = olist[2*i+1];
+				/* GDAL_SKIP is where the disallowed drivers are set */
+				/* We cannot allow user-level over-ride of that config option */
+				if (strcmp(key, "gdal_skip") == 0) {
+					rtwarn("Unable to set GDAL_SKIP config option");
+					continue;
+				}
+				rtinfo("CPLSetConfigOption(%s)", key);
+				CPLSetConfigOption(key, val);
```
rouault, line 427: there might be security implications in letting arbitrary configuration options to be set. I'm for example thinking to the CPL_LOG config option that can be set to a filename, and that will use the filename to put GDAL debug/error messages. So that could potentially be used to corrupt the postgres data files, etc.

pramsey: yeah... kind of annoying... GDAL is so multi-capability that it is a huge footgun and it's hard to tune access in mixed-permission situations. I could make the options into full-on GUC variables and restrict them to super-user, but that actually still wouldn't entirely avoid issues, and would hobble the ability to people to make use of remote-access rasters. If the authentication stuff were open options rather than config options that would be easier.

rouault: This isn't certainly bullet-proof, but perhaps testing that the value isn't a file or directory could avoid the most obvious exploits.
```
#include "cpl_vsi.h"
VSIStatBufL sStat;
if( VSIStatL(val, &sStat) == 0 )
{
   // this is a file or directory...
}
```

rouault: Another possibility if you're interested only in config options that affect /vsi file systems is to use VSIGetFileSystemOptions("/vsis3/")  that returns a string with XML content with the available config options that apply to the file system. You can get the list of file systems with VSIGetFileSystemsPrefixes()

pramsey: So at startup loop through all the /vsi/ systems and get their legal options (which hopefully don't have terrible side effects?) and then only let those be set via the options facility. Sounds safer to me... you think?

pramsey: Along with VSI_CACHE and VSI_CACHE_SIZE :)

rouault: > Sounds safer to me... you think?

likely, but the words of caution of the X/MIT license still apply :-)

pramsey: Would you believe there are unescaped '&' characters in the Option descriptions of vsis3 (and maybe others?) which cause libxml2 to choke on parsing the Options XML...

rouault: > Would you believe there are unescaped '&' characters in the Option descriptions of vsis3 (and maybe others?) which cause libxml2 to choke on parsing the Options XML...

ah that's unfortunate. I see we use the CPL MiniXML parser to validate them and it is quite lax. You could possibly use it (cpl_minixml.h), but I'll fix that

pramsey: minixml did the trick, thanks

raster/rt_pg/rtpostgis.c
```diff
@@ -239,12 +241,215 @@ rt_pg_debug(const char *fmt, va_list ap)
...
+/*
+* For older versions of GDAL we  have extracted the list of options
+* that were available at the 2.2 release and use that
+* as our set of allowed VSI network file options.
+*/
+static void
+rt_pg_vsi_load_all_options(void)
+{
+	const char *gdaloption;
+	const char *gdaloptions[] = {
```
rouault, line 276: For the array to be fully read-only:
```suggestion
	const char * const gdaloptions[] = {
```

raster/rt_pg/rtpostgis.c
```diff
@@ -239,12 +241,215 @@ rt_pg_debug(const char *fmt, va_list ap)
...
+{
+	const char *vsiname;
+	char *vsilist[] = {
+		"/vsicurl/",
+		"/vsis3/",
+		"/vsigs/",
+		"/vsiaz/",
+		"/vsioss/",
+		"/vsihdfs/",
+		"/vsiwebhdfs/",
```
rouault, line 394:
```suggestion
		"/vsiwebhdfs/",
		"/vsiswift/",
		"/vsiadls/",
```

raster/rt_pg/rtpostgis.c
```diff
@@ -239,12 +241,215 @@ rt_pg_debug(const char *fmt, va_list ap)
...
+		optNode = optNode->psNext;
+	}
+	CPLDestroyXMLNode(root);
+}
+
+static void
+rt_pg_vsi_load_all_options(void)
+{
+	const char *vsiname;
+	char *vsilist[] = {
```
rouault, line 387:
```suggestion
	const char * const vsilist[] = {
```

raster/rt_pg/rtpostgis.c
```diff
@@ -239,12 +241,215 @@ rt_pg_debug(const char *fmt, va_list ap)
...
+		"/vsicurl/",
+		"/vsis3/",
+		"/vsigs/",
+		"/vsiaz/",
+		"/vsioss/",
+		"/vsihdfs/",
+		"/vsiwebhdfs/",
+		NULL
+	};
+	char **vsilistptr = vsilist;
```
rouault, line 397:
```suggestion
	const char * const *vsilistptr = vsilist;
```

PR #623, opened 2021-07-23
Upgrade path for address_standardizier_data_us and fix for upgrades as non-superuser via pgextwlist

jankatins:
* address_standardizier_data_us: provide upgrade path scripts: for address standardizer_data_us so this extension can be upgraded and does not need to be dropped/recreated
* Convert _postgis_upgrade_info to a function -> we use pgextwlist and that sets a security context which prevents creating temporary tables (see e.g. https://github.com/dimitri/pgextwlist/issues/43)

### DIFF
--- a/extensions/address_standardizer/Makefile.in
+++ b/extensions/address_standardizer/Makefile.in
@@ -47,7 +47,7 @@ DATA_built = \
 
 all:  sql/address_standardizer.sql sql/address_standardizer--1.0--$(EXTVERSION).sql sql/$(EXTENSION)--$(EXTVERSION).sql sql/$(EXTENSION)--ANY--$(EXTVERSION).sql \
  sql/$(EXTENSION)_data_us.sql sql/address_standardizer--@POSTGIS_LIB_VERSION@.sql sql/$(EXTENSION)_data_us--@POSTGIS_LIB_VERSION@.sql  sql/$(EXTENSION)_data_us--$(EXTVERSION)--$(EXTVERSION)next.sql \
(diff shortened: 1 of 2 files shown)


robe2:
I haven't had a chance to look at this.  I'll take a look this weekend.

robe2:
Related tickets:

https://trac.osgeo.org/postgis/ticket/4974

I thought I had  a ticket already for the _postgis_upgrade_info as that was a complaint of an Amazon RDS user too.  But they worked around the issue, so maybe I didn't bother putting in a ticket for it before. Related one:

https://trac.osgeo.org/postgis/ticket/4975

jankatins:
for completeness: 
* I think this is the RDS problem you mention (on the mailinglist): https://www.mail-archive.com/postgis-users@lists.osgeo.org/msg08353.html
* Heroku also seems to have run into this error: https://stackoverflow.com/questions/54213361/update-postgis-extensions-on-heroku + https://trac.osgeo.org/postgis/ticket/4186 mention this error message

robe2:
> 
> 
> for completeness:
> 
>     * I think this is the RDS problem you mention (on the mailinglist): https://www.mail-archive.com/postgis-users@lists.osgeo.org/msg08353.html
> 
>     * Heroku also seems to have run into this error: https://stackoverflow.com/questions/54213361/update-postgis-extensions-on-heroku + https://trac.osgeo.org/postgis/ticket/4186 mention this error message

More completeness:

https://lists.osgeo.org/pipermail/postgis-users/2020-May/044225.html

PR #626, opened 2021-08-04
fix lwgeom_boundary bugs: #4956, #4960, #4961

kalenikaliaksandr:
I am continuing @strk [work](https://gitlab.com/postgis/postgis/-/merge_requests/50) on fixing ST_Boundary issues introduced by my rewrite.

to do:
- [x] fix https://trac.osgeo.org/postgis/ticket/4956
- [x] fix https://trac.osgeo.org/postgis/ticket/4960
- [x] fix https://trac.osgeo.org/postgis/ticket/4961
- [x] make sure there are no memory leaks
- [x] increase boundary unit tests coverage
- [x] increase boundary integration tests coverage


liblwgeom/lwgeom.c
```diff
@@ -2550,15 +2550,13 @@ lwgeom_boundary(LWGEOM *lwgeom)
...
-	}
+	case POINTTYPE:
 	case MULTIPOINTTYPE: {
-		return (LWGEOM *)lwmpoint_construct_empty(srid, hasz, hasm);
+		return lwgeom_construct_empty(lwgeom->type, srid, hasz, hasm);
 	}
 	case LINETYPE:
 	case CIRCSTRINGTYPE: {
-		if (lwgeom_is_closed(lwgeom))
+		if (lwgeom_is_closed(lwgeom) || lwgeom_is_empty(lwgeom))
```
strk, line 2559: This is where the BoundaryNodeRule would possibly give a different answer (with some BoundaryNodeRule also the closed lines are considered to have a boundary, I think)

strk: But I do realize this is out of scope for this work, so feel free to ignore :)

strk:
Please be descriptive in your commit logs, try as much as possible to allow understanding what the commit does w/out having to go to external sources. References to trac tickets are important but should NOT be the *only* information, just an auxiliary one, and done with proper syntax (ie: "Fixes #xxx" or "References #xxx" or "Closes #xxx").

Good luck with memory management !

HINT: in addition to the unit tests (which are great to spot memory leaks and memory violations) a regress test will be useful to compare results again older postgis versions.

kalenikaliaksandr:
> Good luck with memory management !

I found out that leak was introduced by the use of lwgeom_homogenize for collections. I had to make lwgeom_homogenize to return deep clone of all geometries so original lwgeom can be freed not breaking homogenized one. I can't say I am happy with the way I handled that but the other way would be iterating through not homogenised collection to find out what geometries are not needed anymore in homogenized one and can be freed.

strk:
We used to have lwgeom_release to drop the "arounds" but keep the
geoms, but it was refactored by @pramsey long time ago, probably it
was never done correctly, ar at least cleanly.

A cunit test for lwgeom_homogenize would be nice too

kalenikaliaksandr:
I fixed memory leak without making lwgeom_homogenize to return deep clone of geometries but had to introduce new function called lwgeom_release_deep -- it iterates through all underlying geometries and frees everything but serialized_pointlist. Looks like someone already thought about introducing such function since it mentioned in https://github.com/postgis/postgis/blob/177de9ca1c1f4fa6cc57271c5810bcc26ebade84/postgis/TODO#L3

liblwgeom/lwgeom.c
```diff
@@ -465,6 +465,76 @@ lwgeom_release(LWGEOM *lwgeom)
 
 }
 
+/* free lwgeom and all underlying objects except serialized_pointlist */
```
strk, line 468: It looks like it's not just the serialized_pointlist but possibly more things ?
For once I see the array of pointers to underlying geometries in the COLLECTION case, and then I wonder about the BBOX cache, sometimes it points to underlying buffer from serialized version and sometimes it's created anew, I think such difference is encoded in flags somehow.

kalenikaliaksandr: lwgeom_release deep leaves serialized_pointlist not freed, everything else including array of pointers is freed.
bbox cache would have to be handled correctly by lwgeom_free, right? I don't see anything specific about bbox in lwgeom_free so I made release_deep to just free memory if pointer is not null.

the same purpose for which I introduced lwgeom_release_deep can be achived by lwgeom_free but then readonly flag should be set to 1 before a call so serialized_pointlist won't be freed.
https://github.com/postgis/postgis/blob/177de9ca1c1f4fa6cc57271c5810bcc26ebade84/liblwgeom/ptarray.c#L323
not sure if it is a better way to go.

strk:
I'd just like the comment to be as much as possibly adhering with the
actual code. And for bbox it may be a good idea to make sure tests do
cover the case of readonly BBOXes (kind of an hack to do, I guess,
from cunit, as in reality the only case in which BBOXes are read-only
is when they are read from SERIALIZED form).

kalenikaliaksandr:
I wrote more tests for GEOMETRYCOLLECTION and discovered more memory leaks... I decided to roll back to original decision of making homogenize return deep clone of geometries and since homogenize already has good unit test coverage I am feeling pretty safe about doing that and not breaking anything.

liblwgeom/cunit/cu_boundary.c
```diff
@@ -0,0 +1,211 @@
...
+	PG_ADD_TEST(suite,boundary_point);
+	PG_ADD_TEST(suite,boundary_line);
+	PG_ADD_TEST(suite,boundary_polygon);
+	PG_ADD_TEST(suite,boundary_triangle);
+	PG_ADD_TEST(suite,boundary_tin);
+	PG_ADD_TEST(suite,boundary_collection);
+#if 0 /* write these */
+	PG_ADD_TEST(suite,boundary_tin);
+#endif
+	//PG_ADD_TEST(suite,boundary_collection);
```
strk, line 210: These 4 lines are to be removed

liblwgeom/lwgeom.c
```diff
@@ -2623,22 +2621,17 @@ lwgeom_boundary(LWGEOM *lwgeom)
...
-			for (uint32_t i = 0; i < lwpoly->nrings; i++)
-			{
-				POINTARRAY *ring = ptarray_clone_deep(lwpoly->rings[i]);
-				lwmline_add_lwline(lwmline, lwline_construct(srid, 0, ring));
-			}
 
-			return (LWGEOM *)lwmline;
-		}
+		LWGEOM *lwout = lwgeom_homogenize((LWGEOM *)lwmline);
+		lwgeom_free((LWGEOM *)lwmline);
```
strk, line 2632: Is homogenize really needed here ? We don't expect to be able to construct non-homogenous things via lwmline_add_lwline, are we ? Don't we get an exception if we tried ?

strk (commented):
Left a couple of comments, could you also please rebase to current main/master branch ?

robe2:
Is this ready for commit or more to do here?

liblwgeom/lwgeom.c
Komzpa: I've checked - it is converting single-ring polygons to non-multi linestring. Added a comment on merge.

PR #628, opened 2021-08-23
ST_StartPoint support any geometry

kalenikaliaksandr:
https://trac.osgeo.org/postgis/ticket/4981#ticket


robe2:
If the use case is to get the fastest point possible from a geometry, wouldn't it make sense to use ST_PointOnSurface instead which already does that.

PR #642, opened 2021-09-21
ST_LocateBetweenElevations: fix segfault on interrupt

kalenikaliaksandr:
pointarray returned by `ptarray_clamp_to_ordinate_range` need to be null checked before dereferencing since when request is interrupted, null is returned here https://github.com/postgis/postgis/blob/5476f08d1eafe2807b206ac95197d7ce46d20549/liblwgeom/lwlinearreferencing.c#L522 and segmentation fault happens.

### DIFF
--- a/liblwgeom/lwlinearreferencing.c
+++ b/liblwgeom/lwlinearreferencing.c
@@ -736,6 +736,9 @@ lwpoly_clip_to_ordinate_range(const LWPOLY *poly, char ordinate, double from, do
 		/* Ret number of points */
 		POINTARRAY *pa = ptarray_clamp_to_ordinate_range(poly->rings[i], ordinate, from, to, LW_TRUE);
 
+		if (!pa)
+			return NULL;
+
 		if (pa->npoints >= 4)
 			lwpoly_add_ring(poly_res, pa);
 		else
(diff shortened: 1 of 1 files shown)


liblwgeom/lwlinearreferencing.c
```diff
@@ -736,6 +736,7 @@ lwpoly_clip_to_ordinate_range(const LWPOLY *poly, char ordinate, double from, do
 		/* Ret number of points */
 		POINTARRAY *pa = ptarray_clamp_to_ordinate_range(poly->rings[i], ordinate, from, to, LW_TRUE);
 
+		assert(pa);
```
Komzpa, line 739: assert will crash it on the spot here instead of the other spot. does not look like a lot of change?

Komzpa: it has to be more like `if (!pa) return NULL;` in whole call chain and `lwerror` on the very upper level. 

or, why don't we lwerror right in LW_ON_INTERRUPT? it's not like the calculation is going to get finished anyway.

kalenikaliaksandr: agree about assert -- replaced with `if (!pa) return NULL;`
`lwgeom_clip_to_ordinate_range` will handle NULL here.
https://github.com/postgis/postgis/blob/5476f08d1eafe2807b206ac95197d7ce46d20549/liblwgeom/lwlinearreferencing.c#L876-L877

PR #644, opened 2021-10-21
Clarify ST_ShiftLongitude docs

LKajan:
The ST_ShiftLongitude function behave symmetrical with inverse and reverse transformations. Clarify this in the docs.

resolves #5010
https://trac.osgeo.org/postgis/ticket/5010

### DIFF
--- a/doc/reference_editor.xml
+++ b/doc/reference_editor.xml
@@ -1769,20 +1769,28 @@ LINESTRING(0 0,1 1,0 0,3 3,4 4)
       <refsection>
         <title>Examples</title>
 
-        <programlisting>--forward transform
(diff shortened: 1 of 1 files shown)


robe2:
@LKajan can you make changes based on @pramsey input or comment.  Then we can accept this.

LKajan:
Thanks @pramsey for your comments!
I changed back to using term shift. I trust your view here being a native speaker.
My reasoning to change the term was that I see we have here just two different representations of a same coordinate point and we just swap the representations back and forth.

In the project do you like more amended commits and force push to the pr or separate commits and squash before merge? I now added these fixes as separate commits.

PR #643, opened 2021-09-27
Add cirrus_ci for FreeBSD

lbartoletti:
@robe2 [Cirrus ci] is a service that provides similar services as travis, GH action and many others BUT with [FreeBSD support](https://cirrus-ci.org/guide/FreeBSD/)

As I'm the postgis maintainer on FreeBSD, I have several scripts to test and deploy releases. For postgis, we have already bessie/bessie32 as a CI, but in this CI, I use latest pkg. I init this CI for myself, but if it can be useful for the project feel free to adopt it, or I keep it for me :)

Regards

PR #646, opened 2021-10-28
Add cirrus-ci badge

lbartoletti:
Folluw up #643


robe2:
I had started setting it up but couldn't find where the badges are to put up.  So I'm not sure I set it up right or not as I didn't see any runs.  But I wouldn't have seen runs since nothing was committed since.  I'll push this one and see if it fixes things.

robe2:
@lbartoletti  Hey it must be working as I  see a freebsd13-amd64 and freebsd12-amd64 failing. 

Can you look at these when you have the chance.  I'll commit anyway  https://github.com/postgis/postgis/pull/646/checks?check_run_id=4030878164

PR #647, opened 2021-11-22
lwgeom_wagyu: replace wagyu's quick_lr_clip with lwgeom_clip_to_ordinate_range

kalenikaliaksandr:
Replacement of wagyu's quick_lr_clip with lwgeom_clip_to_ordinate_range gives a little performance improvement. Although both functions are intended to do the same thing, performance improvement is explained by clipping geometry before making transformation into format compatible with wagyu.


deps/wagyu/lwgeom_wagyu.cpp
```diff
@@ -62,62 +54,48 @@ ptarray_to_wglinearring(const POINTARRAY *pa, const wagyu_box &tile)
...
-	/* All points are inside the box */
 	return lr;
 }
 
 static vwpolygon
-lwpoly_to_vwgpoly(const LWPOLY *geom, const wagyu_box &box)
+lwpoly_to_vwgpoly(const LWPOLY *geom, const GBOX *box)
 {
+	LWCOLLECTION *geom_clipped_x = lwgeom_clip_to_ordinate_range((LWGEOM *)geom, 'X', box->xmin, box->xmax, 0);
+	LWCOLLECTION *geom_clipped = lwgeom_clip_to_ordinate_range((LWGEOM *)geom_clipped_x, 'Y', box->ymin, box->ymax, 0);
```
Komzpa, line 68: can you free the x clipped one right away?

mngr777 (approved):

pramsey:
Would like to see @Algunenano opinion on this. Maintaining valid outputs from MVT processing is just as important as performance. Not sure if our test chain has everything it needs to ensure we are still doing as good a job generating quality outputs.

Algunenano:
Changing how the clipping is done should be ok (famous last words) but I don't see any benchmarks. Is `lwgeom_clip_to_ordinate_range` faster than `quick_lr_clip` in all cases or only some?

Komzpa:
Benchmark anecdata: with this patch the layer that was tiled in 63 hours now tiles in 49 hours, and a subpart that used to render in 4m50s now takes 4m20s. No visual difference in results.

PR #648, opened 2021-11-25
Pixel as centroids optimization

mngr777:
- `ST_PixelAsCentroids`, `ST_PixelAsCentroid` rewritten to use new C function  `RASTER_getPixelCentroids` (mostly copied from `RASTER_getPixelPolygons`).
- Test updated to add a case for rotated raster.

`ST_PixelAsCentroids(rast, 1)` execution time comparison for some sizes of raster:
| Raster size | Exec. time before | Exec. time after |
|-------------|-------------------|------------------|
| 100x100 | 44.677 ms | 4.228 ms |
| 200x200 | 183.234 ms | 17.441 ms |
| 500x500 | 1144.930 ms | 141.880 ms
| 1000x1000 | 4630.656 ms | 573.416 ms
| 2000x2000 | 18302.578 ms | 2290.872 ms
| 5000x5000 | 115014.318 ms | 14121.408 ms

(On Intel Core i7-11370H @ 3.30GHz × 8, 16 Gb RAM)


raster/rt_pg/rtpg_geometry.c
```diff
@@ -651,6 +654,324 @@ Datum RASTER_getPixelPolygons(PG_FUNCTION_ARGS)
...
+
+					MemoryContextSwitchTo(oldcontext);
+					elog(ERROR, "RASTER_getPixelCentroids: Could not get pixel centroid");
+					SRF_RETURN_DONE(funcctx);
+				}
+
+				/* allocate space for new point */
+				if (!pixcount) {
+					pix = palloc(sizeof(struct rt_pixel_t) * (pixcount + 1));
+				} {
```
Komzpa, line 849: this anonymous block looks sus. repalloc right after palloc?

raster/rt_pg/rtpg_geometry.c
```diff
@@ -651,6 +654,324 @@ Datum RASTER_getPixelPolygons(PG_FUNCTION_ARGS)
...
+				errcode(ERRCODE_OUT_OF_MEMORY),
+				errmsg("Could not deserialize raster")
+			));
+			MemoryContextSwitchTo(oldcontext);
+			SRF_RETURN_DONE(funcctx);
+		}
+
+		/* raster empty, return NULL */
+		if (rt_raster_is_empty(raster)) {
+			elog(NOTICE, "Raster is empty. Returning NULL");
```
Komzpa, line 747: technically it's not a null but an empty set?

Komzpa (changes requested):
Could you please also add a line about this change into NEWS file that will credit it properly? Thanks.

raster/rt_pg/rtpg_geometry.c
mngr777: Fixed. Good thing you caught it.

raster/rt_pg/rtpg_geometry.c
mngr777: This was copied from `RASTER_getPixelPolygons`, `rtpg_geometry.c:440` and a couple more, should I change these to "Returning empy set" too?

Komzpa: yes, either fix or just erase this debug print, if it was wrong for years and nobody noticed there's no value in it.

mngr777: Done.

PR #652, opened 2021-12-08
 Avoid "no geometry column found" in ST_AsMVT when geometry columns contain upprecase characters

jimjonesbr:
The function `ST_AsMVT` won't find the geometry column if the column name uses uppercase characters. The following query fails, as the alias of `ST_AsMVTGeom` in the CTE is called `MVTGeom`

```sql
WITH j AS (
 SELECT ST_AsMVTGeom(geom, ST_MakeBox2D(ST_Point(0, 0), ST_Point(4096,4096))) AS MVTGeom FROM t
)
SELECT ST_AsMVT(j,'foo',4096,'MVTGeom') FROM j;

ERROR:  parse_column_keys: no geometry column found
```

Despite the error message, the geometry column is there, but can only be found with lowercase:

```sql
WITH j AS (
 SELECT ST_AsMVTGeom(geom, ST_MakeBox2D(ST_Point(0, 0), ST_Point(4096,4096))) AS MVTGeom FROM t
)
SELECT ST_AsMVT(j,'foo',4096,'mvtgeom') FROM j;
                            st_asmvt                            
----------------------------------------------------------------
 \x1a1c0a03666f6f12101803220c0914d83f1a14272800133c0f2880207802
 ```
 
 By using `strcasecmp` instead of `strcmp` I intend to achieve the following behaviour: 
 
  ```sql
 WITH j AS (
 SELECT ST_AsMVTGeom(geom, ST_MakeBox2D(ST_Point(0, 0), ST_Point(4096,4096))) AS MVTGeom
 FROM t
)
SELECT ST_AsMVT(j,'foo',4096,'MVTGeom') FROM j;
                            st_asmvt                            
----------------------------------------------------------------
 \x1a1c0a03666f6f12101803220c0914d83f1a14272800133c0f2880207802
```
 
```sql 
WITH j AS (
 SELECT ST_AsMVTGeom(geom, ST_MakeBox2D(ST_Point(0, 0), ST_Point(4096,4096))) AS "MVT Geom"
 FROM t
)
SELECT ST_AsMVT(j,'foo',4096,'mvt GEOM') FROM j;
                            st_asmvt                            
----------------------------------------------------------------
 \x1a1c0a03666f6f12101803220c0914d83f1a14272800133c0f2880207802

```


Komzpa:
The initial assumption is wrong - it works as expected. The only thing you did not account for is that for postgres not to auto-lowercase the identifier it has to be quoted:

```
WITH j AS (
 SELECT ST_AsMVTGeom(geom, ST_MakeBox2D(ST_Point(0, 0), ST_Point(4096,4096))) AS "MVTGeom" FROM t
)
SELECT ST_AsMVT(j,'foo',4096,'MVTGeom') FROM j;
```

jimjonesbr:
Hi @Komzpa, I agree that it works as expected - as PostgreSQL auto-lowercase the column names, but I thought it would be a nice addition to make queries work despite this behavior. Of course, if it does not break anything else :)

Komzpa:
I would say I rather merge a documentation patch that will make it easier for user of ST_AsMVTGeom to learn about how postgres handles capitalization, rather than merging this behavior.

Komzpa (changes requested):
- diff went horribly wrong (do a fresh branch?)

 - change to be only a documentation change pointing to postgres' caps handling in ST_AsMVT docs.

jimjonesbr:
Sorry, ignore my review request.. I clicked on the wrong button.

PR #664, opened 2022-02-11
Build spatial_ref_sys.sql from a data file 

pramsey:
And also build an upgrade file that is included in the extension upgrade script.

### DIFF
--- a/utils/create_spatial_ref_sys.pl
+++ b/utils/create_spatial_ref_sys.pl
@@ -1,0 +1,58 @@
+#!/usr/bin/env perl
+
+#
+# PostGIS - Spatial Types for PostgreSQL
+# http://postgis.net
+#
+# Copyright (C) 2022 Paul Ramsey <pramsey@cleverelephant.ca>
+#
+# This is free software; you can redistribute and/or modify it under
+# the terms of the GNU General Public Licence. See the COPYING file.
+#
+
+use warnings;
+use strict;
+
+(@ARGV) ||
+	die "Usage: perl $0 -u <spatial_ref_sys>\n" .
+		"Creates an SQL script load or update records of spatial_ref_sys.\n";
+
+my $UPDATE = 0;
+my $SRSFILE = 0;
+my $SRSTABLE = "spatial_ref_sys";
+
+for (my $i=0; $i<@ARGV; $i++)
+{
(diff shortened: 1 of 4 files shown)


strk:
What's the rationale for the upgrade SQL script being different from the create SQL script ? Wouldn't your proposed upgrade script behave exactly as the create one for a starting empty table ?

utils/create_spatial_ref_sys.pl
```diff
@@ -0,0 +1,57 @@
...
+	chop;
+	my ($srid, $auth_name, $auth_srid, $srtext, $proj4text) = split (/\t/);
+	$auth_name =~ s/\'/\'\'/g;
+	$srtext =~ s/\'/\'\'/g;
+	$proj4text =~ s/\'/\'\'/g;
+	print "INSERT INTO \"$SRSTABLE\" (\"srid\", \"auth_name\", \"auth_srid\", \"srtext\", \"proj4text\") VALUES ($srid, '$auth_name', $auth_srid, '$srtext', '$proj4text')";
+	if ($UPDATE) {
+		print "\nON CONFLICT (srid) DO NOTHING";
+	}
+	print ";\n\n";
```
strk, line 53: Do you see any problem in doing this in a single statement, with iteration only adding a new `(...)` record ?
Would `ON CONFLICT` not work correctly in that case ?

strk:
Please make sure builds are still supported out of source tree (Dronie should actually be doing that, if it is running for this PR).

To recap: this change would basically ADD new SRIDs shipped by PostGIS but only if those SRIDs are not already assigned by the user, and would NEVER update existing records (in case, like it happened between some PostGIS versions, they are changed). This change will result in a bigger postgis upgrade script (by how much, roughly?) and a consequent slower upgrade procedure (by how much, roughtly ?)

pramsey:
Superceded by #665

utils/create_spatial_ref_sys.pl
pramsey: New PR does a multi-valued insert, and the on conflict works, so yay!

pramsey:
> What's the rationale for the upgrade SQL script being different from the create SQL script ? Wouldn't your proposed upgrade script behave exactly as the create one for a starting empty table ?

You're right, an upsert works for both cases, so I've ditched the machinery and done with a minimal patch in #665

PR #667, opened 2022-02-17
Minor changes to test cases

jimjonesbr:
As pointed out by @strk , the tests were failing on 32bit architecture. 
This PR adds a missing `ST_AsText(geom,5)` in the input tests and also adds an `ORDER BY` to the output queries.


strk:
As the only CI failing was gitlab-ci this PR would be best created over there: gitlab.com/postgis/postgis

Is it normal for this PR to contain 85 commits ?

jimjonesbr:
> As the only CI failing was gitlab-ci this PR would be best created over there: gitlab.com/postgis/postgis

Is it ok to have two (slightly) distinct regress tests in github and gitlab?

> Is it normal for this PR to contain 85 commits ?

I guess it contains the 80+ commits prior to this last PR, since Paul Ramsey closed my first PR and copied the content to master himself.

PR #672, opened 2022-02-24
Allowing archiver program and flags options in ./configure

mngr777:
This change allows to build PostGIS with `-flto` flag using clang by specifying `llvm-ar` as archiver.
```
./configure CFLAGS='-flto' LDFLAGS='-flto -L/usr/local/lib' AR='llvm-ar' RANLIB='llvm-ranlib'
```
Tested on FreeBSD 13.0. (`-L/usr/local/lib` is required to link against `libiconv`, which is a separate issue.)


lbartoletti (approved):

configure.ac
```diff
@@ -26,6 +26,8 @@ AC_CONFIG_MACRO_DIR([macros])
 AC_CONFIG_AUX_DIR([build-aux])
 AC_PROG_INSTALL
 
+# Set default AR_FLAGS before libtool does
+: ${AR_FLAGS=rs}
```
strk, line 30: What's this line doing ? Sounds like a no-op to me ?

mngr777: `${AR_FLAGS=rs}` sets AR_FLAGS to 'rs' if not set.
`:` is a no-op (but it's arguments are still interpreted) and it prevents the shell from interpreting the result of assignment. E.g. it is used in this way in libtool.m4: https://github.com/autotools-mirror/libtool/blob/f5eb6f11698c65e550bfdcc1a0a1943df6d70f28/m4/libtool.m4#L145, https://github.com/autotools-mirror/libtool/blob/f5eb6f11698c65e550bfdcc1a0a1943df6d70f28/m4/libtool.m4#L1493

strk:
Thanks, this was merged as 62e5bf1177595eee23190fc4be4ad6011778f0c7

PR #673, opened 2022-02-25
Marc21 regress

jimjonesbr:
Reduce decimal precision to address a testing issue reported on [#5104](https://trac.osgeo.org/postgis/ticket/5104)


robe2:
Okay looks like this patch did fix regress on winnie.  Still running thru paces, but she already past this part.  So must be some junk on my system.

PR #675, opened 2022-03-08
Remove GEOS geometry validation from RT_ST_DumpAsPolygons

kalenikaliaksandr:
I would like to remove geometry validation from RT_ST_DumpAsPolygons and I see several reasons to do that:
- Removing GEOS validation and geometry fix calls makes RT_ST_DumpAsPolygons work 2.5x-3x faster for my data
- Since time this function was written GDAL seem to be improved and do better job at producing valid polygons

Also there will still be an option to call ST_MakeValid on output but performance of RT_ST_DumpAsPolygons itself will be better.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,8 @@ PostGIS 3.3.0dev
   - Drop support for PostgreSQL 9.6 and 10 (Regina Obe)
   - Change output for WKT MULTIPOINT. All points now
     wrapped in parens. (Even Roualt)
+  - GH674, geometry validation and fixing is disabled for ST_DumpAsPolygons
+    so it works faster but might produce invalid polygons.
 
  * Enhancements *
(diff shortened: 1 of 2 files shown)


Komzpa (approved):

robe2:
I think 2.5x to 3x is enough reason to remove it.  Maybe just add a note it could result in invalid and to run makevalid to be safe. For my purposes I'd much prefer an invalid output but faster output.

To @pramsey point 3.10 is faster, though it seems all the systems I see are running 3.9 at best and often 3.8.  People tend to upgrade their PostGIS  / PostgreSQL more frequently than GEOS.  It's sad but reality.

robe2 (approved):

robe2:
regression - https://trac.osgeo.org/postgis/ticket/5136

PR #677, opened 2022-03-08
[cirrus-ci] use sudo

strk:
Trying to avoid a timeout


lbartoletti:
You could up the timeout delay with:

```
task: 
  timeout_in: 90m
  ...
```

strk:
The thing is that it should not take 1h to `createdb`, that's what we need to find the cause of

strk:
See https://cirrus-ci.com/task/6474763102060544 -- 55 minutes waiting here:
```
sudo -u postgres createuser -s `whoami`
Shell session I/O error: context deadline exceeded
```

strk:
It looks like CI is green now, so I guess nothing is needed anymore here

PR #678, opened 2022-03-14
LTO enabled by default if supported; selecting ar, ranlib implementat…

mngr777:
…ion based on compiler vendor

Selecting `ar` (`llvm-ar`, `gcc-ar`) and `ranlib` (`llvg-ranlib`, `gcc-ranlib`) matching selected compiler for gcc and Clang.
This allows to enable LTO by default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,8 +26,15 @@ AC_CONFIG_MACRO_DIR([macros])
 AC_CONFIG_AUX_DIR([build-aux])
 AC_PROG_INSTALL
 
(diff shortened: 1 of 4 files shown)


mngr777:
Failing tests seem to be due to missing GDAL (not sure about "CI (pg13-geos39-gdal31-proj71, usan_gcc)" https://github.com/postgis/postgis/runs/5552651399?check_suite_focus=true).
Build does not start on FreeBSD test servers.

PR #683, opened 2022-04-12
aggregate ST_3DUnion

mngr777:
Adding aggregate ST_3DUnion on postgis_sfcgal extension creation; documentation update.


robe2:
I think this should read 3.3.0 instead of 3.0.0.  Probably better to also just have it on it's own line

``+		<para>Availability: 2.2.0, aggregate variant: 3.0.0</para>``

like


``+		<para>Availability: 3.3.0 aggregate variant was added</para>``

robe2:
Added trac ticket here - https://trac.osgeo.org/postgis/ticket/5132  and testing out on windows before merge.

robe2:
Oh wait just realized we are missing tests for this.  Can you add a test to sfcgal/regress/regress_sfcgal.sql

PR #684, opened 2022-04-13
enable 2d sortsupport function for postgres >= 15

mngr777:
PostgreSQL 15 uses opclass' `picksplit` function to find better partitioning for index records when building GiST index using sorted build method. This produces indices performing similarly to those built by default method while reducing build time by approx. 40%.

Previously `sortsupport` function had to be removed from `gist_geometry_ops_2d` opclass due to performance degradation: https://lists.osgeo.org/pipermail/postgis-devel/2021-November/029225.html

Postgres patch addressing this issue: https://commitfest.postgresql.org/36/3488/, please see Emails link for more details and benchmarking results.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -24,6 +24,7 @@ PostGIS 3.3.0dev
   - ST_ConcaveHull GEOS 3.11+ native implementation (Paul Ramsey, Martin Davis)
   - #4574, GH678, #5121 Enable Link-Time Optimizations using --enable-lto (Sergei Shoulbakov)
   - GH676, faster ST_Clip (Aliaksandr Kalenik)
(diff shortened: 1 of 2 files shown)


robe2:
@mngr777 @Komzpa for some reason I thought this was already committed though I don't see mention of it in the news.

Komzpa (approved):
This was disabled by Paul before release of 3.2, but now it can be switched on I believe with changes in PG15. Benchmarks are on the mailing list, there are improvements over status quo.

(NEWS entry needed indeed)

robe2:
@Komzpa  OK.  I'll add to the news and test in my performance tests to confirm the speed improvement.

PR #686, opened 2022-05-10
raster alignment check fix, #5148

mngr777:
Fixes #5148 "ST_Clip ERROR: rt_raster_from_two_rasters: The two rasters provided do not have the same alignment"


raster/rt_core/rt_raster.c
```diff
@@ -3072,7 +3072,7 @@ rt_raster_gdal_rasterize(
 			double _w[2] = {0};
 
 			/* raster is already aligned */
-			if (FLT_EQ(*grid_xw, extent.UpperLeftX) && FLT_EQ(*grid_yw, extent.UpperLeftY)) {
```
Komzpa, line 3075: may it happen that really here the `=` is needed?

I am concerned about how this plays out with FLT_NEQ in 3118 and 3146, it looks like it they expect to use negation of same comparator?

Komzpa (approved):

raster/rt_core/rt_raster.c
mngr777: `DBL_EQ` seems strict enough so that raster may be considered aligned for all purposes, while ignoring insignificant rounding errors.

I mostly tried to change as little as possible. If `DBL_NEQ` is used in 3118 and 3146 then for those cases where difference between raster and grid coordinate is in `(DBL_EPSILON, FLT_EPSILON)` (where currently alignment is skipped) the raster will grow by 1 pixel (when called from `RASTER_clip`) or be rescaled. E.g in my artificial test case for 2 rectangles `(0 1, 1 1, 1 0, 0 0)` and `(0 0.9.., 1 0.9..., 1 0.9, 0 9.9)`, scale = `0.01`, the raster returned by `rt_raster_gdal_rasterize` will be 10x11 instead of 10x10.
The alignment check here is basically the same as in `rt_raster_same_alignment` and if uses `FLT_EQ`.

kalenikaliaksandr (approved):

PR #687, opened 2022-05-15
Revert "Read spatial_ref_sys data directly out from proj (WIP)"

Komzpa:
Reverts postgis/postgis#668

I typoed and merged it instead of postgis/postgis#686, sorry everyone :(


robe2:
Is there a way to reopen this pull request or does @pramsey just have to create another one?

Komzpa:
https://stackoverflow.com/posts/12838195/revisions says that create another one is the only way alas :(

PR #688, opened 2022-05-27
ST_LineMerge overload with "directed" argument.

mngr777:
Closes #5161


robe2:
the function has to exist even if < GEOS 3.11 so put the check in the C side. Look at ST_FreschetDistance for example.

mngr777:
The check in C code is in `lwgeom_linemerge_directed`, similar to [`lwgeom_unaryunion_prec`](https://github.com/postgis/postgis/blob/058fd88a1c5639226e2f06dda66799e85aa80011/liblwgeom/lwgeom_geos.c#L806): error is thrown when `directed` is TRUE and GEOS < 3.11.

With GEOS 3.10.2:
```
# SELECT ST_AsText(ST_LineMerge('MULTILINESTRING((60 30, 10 70), (120 50, 60 30), (120 50, 180 30))', false));
               st_astext               
---------------------------------------
 LINESTRING(180 30,120 50,60 30,10 70)
(1 row)

# SELECT ST_AsText(ST_LineMerge('MULTILINESTRING((60 30, 10 70), (120 50, 60 30), (120 50, 180 30))', true));
ERROR:  Directed line merging requires GEOS-3.11 or higher
```

PR #689, opened 2022-06-07
Add ST_SimplifyPolygonHull() with GEOS 3.11

pramsey:
Add `ST_SimplifyPolygonHull(geom geometry, vertex_fraction float8, is_outer boolean DEFAULT true)`.

References https://trac.osgeo.org/postgis/ticket/5169#ticket

See also:
* http://lin-ear-th-inking.blogspot.com/2022/04/outer-and-inner-concave-polygon-hulls.html
* http://lin-ear-th-inking.blogspot.com/2022/05/using-outer-hulls-for-smoothing.html


doc/reference_processing.xml
```diff
@@ -2403,6 +2403,58 @@ FROM (SELECT  'POLYGON((0 0, 8 8, 0 16, 0 0))'::geometry geom) As foo;
           </refsection>
     </refentry>
 
+
+    <refentry id="ST_SimplifyPolygonal">
+      <refnamediv>
+        <refname>ST_SimplifyPolygonal</refname>
+        <refpurpose>Computes a boundary-respecting hull of a polygonal geometry, with hull shape determined by a target parameter.</refpurpose>
```
Komzpa, line 2410: I do not understand phrase "with hull shape determined by a target parameter".

doc/reference_processing.xml
```diff
@@ -2403,6 +2403,58 @@ FROM (SELECT  'POLYGON((0 0, 8 8, 0 16, 0 0))'::geometry geom) As foo;
...
+      </refnamediv>
+
+      <refsynopsisdiv>
+        <funcsynopsis>
+         <funcprototype>
+            <funcdef>geometry <function>ST_SimplifyPolygonal</function></funcdef>
+            <paramdef><type>geometry </type> <parameter>param_geom</parameter></paramdef>
+            <paramdef><type>float </type> <parameter>vertex_fraction</parameter></paramdef>
+            <paramdef choice="opt"><type>integer </type> <parameter>is_outer = 1</parameter></paramdef>
```
Komzpa, line 2419: sounds like bool really?

doc/reference_processing.xml
```diff
@@ -2403,6 +2403,58 @@ FROM (SELECT  'POLYGON((0 0, 8 8, 0 16, 0 0))'::geometry geom) As foo;
...
+        <refname>ST_SimplifyPolygonal</refname>
+        <refpurpose>Computes a boundary-respecting hull of a polygonal geometry, with hull shape determined by a target parameter.</refpurpose>
+      </refnamediv>
+
+      <refsynopsisdiv>
+        <funcsynopsis>
+         <funcprototype>
+            <funcdef>geometry <function>ST_SimplifyPolygonal</function></funcdef>
+            <paramdef><type>geometry </type> <parameter>param_geom</parameter></paramdef>
+            <paramdef><type>float </type> <parameter>vertex_fraction</parameter></paramdef>
```
Komzpa, line 2418: is that a useful call signature? I can imagine having vertex_count if I want to try this instead of ST_Subdivide in some place. I can see going down to some threshold in either area or vertex count, but not really a case for "let's remove 10% points from each geometry, and yes just one point from this 10-point one".

liblwgeom/lwgeom_geos.c
```diff
@@ -2118,4 +2118,31 @@ lwgeom_concavehull(const LWGEOM* geom, double ratio, uint32_t allow_holes)
...
+	LWGEOM* result;
+	int32_t srid = RESULT_SRID(geom);
+	uint8_t is3d = FLAGS_GET_Z(geom->flags);
+	GEOSGeometry *g1, *g3;
+
+	initGEOS(lwnotice, lwgeom_geos_error);
+
+	if (!(g1 = LWGEOM2GEOS(geom, AUTOFIX))) GEOS_FAIL();
+
+	g3 = GEOSPolygonHullSimplify(g1, is_outer, vertex_fraction);
```
Komzpa, line 2134: I think word "hull" is more important here than "polygon".

What happens if you pass lines or multipoints? Can you chain this a-la ST_RemoveRepeatedPoints / ST_Simplify to only affect polygons on input?

Komzpa (commented):
I believe that the interesting part of this algorithm is not that it is "polygonal simplify" (that title should go to SimplifyVW really?) but that it produces a result that covers the original geometry. 

I've needed that in the past and have used ST_Simplify(ST_Buffer(geom, eps), eps/2) to get similar effect in poor man's tools. May be worth mentioning in docs.

Can it be treated as a variant of SimplifyVW that is only area-extending? like SimplifyVW(covering=true).

This can be useful for building collection overviews. Does this work on rings or on triangulations? What will happen to poly with two rings nearby - will they stick together?

A pass-through for non-polygon inputs will make this interesting for brainless application before visualization, like simplification modes in QGIS postgis driver.

doc/reference_processing.xml
pramsey: I think the question of "how much" is hard to nail down, different people will have different takes. "Somewhat less detailed" for some and "no more than this many vertices" for others.

doc/reference_processing.xml
Komzpa: what is "middle polygon hull"? is it SimplifyVW really?

liblwgeom/lwgeom_geos.c
pramsey: Right now we pass stuff down to GEOS which eventually complains about non-polygonal input. Could execute a passthrough at the PostGIS level or at at the GEOS level (probably the PostGIS level).

doc/reference_processing.xml
Komzpa: In Postgis we're normally working on sets of map objects, unlike GEOS that works on one. A thing that feels natural to me to expect from "simplification" is that things that are simple already aren't touched, and I believe this assumption is not followed with this signature. 

Do I get it right that the actual controllable range is between ST_NPoints(geom) and ST_NPoints(ST_ConvexHull(geom)) for outer hull and ST_NPoints(geom) and 3 for inner hull?

doc/reference_processing.xml
dr-jts: Close, but it will preserve polygonal topology, but SimplifyVW does not do right now.  I guess we could expose it under `SimplifyVW` for the polygonal case instead.

dr-jts:
Options for naming this:
* `ST_SimplifyPolygonal`- leaves the door open for adding a 3rd variant which isn't a hull, just simplfied
* `ST_SimplifyPolygonHull` - promotes the `Hull` concept.  A non-hull simplify algorithm (TBD) would probably have to go under `ST_SimplifyVW` - which isn't terrible
* other?

Thoughts?  @Komzpa @pramsey @robe2

Komzpa:
Thinking a bit more this can be treated not as a simplification method but something else.

The outer hull thingy is really in ST_ConcaveHull already (in the other PR) if I understand correctly, with a bit different signature but well known already. If they're not the same this development needs more documentation :) Then the line-point part can be handled by other algorithms and behavior would be consistent.

The inner hull thingy was not obtainable before and is new. Looking around it may be called Inscribed Polygon (per https://en.wikipedia.org/wiki/Inscribed_figure) and that feels like it makes more sense with working only on areas: you can't inscribe a polygon into point or line.

What do you think?

dr-jts:
The underlying algorithm also supports a parameter specifying the maximum change in area, as a ratio (`AreaDeltaRatio`).  This  could be useful when using this function to simplify as a way of speeding up intersection tests. E.g. by specifying an Area Delta Ratio of 0.1 (10%), the result is guaranteed to not exceed (reduce) the original area by more than 10%.

dr-jts:
> The outer hull thingy is really in ST_ConcaveHull already (in the other PR) if I understand correctly, with a bit different signature but well known already. If they're not the same this development needs more documentation :) Then the line-point part can be handled by other algorithms and behavior would be consistent.

Not exactly.  The Concave Hull for Polygonal geoms doesn't produce as good a result on single polygons. And for MultiPolygons it is quite different - it joins them together, where as this function keeps them separate.  Also, the parameter is different (length VS vertex count or area).

In fact, that's why we changed this recently to emphasize the "simplify" concept - it seemed more similar to simplification than to Concave Hull.

This was original just called PolygonHull.  Maybe we should just use that name in PostGIS? (`ST_PolygonHull`).  

> The inner hull thingy was not obtainable before and is new. Looking around it may be called Inscribed Polygon (per https://en.wikipedia.org/wiki/Inscribed_figure) and that feels like it makes more sense with working only on areas: you can't inscribe a polygon into point or line.

Inscribed Polygon is a nice concept.  I guess the opposite would be Exscribed Polygon?  But in fact both are produced by the same algorithm, and have the same parameters.  So I think it's better/simpler to keep them under the same function.

robe2:
> Options for naming this:
> 
>     * `ST_SimplifyPolygonal`- leaves the door open for adding a 3rd variant which isn't a hull, just simplified
> 
>     * `ST_SimplifyPolygonHull` - promotes the `Hull` aspect.  A non-hull simplify algorithm (TBD) would have to go under `ST_SimplifyVW` - which is fine.
> 
>     * `ST_PolygonHull` - even more focus on hull aspect.  Can mention simplification in doc, with links.
> 
>     * other?
> 
> 
> Thoughts? @Komzpa @pramsey @robe2

ST_SimplifyPolygonHull

As I understand it, unlike ST_ConcaveHull if you have a multipolygon you would still have two polygons, so it works at the polygon level.  But in the end what you get back is simplified.  

You can have an inner hull and an outer hull with this right and the outer hull is guaranteed to cover the original geometry?

dr-jts:
> ST_SimplifyPolygonHull

> As I understand it, unlike ST_ConcaveHull if you have a multipolygon you would still have two polygons, so it works at the polygon level. But in the end what you get back is simplified.

Correct.

> You can have an inner hull and an outer hull with this right and the outer hull is guaranteed to cover the original geometry?

Yes, that's right.  The outer hull covers the original geometry, and the inner hull is within the original geometry.

strk:
If I understood what the function would do I'd call it `ST_SimplifiedPolygonHull(polygon geometry, tolerance float8)`, to be extended with different interpretations of the tolerance and algorithms to use ? 

Would a call to the function with tolerance 0 always produce a copy of the input ?

dr-jts:
> If I understood what the function would do I'd call it `ST_SimplifiedPolygonHull(polygon geometry, tolerance float8)`, to be extended with different interpretations of the tolerance and algorithms to use ?

Sounds good, thanks.

> Would a call to the function with tolerance 0 always produce a copy of the input ?

Actually tolerance (vertex fraction) = 1  always produces a copy of the input.  
Vertex Fraction = 0 produces the convex hull for an outer hull, and a triangle for the inner hull.

PR #690, opened 2022-06-07
Expose GEOS polygonal concave hull in ST_ConcaveHull

pramsey:
References https://trac.osgeo.org/postgis/ticket/5168


Komzpa:
Why is this not inside GEOS C API instead? I believe other consumers may need that, if that's unconditional.

pramsey:
GEOS C API exposes the "concave full of points" (`GEOSConcaveHull()`) separately from the "concave hull of a polygonal area" (`GEOSConcaveHullOfPolygons()`). In PostGIS we wrap them both under `ST_ConcaveHull` and switch between them based on the input. If someone wants to get a "pointwise" hull of a polygon, they can just wrap it in `ST_Points` and feed it to `ST_ConcaveHull`. This also apes the pre-existing `ST_ConcaveHull` implementation in plpgsql, which had different behaviours for different coord dimensioned inputs. 

The principle of "least surprise" seems to dictate that for PostGIS we wrap the functionality and auto-switch, but for GEOS we make the API explicit and let devs choose.

PR #691, opened 2022-06-14
ST_TriangulatePolygon for contrained polygon triangulation

pramsey:
GEOS 3.11 only, "constrained polygon triangulation" outputs a set of triangles that exactly covers the input polygon.


Komzpa:
This is already quite literally present as ST_Tesselate: https://postgis.net/docs/ST_Tesselate.html

Once we git rid of postgis.backend GUC switch, does it make sense to have this duplicated again?

pramsey:
Can take over the ST_Tesslate() function from CGAL if that's OK with you. Otherwise it needs *some* other name?

Komzpa:
I'd say if it brings a profit (faster, stronger, robuster) then do it for the handled cases (GEOS is 2D only, right? not sure it won't break on vertical sides of cube). This will also save a breakage on non-3.11 geos as it will just use the old SFCGAL impl. I think we did similar thing to something like ST_3DIntersects or something like that?

If it does not bring user a profit against ST_Tesselate it does not need to be.

pramsey:
Most obvious profit is availability. Lots of packagers omit CGAL.

pramsey:
Is CGAL ST_Tesselate going to handle a cube? Can't really take over that signature then. A run-time switch implies both GEOS and CGAL installed at the same time, which they won't be.

pramsey:
Yes, it does handle POLYHEDRALSURFACE, which says to me we cannot just hijack that signature and we need another signature for the GEOS-enabled function.

robe2:
@pramsey  I thought we had decided on a different name for this, something like ST_Triangulate?

But as you noted it's not a replacement for ST_Tesselate since that can handle 3D Surfaces and the geos one can not.  We had discussed moving forward, prefixing the name of SFCGAL functions since we can no longer rely on backend switching and that was confusing anyway.

I was thinking more like CG_Tesselate if we go down that path.  So renaming ST_Tesselate to CG_Tesselate in SFCGAL would be a breaking change in 3.3, and then ST_Tesselate could be used for the GEOS one.

pramsey:
How about CGAL gets ST_Tesselate3D and ST_Tesselate moves to core / GEOS.

robe2:
Just throwing it out there as another option.  Tesselate has two spellings.

ST_Tesselate -- currently used (is US spelling)

and 

ST_Tessellate -- which is the more common spelling.

We had had this argument when ST_Tesselate came into being that it should have been ST_Tessellate.

You can call it ST_Tessellate.

No conflict in name, though it may exasperate people who have used the other.

pramsey:
@robe2 says
>  I think ST_TriangulatePolygon is the clearest name for what it does, plus the side benefit of it can be spelled only one way

PR #692, opened 2022-06-15
Fix unused variable

lbartoletti:
Some variables are (or considered as) unused, and clang `Werror` doesn't like that.


pramsey:
Committed, thank you!

PR #693, opened 2022-06-15
[CI][Cirrus] Fix and update FreeBSD CI

lbartoletti:
There was two problems:

- `strerror_l@FBSD_1.6` it was an ABI problem, the package are for 12.3 and the CI was still at 12.2
- the build didn't stop: it's a [cirrus problem](https://github.com/cirruslabs/cirrus-ci-docs/issues/910). I have to stop postgresql's service to release the CI – but have to use only one script parts to configure and build.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -19,11 +19,11 @@ task:
     - service postgresql oneinitdb
     - service postgresql onestart
     - su postgres -c "createuser -s `whoami`"
-    - gmake -j8
-    - gmake -j8 check
-    - gmake -j8 install
-    - gmake -j8 check RUNTESTFLAGS="-v --extension"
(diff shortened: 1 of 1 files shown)


lbartoletti:
13.1-RELEASE fail because needs https://github.com/postgis/postgis/pull/692

robe2:
This looks to be failing because of the -Werror,-Wunused-but-set-variable]

I'm wondering why 12.3 doesn't have the same issue.  Are you setting the switches differently or is that the default for 13.1?  We should probably just fix the issue in PostGIS.

robe2:
@lbartoletti  Sorry didn't see your note earlier okay.

lbartoletti:
> I'm wondering why 12.3 doesn't have the same issue.

It's not the same version of clang.

pramsey:
Merged, thanks!

PR #697, opened 2022-06-16
gserialized.c: Fix an always true condition

lbartoletti:
At line 366 we always check that bsz1 is different than bsz2. So, if
bsz1 is not inferior to bsz2, it always true than bsz1 is superior to
bsz2.
We could remove the else if part and return 1.

### DIFF
--- a/liblwgeom/gserialized.c
+++ b/liblwgeom/gserialized.c
@@ -367,7 +367,7 @@ int gserialized_cmp(const GSERIALIZED *g1, const GSERIALIZED *g2)
 		{
 			if (bsz1 < bsz2)
 				return -1;
-                        return 1;
+			return 1;
 		}
 
 		/* If SRID is not equal, sort on it */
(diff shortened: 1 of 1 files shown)


pramsey:
Sure seems fine. Since it's an un-bracketed statement, maybe keep the unconditional else
```
if (bsz1 < bsz2)
  return -1;
else 
  return 1;
```

lbartoletti:
> Since it's an un-bracketed statement

I don't like un-bracketed statement :) But, it seems the postgis style here.

> , maybe keep the unconditional else
> 
> ```
> if (bsz1 < bsz2)
>   return -1;
> else 
>   return 1;
> ```

I don't like to use a useless `else`, if possible I'll write:

```
if (bsz1 < bsz2)
{
  return -1;
}

return 1;
```

But, I follow the postgis rule.

pramsey:
Thanks!  223a9904eeb0f82a88c9d039c3bdabf0824e1034

PR #699, opened 2022-06-16
gserialized1.c: fix a Null pointer dereference

lbartoletti:
If lwgeom is NULL we write an error log thanks to lwerror.

However, lwgeom is used potentially NULL lines after.

This commit fix a potential crash returning NULL if lwgeom is NULL.


lbartoletti:
Indeed, I read too fast, `lwerror` calls `default_errorreporter` which `exit(1)`https://github.com/postgis/postgis/blob/master/liblwgeom/lwutil.c#L142-L148

PR #701, opened 2022-07-26
Use 2D index operators for 2D functions (supportfn)

pramsey:
Ensure that additional operators are not appended when the function and opfamily disagree about dimensionality. References [TRAC-5025](https://trac.osgeo.org/postgis/ticket/5025#comment:8)


robe2:
@pramsey  any chance you can rebase this.  I think the failure might be an issue of upgrade that we've since fixed since this pull request was done.

robe2:
This was already committed to PostGIS 3.2.3 and 3.3.0.

PR #705, opened 2022-08-29
Add ST_TransformPipeline

rcoup:
As discussed in [Trac #5006](https://trac.osgeo.org/postgis/ticket/5006), add the ability to use a specific PROJ conversion pipeline instead of the automatically selected one via `ST_Transform()` (`proj_crs_to_crs()`).

Thanks to @rouault for all his help at the FOSS4G 2022 code sprint getting it working 😃  Commit messages should be relatively self-explanatory.

---

```SQL
geometry ST_TransformPipeline(geom geometry, pipeline text[, to_srid integer])
geometry ST_InverseTransformPipeline(geom geometry, pipeline text[, to_srid integer])
```

Transformation pipelines are defined using any of the following string formats:

  - `urn:ogc:def:coordinateOperation:AUTHORITY::CODE`. Note that a simple `AUTHORITY:CODE` string does not uniquely identify a coordinate operation: the same code can be used for a CRS definition.
  - a PROJ pipeline string of the form: `+proj=pipeline ...`. Automatic axis normalisation will not be applied, and if necessary the caller will need to add an additional pipeline step.
  - concatenated operations of the form: `urn:ogc:def:coordinateOperation,coordinateOperation:EPSG::3895,coordinateOperation:EPSG::1618`

The SRID of the input geometry is ignored, and the SRID of the output geometry will be set to zero unless a value is provided via the optional `to_srid` parameter.

When using `ST_TransformPipeline()` the pipeline is executed in a forward direction. Using `ST_InverseTransformPipeline()` the pipeline is executed in the inverse direction.

---

GDA2020 example from the docs:

```sql
-- using ST_Transform with automatic selection of a conversion pipeline.
SELECT ST_AsText(ST_Transform('SRID=4939;POINT(143.0 -37.0)'::geometry, 7844)) AS gda2020_auto;

                 gda2020_auto
-----------------------------------------------
 POINT(143.00000635638918 -36.999986706128176)
(1 row)

-- using a defined conversion (EPSG:8447)
SELECT ST_AsText(ST_TransformPipeline('SRID=4939;POINT(143.0 -37.0)'::geometry,
  'urn:ogc:def:coordinateOperation:EPSG::8447')) AS gda2020_code;

                   gda2020_code
----------------------------------------------
 POINT(143.0000063280214 -36.999986718287545)
(1 row)

-- using a PROJ pipeline definition matching EPSG:8447, as returned from
-- 'projinfo -s EPSG:4939 -t EPSG:7844'.
-- NOTE: any 'axisswap' steps must be removed.
SELECT ST_AsText(ST_TransformPipeline('SRID=4939;POINT(143.0 -37.0)'::geometry,
  '+proj=pipeline
   +step +proj=unitconvert +xy_in=deg +xy_out=rad
   +step +proj=hgridshift +grids=au_icsm_GDA94_GDA2020_conformal_and_distortion.tif
   +step +proj=unitconvert +xy_in=rad +xy_out=deg')) AS gda2020_pipeline;

                   gda2020_pipeline
----------------------------------------------
 POINT(143.0000063280214 -36.999986718287545)
(1 row)
```

### DIFF
--- a/doc/reference_srs.xml
+++ b/doc/reference_srs.xml
@@ -2,130 +2,130 @@
   <sect1 id="SRS_Functions">
     <sect1info>
     <abstract>
-	<para>These functions work with the Spatial Reference System of geometries
+  <para>These functions work with the Spatial Reference System of geometries
     as defined in the <varname>spatial_ref_sys</varname> table.</para>
    </abstract>
     </sect1info>
 
-	  <title>Spatial Reference System Functions</title>
+    <title>Spatial Reference System Functions</title>
 
-	<refentry id="ST_SetSRID">
-	  <refnamediv>
-		<refname>ST_SetSRID</refname>
+  <refentry id="ST_SetSRID">
+    <refnamediv>
+    <refname>ST_SetSRID</refname>
 
-		<refpurpose>Set the SRID on a geometry.</refpurpose>
-	  </refnamediv>
+    <refpurpose>Set the SRID on a geometry.</refpurpose>
+    </refnamediv>
 
-	  <refsynopsisdiv>
-		<funcsynopsis>
-		  <funcprototype>
-			<funcdef>geometry <function>ST_SetSRID</function></funcdef>
+    <refsynopsisdiv>
+    <funcsynopsis>
+      <funcprototype>
+      <funcdef>geometry <function>ST_SetSRID</function></funcdef>
 
-			<paramdef><type>geometry </type>
-			<parameter>geom</parameter></paramdef>
+      <paramdef><type>geometry </type>
+      <parameter>geom</parameter></paramdef>
 
-			<paramdef><type>integer </type>
-			<parameter>srid</parameter></paramdef>
-		  </funcprototype>
-		</funcsynopsis>
-	  </refsynopsisdiv>
+      <paramdef><type>integer </type>
+      <parameter>srid</parameter></paramdef>
+      </funcprototype>
+    </funcsynopsis>
+    </refsynopsisdiv>
 
-	  <refsection>
-		<title>Description</title>
+    <refsection>
+    <title>Description</title>
 
-		<para>Sets the SRID on a geometry to a particular integer value.
-		Useful in constructing bounding boxes for queries.</para>
+    <para>Sets the SRID on a geometry to a particular integer value.
+    Useful in constructing bounding boxes for queries.</para>
 
-		<note>
-		  <para>This function does not transform the geometry coordinates in any way -
-		  it simply sets the meta data defining the spatial reference system the geometry is assumed to be in.
-		  Use <xref linkend="ST_Transform"/> if you want to transform the
-		  geometry into a new projection.</para>
-		</note>
-		<para>&sfs_compliant;</para>
-		<para>&curve_support;</para>
-	  </refsection>
+    <note>
+      <para>This function does not transform the geometry coordinates in any way -
+      it simply sets the meta data defining the spatial reference system the geometry is assumed to be in.
+      Use <xref linkend="ST_Transform"/> if you want to transform the
+      geometry into a new projection.</para>
+    </note>
+    <para>&sfs_compliant;</para>
+    <para>&curve_support;</para>
+    </refsection>
 
-	  <refsection>
-			<title>Examples</title>
-			<para>-- Mark a point as WGS 84 long lat --</para>
-			<programlisting>SELECT ST_SetSRID(ST_Point(-123.365556, 48.428611),4326) As wgs84long_lat;
+    <refsection>
+      <title>Examples</title>
+      <para>-- Mark a point as WGS 84 long lat --</para>
+      <programlisting>SELECT ST_SetSRID(ST_Point(-123.365556, 48.428611),4326) As wgs84long_lat;
 -- the ewkt representation (wrap with ST_AsEWKT) -
 SRID=4326;POINT(-123.365556 48.428611)
-			</programlisting>
-			<para>-- Mark a point as WGS 84 long lat and then transform to web mercator (Spherical Mercator) --</para>
-			<programlisting>SELECT ST_Transform(ST_SetSRID(ST_Point(-123.365556, 48.428611),4326),3785) As spere_merc;
+      </programlisting>
+      <para>-- Mark a point as WGS 84 long lat and then transform to web mercator (Spherical Mercator) --</para>
+      <programlisting>SELECT ST_Transform(ST_SetSRID(ST_Point(-123.365556, 48.428611),4326),3785) As spere_merc;
 -- the ewkt representation (wrap with ST_AsEWKT) -
 SRID=3785;POINT(-13732990.8753491 6178458.96425423)
-			</programlisting>
-		</refsection>
+      </programlisting>
+    </refsection>
 
-	  <refsection>
-		<title>See Also</title>
+    <refsection>
+    <title>See Also</title>
 
-		<para><xref linkend="spatial_ref_sys" />, <xref linkend="ST_SRID"/>, <xref linkend="ST_Transform"/>, <xref linkend="UpdateGeometrySRID"/></para>
-	  </refsection>
+    <para><xref linkend="spatial_ref_sys" />, <xref linkend="ST_SRID"/>, <xref linkend="ST_Transform"/>, <xref linkend="UpdateGeometrySRID"/></para>
+    </refsection>
 
-	</refentry>
+  </refentry>
 
-	<refentry id="ST_SRID">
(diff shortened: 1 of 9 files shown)


robe2:
@rcoup I have a couple of issues with your commit:


1) You are missing an availability notice on the new function.

It should read:
```
<para>Availability: 3.4.0 - Requires PostGIS be compiled with PROJ v6.1+.  Use <xref linkend="PostGIS_Full_Version" /> to confirm you have PROJ support compiled in.</para>
```

2) Your tests also don't seem to guard against running on PROJ < 6.1, but that might be okay.
I think come 3.4.0 we might dispense with PROJ < 6.1. As I recall @pramsey  is the only one that had qualms about dropping PROJ < 6.  @pramsey  you still feel on the fence on that.

3) Your sql functions have too many prototypes and have the wrong availability, they should read

Availability: 3.4.0  

not 3.3.0

I would also reduce down to 1 SQL interface of:

```geometry ST_TransformPipeline(geom geometry, pipeline text, forward boolean DEFAULT true, to_srid integer DEFAULT 0)
```

4) Also our preferred datatype is **boolean** instead of **bool** for the sql api.  I know we still have some dangling bool's in there. so sorry for that.

rcoup:
@robe2 thanks! Can easily make those changes.

> Your tests also don't seem to guard against running on PROJ < 6.1, but that might be okay.

Is this the [best example of such a guard](https://github.com/postgis/postgis/blob/a17e6122a6b54f19e25a62a2fde66f73b843af68/regress/core/tickets.sql#L589-L610), or is there an easier mechanism?

rcoup:
> Is the "forward" parameter really needed ?
> I mean: can a pipeline not always be expressed differently,
> to be backward ? I'm thinking of a function that, taken a pipeline,
> returns it's "reverse", so for non-forward you'd do:

@rouault can explain better than me, but my understanding is:

* the EPSG DB defines only forward conversions (eg: EPSG:8447)
* PROJ can perform a transform in either forward or reverse
* while if you take the proj pipeline text (`+proj=pipeline`) and reverse the steps, you'll _generally_ get the same answers, the PROJ pipeline definition can't express everything that the ISO spec / defined CRS do
* so having "execute in reverse" is a better approach

rouault:
> * so having "execute in reverse" is a better approach

what is better or not is a master of taste, but it is indeed more straightforward as implementation. That said, ``PJ* proj_coordoperation_create_inverse(PJ_CONTEXT *ctx, const PJ *obj) exists`` and could in theory be used to implement a ST_ReverseTransformPipeline().
To be noted that WKT obtained from such an inversed transformation object is generally not interoperable with other implementations, due to falling back to transformation method names like "Inverse of {forward_method_name}" to indicate that the method should be taken in the reverse path (not sure if there are other implementations than PROJ that accept WKT for transformations anyway...), when the transformation cannot be reversed with an existing foward method. This remarks only matter if PostGIS would have a WKT output of transformation pipelines.

robe2:
> @robe2 thanks! Can easily make those changes.
> 
> > Your tests also don't seem to guard against running on PROJ < 6.1, but that might be okay.
> 
> Is this the [best example of such a guard](https://github.com/postgis/postgis/blob/a17e6122a6b54f19e25a62a2fde66f73b843af68/regress/core/tickets.sql#L589-L610), or is there an easier mechanism?

Our general mechanism is to exclude the test if version is not a specific version.
We have those for GEOS / SFCGAL - such as:

https://github.com/postgis/postgis/blob/a17e6122a6b54f19e25a62a2fde66f73b843af68/regress/core/tests.mk.in#L151

But you'd need to add at the top:

```
POSTGIS_PROJ_VERSION=@POSTGIS_PROJ_VERSION@

```

I think this variable is picked up at configure time -- here https://github.com/postgis/postgis/blob/a17e6122a6b54f19e25a62a2fde66f73b843af68/configure.ac#L898

So should just work if you add to the top. @strk  am I missing anything here?

strk:
I didn't look at the link (sorry) but just make sure you do get a
specific user error if trying to use the function, something like
"PostGIS was built with proj < $version so this functionlity is not
available, please rebuild" and not something like "function not
found". It's important that the SQL and C functions do exist, for
easier upgrades later (or this is our policy at the moment I think).

robe2:
> I didn't look at the link (sorry) but just make sure you do get a specific user error if trying to use the function, something like "PostGIS was built with proj < $version so this functionlity is not available, please rebuild" and not something like "function not found". It's important that the SQL and C functions do exist, for easier upgrades later (or this is our policy at the moment I think).

@rcoup  Regarding what @strk  said above.  We always define the SQL function even when compiling with a lower version, but the safeguard is put in place and stubbed with an error if PostGIS is compiled with a lower version.

Such as this -- https://github.com/postgis/postgis/blob/master/postgis/lwgeom_transform.c#L155
and then throw an informative error such as this:

https://github.com/postgis/postgis/blob/master/postgis/lwgeom_geos.c#L347

But if I get my wish -- all of this will be unnecessary, cause we'll drop support for Proj < 6.1 so no ifdef will be needed -  as noted here:

https://lists.osgeo.org/pipermail/postgis-devel/2022-August/029786.html

BTW opinions on the vote are welcome from anybody and will sway the PSC decision.

rcoup:
> > But if I get my wish -- all of this will be unnecessary, cause we'll drop support for Proj < 6.1 so no ifdef will be needed - as noted here:
>
> https://lists.osgeo.org/pipermail/postgis-devel/2022-August/029786.html
>
> BTW opinions on the vote are welcome from anybody and will sway the PSC decision.

Ah, I'm only subscribed to postgis-users, so I didn't see this. Certainly makes sense to me: seems unlikely people will want latest-greatest PostGIS from 2022 but aren't prepared to bring along a vaguely modern PROJ (6.1 was released May 2019).

Will wait to update this PR until that decision is settled, but if it passes it's probably worth rebasing this on a PROJ-removal PR. I'm happy to do that if I can find some time.

rcoup:
> could in theory be used to implement a ST_ReverseTransformPipeline()

Feel to me like `ST_TransformPipeline()` is already extra-for-experts territory: PostGIS has got a _long_ way with `ST_Transform()`, and the people dealing with more complicated transformations probably know a bit more about what they're trying to do.

robe2:
> Ah, I'm only subscribed to postgis-users, so I didn't see this. Certainly makes sense to me: seems unlikely people will want latest-greatest PostGIS from 2022 but aren't prepared to bring along a vaguely modern PROJ (6.1 was released May 2019).
> 
> Will wait to update this PR until that decision is settled, but if it passes it's probably worth rebasing this on a PROJ-removal PR. I'm happy to do that if I can find some time.

Motion passed.  
https://lists.osgeo.org/pipermail/postgis-devel/2022-September/029793.html

But let me first start removing bots testing with < 6.1 and then we can rerun.
We also voted to remove PG 11 too.

robe2:
@rcoup Okay done ripping out support of PROJ < 6.1 and PostgreSQL < 12.  So now you can rebase.

https://trac.osgeo.org/postgis/ticket/5229

robe2:
@rcoup Just checking if you are ready to rebase and reapply.  As mentioned no need for the conditional proj since we no longer support PROJ < 6.1 for 3.4

strk:
Thanks for the explanation. Would encoding inverse in the function
name be more elegant then ?

  ST_ReverseTransformPipeline() ?

rcoup:
@robe2 @strk I've made those changes now.

I went with adding `ST_InverseTransformPipeline()` instead of the `forward` parameter.

Feedback welcome 😃

rcoup:
ok, so `make check` seems to work locally (and exits with 0). But looking closer:
```
Loading PostGIS into 'postgis_reg-3.4'
 failed (Error encountered loading /Users/rcoup/code/postgis/regress/00-regress-install/share/contrib/postgis/postgis_comments.sql: /tmp/pgis_reg/regress_log)
```

```
$ cat /tmp/pgis_reg/regress_log
NOTICE:  schema "public" already exists, skipping
NOTICE:  schema "public" already exists, skipping
psql:/Users/rcoup/code/postgis/regress/00-regress-install/share/contrib/postgis/postgis_comments.sql:282: ERROR:  function st_transformpipeline(geometry, text) does not exist
NOTICE:  schema "public" already exists, skipping
...
```

Which leads to a syntax issue in the docs. Easy fix 👍 

I can't spot the area in the twisty maze of Makefiles where it's actually ignoring the error.

strk:
The false success should be fixed with 009004868bda8b495b4c8340239abe26b828250d

rcoup:
@strk thanks. I re-added a similar error locally & `make check` fails (returns 2) now. 👍

PR #706, opened 2022-09-15
Remove extraneous #if in postgis.sql.in

rjuju:
Oversight in commit 3e38e0b83b98925122bb22dff958a7a603ee4a42.

For the record the CI does notice this problem but keeps going.  For instance https://github.com/postgis/postgis/actions/runs/3057481917/jobs/4932676006 in the "Build & Test" item line 646:

> postgis.sql.in:4436: error: unterminated #if
>  4436 | #if POSTGIS_PGSQL_VERSION >= 120
>       |


strk:
On Wed, Sep 14, 2022 at 09:38:47PM -0700, Julien Rouhaud wrote:

> For the record the CI does notice this problem but keeps going.
> For instance https://github.com/postgis/postgis/actions/runs/3057481917/jobs/4932676006 in the &quot;Build &amp; Test&quot; item line 646:

This is a bug on itself, could you please file it ? Or fix it :)
Is it only in CI or in any build process ? (bug in Makefile?)

--strk;

rjuju:
This happens in any build process. I just happened to saw it after a "make -s" while pretty much everything else was already compiled so it was kind of obvious.

Is comes from https://github.com/postgis/postgis/blob/master/postgis/Makefile.in#L239-L241:

```
%.sql: %.sql.in
	$(SQLPP) -I@top_srcdir@/libpgcommon -I@builddir@ $< | grep -v '^#' | \
	$(PERL) -lpe "s'MODULE_PATHNAME'\$(MODULEPATH)'g;s'@extschema@\.''g" > $@
```

the SQLPP command itselfs errors out, but since the output is piped make sees the result of the last command, which doesn't fail.

I'm not really sure how to fix it.  I could add this after:

```
test ${PIPESTATUS[0]} -eq 0
```

but I don't know how portable it's (especially on windows).

Note that the same pattern is used in many other makefiles.

strk:
> I'm not really sure how to fix it.  I could add this after:

How about 97903f89960c8079e8edf9c9b9db74dd66c1e194 ?

rjuju:
It works too :)

Isn't it missing the `rm -f $@.tmp` in the other makefiles?

Note that for postgis.sql it leaks the .tmp file in case of error, but it will eventually be cleaned up once the source file is fixed so it doesn't seem like a problem.

strk:
> Isn't it missing the `rm -f ***@***.***` in the other makefiles?

Yeah, I was wondering if this would have been best done
by "make clean" (and I think it would, yes). But I was too lazy
to fix that. Unless there's a "distcheck" rule that spots this
problem I'll leave this as an excercise for the readers :)

strk:
This was merged as 229eeb7847110007ce672ed148e9f3b194317452

strk:
I love our testsuite. THere IS actually a bot running `distcheck`:
https://gitlab.com/postgis/postgis/-/jobs/3042375176#L5376

My .tmp files didn't go unnoticed there!

PR #709, opened 2022-09-30
WIP: fix cirrus-CI build

strk:
CirrusCI is failing not finding the file.
Chances are multi-process make is building the same file multiple times and thus not finding it upon removing it

A better fix would be useful (serializing those commands?)

Should fix #5250

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -19,11 +19,11 @@ task:
     - service postgresql oneinitdb
     - service postgresql onestart
     - su postgres -c "createuser -s `whoami`"
-    - gmake -j8 || { service postgresql onestop; exit 1;}
-    - gmake -j8 check || { service postgresql onestop; exit 1;}
-    - gmake -j8 install || { service postgresql onestop; exit 1;}
-    - gmake -j8 check RUNTESTFLAGS="-v --extension" || { service postgresql onestop; exit 1;}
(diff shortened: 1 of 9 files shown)


strk:
\cc @lbartoletti

strk:
@lbartoletti I'm confused by the cirrus-CI jobs not appearing in the Action tab of github:
https://github.com/postgis/postgis/actions/runs/3159812099

Is Cirrus-CI using the old API ? Any idea how to verify it worked with my `rm -f` change ? Because I don't see it failing when using the debug-run in cirrus-ci but I think it was marked as failing here in the PR check.
Are we marking Cirrus-CI as non-mandatory to pass here ?

strk:
Sell I've merged 03cc059dc01b0fa39564d97295748e6f12bfebbc

lbartoletti:
Sorry for the delay...

> @lbartoletti I'm confused by the cirrus-CI jobs not appearing in the Action tab of github: https://github.com/postgis/postgis/actions/runs/3159812099

It does not appear in the Action tab. You can see other CI, only by clicking on the red cross or green mark, on a PR or the history.

> Is Cirrus-CI using the old API ? 

I don't know about the API

> Any idea how to verify it worked with my `rm -f` change ? Because I don't see it failing when using the debug-run in cirrus-ci but I think it was marked as failing here in the PR check. 

You should see jobs on http://cirrus-ci.com/github/postgis/postgis

> Are we marking Cirrus-CI as non-mandatory to pass here ?

You can, since we have already bessie and bessie32.

PR #711, opened 2022-10-20
Convert GeoJSON output to use stringbuffer

pramsey:
Instead of count-n-fill procedure. Much less code, have not measured performance.


robe2:
what's left to do here?  Just performance tests?

pramsey:
Yeah, I fear it'll end up slower. Haven't tested yet.

pramsey:
Mixed performance results. New code seems about 10% faster on a "small object" case, converting a large number of points, or a large number of two-point lines. New code seems about 5% slower on a "large object" case, converting the naturalearth admin layer. Not sure where that leaves us in terms of whether we want the change or not.

pramsey:
Even after change, the timings above still hold.

PR #712, opened 2022-10-31
Mention "NOT ST_Intersects" as index-using alternative to ST_Disjoint

kannes:
I had no idea and I am sure others might benefit from this as well. Not sure if the wording is perfect but I tried to adjust to other sentences from the docs. 
I have to admit that I did not try to build the docs.

(@pramsey said it sounds ok to add this (ref https://twitter.com/cartocalypse/status/1585223073418350593))

### DIFF
--- a/doc/reference_relationship.xml
+++ b/doc/reference_relationship.xml
@@ -794,7 +794,8 @@ WHERE ST_Crosses(roads.geom, highways.geom);</programlisting>
       <para>Performed by the GEOS module</para>
       <note>
         <para>This function call does not use indexes. A negated <xref linkend="ST_Intersects" /> predicate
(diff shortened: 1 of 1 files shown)


dr-jts:
MIght want to include the `NOT ST_Intersects` syntax explicitly in the note, so that it is 100% clear what "negated" means, and to (slightly) lower the cognitive effort required to use this pattern.

kannes:
With pleasure! I like docs that are explicit and detailed. I added a small code-ish example, inspired by https://postgis.net/docs/ST_Within.html
Hopefully it will render correctly.

robe2:
Committed at https://github.com/postgis/postgis/commit/c7c9e22255696e1461aeeb96b424ff2e22b45468

PR #713, opened 2022-12-08
More resolution on statistics for small tables

pramsey:
Handle smaller tables with more aplomb. Potential remaining issues are around the use of the range_quintile() calculation to determine the extrema of the distribution. It might be smarter to go with the full series range intead of the 1st to 4th quintile range, since, for example, a table with a pile of points in exactly one spot may end up with those points not counting against the "distribution range" of the dimension, and thus washed out into fewer cells than it should reasonably get. I'm not sure the interquintile distance offers the protection against "outliers" that I thought it did when using it, a bit too much belt and suspenders.


Komzpa (approved):
been there, useful change.

robe2 (approved):

PR #714, opened 2022-12-14
Selectivity estimates on @/~ operators

pramsey:
Historically, we have not had selectivity estimates added to the contains/within (@/~) operators. This has resulted in bad plans, probably forever, in more complex queries that include a contains/within spatial test. (Actually historically, we didn't use the ~/@ operators in our ST_Contains/ST_Within wrappers until version 2.2, so that's how long we have been broken.)

This PR adds in the selectivity estimates, without making any attempt to rationalize for upgrades. Just thinking about that makes me shiver.

### DIFF
--- a/postgis/gserialized_estimate.c
+++ b/postgis/gserialized_estimate.c
@@ -523,28 +523,28 @@ nd_stats_to_json(const ND_STATS *nd_stats)
 * Caller is responsible for freeing.
 * Currently only prints first two dimensions.
 */
-// static char*
-// nd_stats_to_grid(const ND_STATS *stats)
-// {
-//  char *rv;
-//  int j, k;
-//  int sizex = (int)roundf(stats->size[0]);
-//  int sizey = (int)roundf(stats->size[1]);
-//  stringbuffer_t *sb = stringbuffer_create();
-//
-//  for ( k = 0; k < sizey; k++ )
-//  {
-//      for ( j = 0; j < sizex; j++ )
-//      {
-//          stringbuffer_aprintf(sb, "%3d ", (int)roundf(stats->value[j + k*sizex]));
-//      }
-//      stringbuffer_append(sb,  "\n");
-//  }
-//
-//  rv = stringbuffer_getstringcopy(sb);
-//  stringbuffer_destroy(sb);
-//  return rv;
-// }
+static char*
+nd_stats_to_grid(const ND_STATS *stats)
+{
+ char *rv;
+ int j, k;
+ int sizex = (int)roundf(stats->size[0]);
+ int sizey = (int)roundf(stats->size[1]);
+ stringbuffer_t *sb = stringbuffer_create();
+
+ for ( k = 0; k < sizey; k++ )
+ {
+     for ( j = 0; j < sizex; j++ )
+     {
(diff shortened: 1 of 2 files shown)


robe2:
You are missing CHANGED comment 

something like:

`-- Changed: 3.4.0 changed to use selectivity estimates for contains and covers`

strk:
Upgrade could be done with ALTER OPERATOR, see https://www.postgresql.org/docs/9.6/sql-alteroperator.html

In order for this to be automatic I'd put the "Changed" comment right above the actual setting (RESTRICT and JOIN in this case). Doing so the upgrade script could see that specific parameter needs to be updated and use ALTER OPERATOR.

strk:
Rather, I guess were are missing TESTS for estimation using these operators. The test should use EXPLAIN and check that the estimation matches a given expectancy. There are a few test files doing that already, see regress/core/regress_index.sql for one example (it would not be bad to rationalize/reorganize them all)

strk:
I'm not sure now but I seem to remember we kept a min/max/average area in the stats too, so we could estimate how many objects may be contained in a query geometry by looking at the size too ? It's not something I'm planning to do any time soon so I've to say as long as our upgrade script is able to CHANGE that estimator we should be ok.

strk:
I'm adding a test to verify upgrades don't result in broken definitions: https://git.osgeo.org/gitea/postgis/postgis/pulls/118

Meanwhile it would be great if there could be an actual test showing the bad plans. Can you maybe provide that use case in form of a testcase ? It would be good for it to go into regress/core/regress_index.sql which has a function using EXPLAIN ANALYZE to compare estimated and obtained rows.

pramsey:
From memory. For points it's all fairly trivial, you drop the points into the grid and the estimator just counts the grid cells it hits. For polygons and lines, in addition to the "presence in a grid cell" of each object, during calculations we keep track of the average number of grid cells each object covers, so that the estimator can first count the number of presence "hits" in the grid, and then normalize it down using the cells/object average. That's as sophisticated as we get ATM.

strk:
Right, so we have that "average number of grid cells each object covers", meaning for points it'd be 1 on average, so a query box covering 2 cells would be contained in NO points, but could overlap many, depending on where it is

strk:
I've pushed a "selectivity-contains" branch to the official repository, so we can more easily work together on it

robe2:
I don't see mention of this in the NEWS file?  Where should I be looking?  Shouldn't this also be noted here https://postgis.net/docs/manual-dev/ST_Geometry_Contain.html and https://postgis.net/docs/manual-dev/ST_Geometry_Contained.html.
 I don't see @pramsey  name on anything in 3.4, I'm pretty sure he did stuff.

strk:
There was not even a trac ticket. I saw cowboy activity and started riding as well :)

pramsey:
Ride on cowboy, ride on.

strk:
I'm really leaving this to Paul as I'm still not convinced we're
providing something that is BETTER than what we had before.

For instance it strikes me that our estimator would suggest the
planner that a `the_geom ~ :FULLEXTENT:` condition will select
ALL rows while it's very likely it would select NONE ...

PR #718, opened 2022-12-27
ST_QuantizeCoordinates(): speed-up implementation

rouault:
We can completely avoid any floating-point computation by directly using the fact that the IEEE-754 representation of a floating point number contains the integral part of log2(abs(number)).
The improved trim_preserve_decimal_digits() function is ~ 7x times faster.

### DIFF
--- a/liblwgeom/lwgeom.c
+++ b/liblwgeom/lwgeom.c
@@ -2505,37 +2505,34 @@ lwgeom_is_trajectory(const LWGEOM *geom)
 	return lwline_is_trajectory((LWLINE*)geom);
 }
 
-static uint8_t
-bits_for_precision(int32_t significant_digits)
-{
-	int32_t bits_needed = ceil(significant_digits / log10(2));
-
-	if (bits_needed > 52)
-	{
-		return 52;
-	}
-	else if (bits_needed < 1)
-	{
-		return 1;
-	}
-
-	return bits_needed;
-}
+#define STATIC_ASSERT(COND,MSG) typedef char static_assertion_##MSG[(COND)?1:-1]
+STATIC_ASSERT(sizeof(double) == sizeof(uint64_t),this_should_be_true);
 
 static double trim_preserve_decimal_digits(double d, int32_t decimal_digits)
 {
-	if (d == 0)
-		return 0;
-
-	int digits_left_of_decimal = (int) (1 + log10(fabs(d)));
-	uint8_t bits_needed = bits_for_precision(decimal_digits + digits_left_of_decimal);
-	uint64_t mask = 0xffffffffffffffffULL << (52 - bits_needed);
 	uint64_t dint = 0;
-	size_t dsz = sizeof(d) < sizeof(dint) ? sizeof(d) : sizeof(dint);
+	memcpy(&dint, &d, sizeof(double));
+	/* Extract the exponent from the IEEE 754 integer representation, which */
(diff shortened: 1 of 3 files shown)


Komzpa:
clang is not happy about it:
```
Died at ./regress/run_test.pl line 796.
 ./regress/core/quantize_coordinates .. failed (psql exited with an error: /tmp/pgis_reg/test_45_out)
t1|t
psql:quantize_coordinates.sql:4: ERROR:  Must specify precision
t3|t
t4|t
t5|t
psql:quantize_coordinates.sql:14: server closed the connection unexpectedly
make: *** [regress/runtest.mk:24: check-regress] Error 2
	This probably means the server terminated abnormally
	before or while processing the request.
psql:quantize_coordinates.sql:14: error: connection to server was lost
```

liblwgeom/lwgeom.c
```diff
@@ -2505,37 +2505,34 @@ lwgeom_is_trajectory(const LWGEOM *geom)
...
-		return 0;
-
-	int digits_left_of_decimal = (int) (1 + log10(fabs(d)));
-	uint8_t bits_needed = bits_for_precision(decimal_digits + digits_left_of_decimal);
-	uint64_t mask = 0xffffffffffffffffULL << (52 - bits_needed);
 	uint64_t dint = 0;
-	size_t dsz = sizeof(d) < sizeof(dint) ? sizeof(d) : sizeof(dint);
-
-	memcpy(&dint, &d, dsz);
+	memcpy(&dint, &d, sizeof(double));
```
Komzpa, line 2514: I believe that copy-less recommended way of dealing with the type punning in C is using a two-part struct. Kinda like here: https://github.com/postgis/postgis/blob/master/postgis/gserialized_gist_2d.c#L458

rouault: memcpy() is also fine (cf https://stackoverflow.com/a/53181506). It is heavily used in GDAL and optimizers do not do memory copy, but move things between registries (at least on x86). 
I believe what you actually recommend is undefined behaviour (also most compilers do sane things in this situation): cf https://stackoverflow.com/a/69297979

rouault:
> clang is not happy about it:

ok, issue solved by adding a ``int`` cast in the expression ``(int)((dint >> 52) & 2047) - 1023;`` otherwise I suspect the sanitizer flag would complain about unsigned integer underflow when computing negative exponents.

liblwgeom/lwgeom.c
Komzpa: It's UB in C++ but not in C: https://stackoverflow.com/a/25672839. 

I checked disasm and it looks good there indeed, thanks.
image

Komzpa (approved):

liblwgeom/lwgeom.c
rouault: > It's UB in C++ but not in C

interesting... thanks

robe2:
@Komzpa @pramsey @rouault  What's your thought about backporting to 3.3 as well. Seems change is harmless enough.  I'll push to 3.4 (master) once the PG16 new GHA docker build is done.

rouault:
> Seems change is harmless enough.

Personnally I wouldn't backport this. This is really in the enhancement category, and it might change a bit results. I believe the new implementation is less pessimistic than the previous one (that is to say the number of nullified bits can be a bit higher, while meeting the expressed precision), so this will result in slightly different results.

PR #720, opened 2023-02-23
Allow singleton types to insert into multi* typmod

pramsey:
columns by promoting the input geometry.


robe2:
@pramsey  Looks good to me.

robe2 (approved):

robe2:
This was committed 1d9d3f1

PR #723, opened 2023-03-20
ST_ClusterDBSCAN memory leak

FaFre:
I believe I found a memory leak while reading some code. 
Please double check, my C is a bit rusty :)

### DIFF
--- a/postgis/lwgeom_window.c
+++ b/postgis/lwgeom_window.c
@@ -200,6 +200,7 @@ Datum ST_ClusterDBSCAN(PG_FUNCTION_ARGS)
 		}
 
 		lwfree(result_ids);
+		lwfree(is_in_cluster);
 		UF_destroy(uf);
 	}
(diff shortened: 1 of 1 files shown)


robe2:
@FaFre What made you think there is a memory leak here?

I think your change crashes the server, because at that point, is_in_cluster is already freed.

Take a look at https://github.com/postgis/postgis/blob/master/postgis/lwgeom_window.c#L182
To see what I mean.

Going to close this out. Feel free to reopen if I missed something.

FaFre:
I mean... it is only freed when there is an error (https://github.com/postgis/postgis/blob/master/postgis/lwgeom_window.c#L178). See also here: https://github.com/postgis/postgis/blob/a5c61af1a38eeb780290a3feafcb871aaa4e6197/liblwgeom/cunit/cu_geos_cluster.c#L293

PR #724, opened 2023-03-31
Add an interrupt check in dbscan

pramsey:


liblwgeom/lwgeom_geos_cluster.c
```diff
@@ -346,10 +349,16 @@ union_dbscan_minpoints_1(LWGEOM** geoms, uint32_t num_geoms, UNIONFIND* uf, doub
 
 	for (p = 0; p < num_geoms; p++)
 	{
+		int rv = LW_SUCCESS;
 		if (lwgeom_is_empty(geoms[p]))
 			continue;
 
-		dbscan_update_context(tree.tree, &cxt, geoms, p, eps);
+		rv = dbscan_update_context(tree.tree, &cxt, geoms, p, eps);
```
dbaston, line 356: Smaller than a nit, but...can't the variable be defined here in C99? (`int rv = dbascan_update_context...`)

liblwgeom/lwgeom_geos_cluster.c
```diff
@@ -421,11 +430,17 @@ union_dbscan_general(LWGEOM** geoms, uint32_t num_geoms, UNIONFIND* uf, double e
 	for (p = 0; p < num_geoms; p++)
 	{
 		uint32_t num_neighbors = 0;
+		int rv;
 
 		if (lwgeom_is_empty(geoms[p]))
 			continue;
 
-		dbscan_update_context(tree.tree, &cxt, geoms, p, eps);
+		rv = dbscan_update_context(tree.tree, &cxt, geoms, p, eps);
```
dbaston, line 438: Same

dbaston (approved):

liblwgeom/lwgeom_geos_cluster.c
pramsey: I think it can, but I think we also have some bots building with older compilers and generating errors/warnings for doing C99 stuff like that. The discipline of the bots!

PR #725, opened 2023-04-09
Include <math.h> for isnan

fweimer-rh:
This avoids an implicit function declaration and a build failure with future compilers.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


strk:
Thanks, merged as 8340bb63c5904a3a949ab6f62d07bceb40905b1b

PR #727, opened 2023-04-15
Remove invisible unicode in NEWS file

pnorman:
A U+FFFC character got into the NEWS file, causing warning messages in gitea.


strk:
I pushed this commit as of 49b636e4a02116a99c0c42338abf2c0a49b98cc0 -- not sure why github isn't noticing that. Maybe the mirror is broken. @pnorman I suggest you use git.osgeo.org/gitea/postgis/postgis for pull requests :)

PR #728, opened 2023-04-17
[FreeBSD][CI] Update and fix CI

lbartoletti:
FreeBSD 12.3 uses Clang 10 by default. However, there is a new warning flags: -Wnocompound-token-split-by-macro. But, this FreeBSD version will be EOL at 2023-12-31. So it doesn't worth to fix it. rm is a good fix.

BTW, update FreeBSD 13 to 13.2.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -35,10 +35,6 @@ task:
     # - name: 14-CURRENT
     #   freebsd_instance:
     #     image_family: freebsd-14-0-snap
-    - name: 13.1-RELEASE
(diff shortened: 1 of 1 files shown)


strk:
Thanks, pushed as commit f1e368fb3d7b3835e1c83d4e2d6aa8d444a22014

robe2:
This can be closed?  already pushed? or is there more to be done here?

robe2:
Closing this out since we turned off 12.3 testing on Cirrus.

PR #730, opened 2023-04-20
Add CG_StraightSkeleton with distance stored in M

pjanetzek:
Hello this is my first PR to postgis. I've used this addition to generate building roof structures a while ago and currently revisiting this code. Maybe it's useful for other things as well :)

The necessary change has been added to sfcgal in the release 1.3.8 from 2020 https://github.com/Oslandia/SFCGAL/releases
  `Add point_m and is_measured to c api`

### DIFF
--- a/liblwgeom/lwgeom_sfcgal.c
+++ b/liblwgeom/lwgeom_sfcgal.c
@@ -137,8 +137,9 @@ ptarray_from_SFCGAL(const sfcgal_geometry_t *geom, int want3d)
 	assert(geom);
 
 	int is_3d = sfcgal_geometry_is_3d(geom);
+#if POSTGIS_SFCGAL_VERSION >= 10308
 	int is_measured = sfcgal_geometry_is_measured(geom);
-
+#endif
 	switch (sfcgal_geometry_type_id(geom))
 	{
 	case SFCGAL_TYPE_POINT:
@@ -151,9 +152,10 @@ ptarray_from_SFCGAL(const sfcgal_geometry_t *geom, int want3d)
 			point.z = sfcgal_point_z(geom);
 		else if (want3d)
 			point.z = 0.0;
+#if POSTGIS_SFCGAL_VERSION >= 10308
 		if (is_measured)
 			point.m = sfcgal_point_m(geom);
-
+#endif
 		ptarray_set_point4d(pa, 0, &point);
 	}
 	break;
@@ -173,9 +175,10 @@ ptarray_from_SFCGAL(const sfcgal_geometry_t *geom, int want3d)
 				point.z = sfcgal_point_z(pt);
 			else if (want3d)
 				point.z = 0.0;
-                        if (is_measured)
-                                 point.m = sfcgal_point_m(pt);
-
+#if POSTGIS_SFCGAL_VERSION >= 10308
+			if (is_measured)
+				point.m = sfcgal_point_m(pt);
+#endif
 			ptarray_set_point4d(pa, i, &point);
 		}
 	}
@@ -195,9 +198,10 @@ ptarray_from_SFCGAL(const sfcgal_geometry_t *geom, int want3d)
 				point.z = sfcgal_point_z(pt);
 			else if (want3d)
 				point.z = 0.0;
+#if POSTGIS_SFCGAL_VERSION >= 10308
 			if (is_measured)
 				point.m = sfcgal_point_m(pt);
-
+#endif
 			ptarray_set_point4d(pa, i, &point);
 		}
 	}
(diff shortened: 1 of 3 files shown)


liblwgeom/lwgeom_sfcgal.c
```diff
@@ -136,18 +136,23 @@ ptarray_from_SFCGAL(const sfcgal_geometry_t *geom, int want3d)
 
 	assert(geom);
 
+	int is_3d = sfcgal_geometry_is_3d(geom);
+	int is_measured = sfcgal_geometry_is_measured(geom);
```
lbartoletti, line 140: Requires a ifdef SFCGAL >= 1.3.8

According to [Repology](https://repology.org/project/sfcgal/versions) Ubuntu 20.04 and Debian 10 still use SFCGAL 1.3.6

sfcgal/lwgeom_sfcgal.c
```diff
@@ -375,6 +375,30 @@ Datum sfcgal_straight_skeleton(PG_FUNCTION_ARGS)
 	PG_RETURN_POINTER(output);
 }
 
+PG_FUNCTION_INFO_V1(sfcgal_straight_skeleton_distance_in_m);
+Datum sfcgal_straight_skeleton_distance_in_m(PG_FUNCTION_ARGS)
+{
```
lbartoletti, line 380: You should add an ifdef SFCGAL >= 1.3.8

see https://git.osgeo.org/gitea/postgis/postgis/pulls/98/files for example.

lbartoletti:
> I've used this addition to generate building roof structures a while ago and currently revisiting this code. Maybe it's useful for other things as well :)

Nice, I am interested in this contribution, I'm also currently working on this topic :)


Can you update the documentation and add tests? Thanks

robe2:
> > Can you update the documentation and add tests? Thanks
> 
> Thanks for the example - I will add the tests and docs.
> 
> Any recommendation for the postgis function name or should it rather be a parameter of 'ST_StraightSkeleton' to request the geometry with M?

It depends how many people might be using ST_StraightSkeleton in their views and materialized views.
I would think it's one of the more popular functions in postgis_sfcgal.  My general lean would be to have it as an extra argument in ST_StraightSkeleton, but that requires dropping the existing function and adding a new replacement.

Aside from that fewer functions is better.

@lbartoletti any opinion on the matter of if it should be a separate function or just an extra argument.

@pjanetzek If you go the route of a single function, take a look at work done to introduce grid args in ST_Union, ST_Subdivide etc c97b84fb39976897dc1ef621eab5cc6dc1187efe

Note the postgis_before_upgrade steps added. You'd need to do something similar in the sfcgal_before_upgrade.sql.in file.
Since this is the first time we'd be dropping functions in sfcgal, it might not work out of the gate, but I can help if you stumble on this.

lbartoletti:
> @lbartoletti any opinion on the matter of if it should be a separate function or just an extra argument.

yes, you could add an optional boolean with default as False : `sfcgal_straight_skeleton(geometry, use_m_as_distance = False)`

If you use SFCGAL < 1.3.8 or geometry doesn't have M or `use_m_as_distance` you can set the `is_measured` to 1 and use `sfcgal_geometry_straight_skeleton_distance_in_m` instead of `sfcgal_geometry_straight_skeleton`.

The SQL definition could be:

``` SQL
CREATE OR REPLACE FUNCTION ST_StraightSkeleton(geometry, use_m_as_distance boolean DEFAULT false)
       RETURNS geometry
       AS 'MODULE_PATHNAME','sfcgal_straight_skeleton'
       LANGUAGE 'c' IMMUTABLE STRICT PARALLEL SAFE
       COST 100;
```

And the C function something like:

```C
PG_FUNCTION_INFO_V1(sfcgal_straight_skeleton);
Datum sfcgal_straight_skeleton(PG_FUNCTION_ARGS)
{

...


	use_m_as_distance = PG_GETARG_BOOL(1);

       if (sfcgal_version >= 10308) && use_m_as_distance && sfcgal_geometry_is_measured(geom ) result = sfcgal_geometry_straight_skeleton_distance_in_m(geom);
       else result = sfcgal_geometry_straight_skeleton(geom);

...
}
```

robe2 (changes requested):
Please consider the changes that @lbartoletti  detailed in the comments

robe2:
@pjanetzek As of PostGIS 3.5.0 all new functions or functions getting protos will be prefixed with CG and we are deprecating the old ones.

That said, don't bother with the upgrade.sql, since this will be treated as a new function.

Call this CG_StraightSkeleton like for example this new one - https://postgis.net/docs/manual-dev/en/CG_ExtrudeStraightSkeleton.html

I'll put a deprecation notice on ST_StraigthSkeleton and send a link to this new function once you do that.  The backend function name can stay the same, just needs to handle the case when this extra arg is not passed in so will need to check the number of args passed in instead of just assuming there will be one.

PR #731, opened 2023-04-25
Coverage Simplification, Validation, Union

pramsey:
Adds three new functions, 

* ST_CoverageIsValid(geom geometry winset, tolerance float8)
* ST_CoverageSimplify(geom geometry winset, tolerance float8, simplifyBoundary integer)
* ST_CoverageUnion(geom geometry set)

For the purpose of these functions a "coverage" is a set of polygons with perfectly shared edges (no overlaps or gaps and exact vertex matching).

### DIFF
--- a/doc/reference_coverage.xml
+++ b/doc/reference_coverage.xml
@@ -1,0 +1,185 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sect1 id="Coverage">
+  <sect1info>
+    <abstract>
+      <para>These functions operate on collections of polygonal geometry that form "implicit coverages" in which any edges shared by the polygons match exactly. These implicit coverages are fast to compute, and can be operated on with window functions, which retain the implicit topology inside the window partition while altering the edges.</para>
+    </abstract>
+  </sect1info>
+
+  <title>Coverages</title>
+
+  <refentry id="ST_CoverageIsValid">
+    <refnamediv>
+      <refname>ST_CoverageIsValid</refname>
+
+      <refpurpose>Window function that evaluates a set of polygons in the window partition to determine if they form a valid polygonal coverage.</refpurpose>
+    </refnamediv>
+
+    <refsynopsisdiv>
+      <funcsynopsis>
+        <funcprototype>
+          <funcdef>geometry <function>ST_CoverageIsValid</function></funcdef>
+          <paramdef><type>geometry winset </type>
+            <parameter>geom</parameter></paramdef>
+          <paramdef><type>float8 </type>
+            <parameter>tolerance</parameter></paramdef>
+        </funcprototype>
+      </funcsynopsis>
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>Collects all the polygons in the window partition and then checks whether they form a "clean coverage", in which no polygons have any overlap, all vertices on shared edges are identical, and no gaps between polygons exist smaller than the tolerance value.</para>
+      <para>If polygons are found that are not clean relative to their neighbors, the problematic edges will be returned as MULTILINESTRING. For clean polygons the return value will be a NULL. Any non-polygonal or empty geometries on the input will be generate NULL returns.</para>
+      <para>Availability: 3.4.0</para>
+    </refsection>
+
+    <refsection>
+      <title>Examples</title>
+
+      <programlisting>CREATE TABLE coverage (id integer, geom geometry);
+
+INSERT INTO coverage VALUES
+  (1, 'POLYGON((0 0, 10 0, 10.1 5, 10 10, 0 10, 0 0))'),
+  (2, 'POLYGON((10 0, 20 0, 20 10, 10 10, 10.05 5, 10 0))'),
+  (3, 'POLYGON((20 0, 30 0, 30 10, 20 10, 20 0))');
+
+SELECT id, ST_AsText(ST_CoverageIsValid(geom) OVER ())
+  FROM coverage;
+
+ id |           st_astext
+----+--------------------------------
+  1 | LINESTRING(10 0,10.1 5,10 10)
+  2 | LINESTRING(10 10,10.05 5,10 0)
+  3 |
+      </programlisting>
+    </refsection>
+
+    <refsection>
+      <title>See Also</title>
+      <para>
+        <xref linkend="ST_CoverageUnion" />,
+        <xref linkend="ST_CoverageSimplifyVW" />
+      </para>
+    </refsection>
+
+  </refentry>
+
+
+  <refentry id="ST_CoverageSimplifyVW">
+    <refnamediv>
+      <refname>ST_CoverageSimplifyVW</refname>
+
+      <refpurpose>Window function that simplifies the edges of an implicit coverage formed by a set of polygons, applying the Visvalingam–Whyatt algorithm to the edges.</refpurpose>
+    </refnamediv>
+
+    <refsynopsisdiv>
+      <funcsynopsis>
+        <funcprototype>
+          <funcdef>geometry <function>ST_CoverageSimplifyVW</function></funcdef>
+          <paramdef><type>geometry winset </type>
+            <parameter>geom</parameter></paramdef>
+          <paramdef><type>float8 </type>
+            <parameter>tolerance</parameter></paramdef>
+          <paramdef choice="opt"><type>integer </type>
+            <parameter>simplifyBoundary=0</parameter></paramdef>
+        </funcprototype>
+      </funcsynopsis>
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>Collects all the polygons in the window partition, and forms a polgyonal coverage by looking for shared edges between the polygons. Once coverage is formed, the shared edges are simplified using the <ulink url="https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm">Visvalingam–Whyatt algorithm</ulink>. For clean polygonal coverage inputs, this will result in a simplification that is consistent between polygons, and retains the coverage property for the outputs.</para>
+      <para>For invalid coverage inputs, non-shared edges are ignored, and only shared edges are simplified.</para>
+      <para>To only simplify the "internal" edges of the set of polygons (those that are shared by two polytons) set the "simplifyBoundary" parameter to 1.</para>
+      <para>Availability: 3.4.0</para>
+
+    </refsection>
+
+    <refsection>
+      <title>Examples</title>
+
+      <programlisting>CREATE TABLE coverage (id integer, geom geometry);
+
+INSERT INTO coverage VALUES
+  (1, 'POLYGON((0 0, 10 0, 10.1 5, 10 10, 0 10, 0 0))'),
+  (2, 'POLYGON((10 0, 20 0, 20 10, 10 10, 10.1 5, 10 0))'),
+  (3, 'POLYGON((20 0, 30 0, 30 10, 20 10, 20 0))');
+
+SELECT id, ST_AsText(ST_CoverageSimplify(geom, 1.0) OVER ())
+  FROM coverage;
+
+ id |               st_astext
+----+---------------------------------------
+  1 | POLYGON((10 0,10 10,0 10,0 0,10 0))
+  2 | POLYGON((10 0,20 0,20 10,10 10,10 0))
+  3 | POLYGON((20 0,30 0,30 10,20 10,20 0))
+      </programlisting>
+    </refsection>
+
+    <refsection>
+      <title>See Also</title>
+      <para>
+        <xref linkend="ST_CoverageUnion" />,
+        <xref linkend="ST_CoverageIsValid" />
+      </para>
+    </refsection>
+
+  </refentry>
+
+  <refentry id="ST_CoverageUnion">
+    <refnamediv>
+      <refname>ST_CoverageUnion</refname>
+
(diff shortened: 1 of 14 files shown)


robe2:
@pramsey  your documentation is missing mention of minimum required GEOS.

robe2:
Garden crash:

```
SELECT ST_CoverageIsValid( ST_GeomFromText('MULTIPOLYGON (((9 9, 9 1, 1 1, 2 4, 7 7, 9 9)), EMPTY)', 4326)) OVER(ORDER BY random());
```

dr-jts:
> Garden crash:
> 
> ```
> SELECT ST_CoverageIsValid( ST_GeomFromText('MULTIPOLYGON (((9 9, 9 1, 1 1, 2 4, 7 7, 9 9)), EMPTY)', 4326)) OVER(ORDER BY random());
> ```

@robe2 yes, that's a known issue. I'm in the process of fixing it.

robe2:
@pramsey  if you want to commit, I'm fine with that.  GHA latest is passing now after the build.

The berries will fail at their garden test step until the crash issue is fixed in GEOS.

robe2 (approved):

dr-jts:
> The ST_CoverageIsValid doesn't sound like it should return a geometry. The function seems misnamed.

Agreed, this sounds like it should return a boolean.

How about `ST_CoverageInvalidLocation(s)`?  I tend towards the plural form, but either is good.

Aside: we have kicked around the idea of an aggregate function returning a boolean, to report whether a set of polygonal geometries forms a valid coverage.   `ST_CoverageIsValid` would be a good name for this function.  We haven't proceeded on this because it is really just a wrapper over ST_CoverageInvalidLocations, and doesn't return as much information.  But maybe it will become apparent that it is useful in the future.

robe2:
> > The ST_CoverageIsValid doesn't sound like it should return a geometry. The function seems misnamed.
> 
> Agreed, this sounds like it should return a boolean.
> 
> How about `ST_CoverageInvalidLocation(s)`? I tend towards the plural form, but either is good.
> 
> Aside: we have kicked around the idea of an aggregate function returning a boolean, to report whether a set of polygonal geometries forms a valid coverage. `ST_CoverageIsValid` would be a good name for this function. We haven't proceeded on this because it is really just a wrapper over ST_CoverageInvalidLocations, and doesn't return as much information. But maybe it will become apparent that it is useful in the future.

I'm 0.52 for non-plural and 0.48 for plural based on a total weighting of 1

pramsey:
Changed to `ST_CoverageInvalidLocations`

robe2:
@pramsey  I rebuilt the docker image and reran latest.  All tests are passing now.  Anything else you are waiting on before you commit?

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+          <paramdef choice="opt"><type>integer </type>
+            <parameter>simplifyBoundary=0</parameter></paramdef>
+        </funcprototype>
+      </funcsynopsis>
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>Collects all the polygons in the window partition, and forms a polgyonal coverage by looking for shared edges between the polygons. Once coverage is formed, the shared edges are simplified using the <ulink url="https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm">Visvalingam–Whyatt algorithm</ulink>. For clean polygonal coverage inputs, this will result in a simplification that is consistent between polygons, and retains the coverage property for the outputs.</para>
```
dbaston, line 101:
```suggestion
      <para>Simplifies shared edges of a polygonal coverage using the <ulink url="https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm">Visvalingam–Whyatt algorithm</ulink>. For clean polygonal coverage inputs, this will result in a simplification that is consistent between polygons, and retains the coverage property for the outputs.</para>
```

The description of what GEOS is doing internally introduces a distinction between an "implicit coverage" and a "polygonal coverage" that could be confusing and is not relevant to the operation of the PostGIS function, IMO.

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+
+    <refsynopsisdiv>
+      <funcsynopsis>
+        <funcprototype>
+          <funcdef>geometry <function>ST_CoverageSimplify</function></funcdef>
+          <paramdef><type>geometry winset </type>
+            <parameter>geom</parameter></paramdef>
+          <paramdef><type>float8 </type>
+            <parameter>tolerance</parameter></paramdef>
+          <paramdef choice="opt"><type>integer </type>
```
dbaston, line 92: Why not a boolean?

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+      </para>
+    </refsection>
+
+  </refentry>
+
+  <refentry id="ST_CoverageUnion">
+    <refnamediv>
+      <refname>ST_CoverageUnion</refname>
+
+      <refpurpose>Simplifies the edges of an implicit coverage formed by a set of polygons, applying the Visvalingam–Whyatt algorithm to the edges.</refpurpose>
```
dbaston, line 144: Copy-paste error

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+
+ id |           st_astext
+----+--------------------------------
+  1 | LINESTRING(10 0,10.1 5,10 10)
+  2 | LINESTRING(10 10,10.05 5,10 0)
+  3 |
+      </programlisting>
+
+      <programlisting>-- Test entire table for validity
+SELECT true = ALL (
```
dbaston, line 59: This usage is a good argument for `ST_CoverageIsValid`, I think.

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>When the input are a polygonal coverage, the function will provide a high speed merge of adjacent polygons in the aggregation set. This function only works with inputs that are coverages: no overlaps, all shared edges identical down to the vertex level. See <xref linkend="ST_CoverageInvalidLocations" /> for a way to test that a set of polygons are a valid polygonal coverage.</para>
+
+      <note><para>You must ensure that your coverage is valid before running simplification or you may get unexpected results in the output, such as unmerged inputs returning as multipolygons.</para></note>
+
+      <para>Availability: 3.4.0 - requires GEOS &gt;= 3.12.0</para>
```
dbaston, line 164: This is available from GEOS 3.8, I think.

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sect1 id="Coverage">
+  <sect1info>
+    <abstract>
+      <para>These functions operate on collections of polygonal geometry that form "implicit coverages" in which any edges shared by the polygons match exactly. These implicit coverages are fast to compute, and can be operated on with window functions, which retain the implicit topology inside the window partition while altering the edges.</para>
```
dbaston, line 5: I'm not sure what "These implicit coverages are fast to compute" means. What is being computed?

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+            <parameter>tolerance</parameter></paramdef>
+        </funcprototype>
+      </funcsynopsis>
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>Collects all the polygons in the window partition and then checks whether they form a "clean coverage", in which no polygons have any overlap, all vertices on shared edges are identical, and no gaps between polygons exist smaller than the tolerance value.</para>
+      <para>If polygons are found that are not clean relative to their neighbors, the problematic edges will be returned as MULTILINESTRING. For clean polygons the return value will be a NULL. Any non-polygonal or empty geometries on the input will be generate NULL returns.</para>
```
dbaston, line 34:
```suggestion
      <para>If polygons are found that overlap or leave a gap between their neighbors, the problematic edges will be returned as MULTILINESTRING. For clean polygons the return value will be a NULL. Any non-polygonal or empty geometries on the input will be generate NULL returns.</para>
```

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+        </funcprototype>
+      </funcsynopsis>
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>Collects all the polygons in the window partition, and forms a polgyonal coverage by looking for shared edges between the polygons. Once coverage is formed, the shared edges are simplified using the <ulink url="https://en.wikipedia.org/wiki/Visvalingam%E2%80%93Whyatt_algorithm">Visvalingam–Whyatt algorithm</ulink>. For clean polygonal coverage inputs, this will result in a simplification that is consistent between polygons, and retains the coverage property for the outputs.</para>
+      <para>For invalid coverage inputs, non-shared edges are ignored, and only shared edges are simplified.</para>
+      <para>To only simplify the "internal" edges of the set of polygons (those that are shared by two polytons) set the "simplifyBoundary" parameter to 1.</para>
```
dbaston, line 103:
```suggestion
      <para>To only simplify the "internal" edges of the set of polygons (those that are shared by two polygons) set the "simplifyBoundary" parameter to 1.</para>
```

doc/reference_coverage.xml
```diff
@@ -0,0 +1,195 @@
...
+          <paramdef><type>geometry set</type>
+            <parameter>geom</parameter></paramdef>
+        </funcprototype>
+      </funcsynopsis>
+    </refsynopsisdiv>
+
+    <refsection>
+      <title>Description</title>
+
+      <para>When the input are a polygonal coverage, the function will provide a high speed merge of adjacent polygons in the aggregation set. This function only works with inputs that are coverages: no overlaps, all shared edges identical down to the vertex level. See <xref linkend="ST_CoverageInvalidLocations" /> for a way to test that a set of polygons are a valid polygonal coverage.</para>
```
dbaston, line 160:
```suggestion
      <para>When the inputs are a polygonal coverage, the function will provide a high speed union of adjacent polygons in the aggregation set. This function only works with inputs that are coverages: no overlaps, all shared edges identical down to the vertex level. See <xref linkend="ST_CoverageInvalidLocations" /> for a way to test that a set of polygons are a valid polygonal coverage.</para>
```

doc/reference_coverage.xml
dr-jts: The idea is to leave the option open for a "Simplify Outer Only" choice as well.  But not sure that will ever eventualize, and I agree that a `boolean` is simpler.  

We could make this a boolean meaning "simplify inner edges only" with default `false`.  If needed another boolean option meaning "simplify outer edges only" with default `false`.

doc/reference_coverage.xml
dr-jts: Essentially all polygonal coverage operations are faster to compute if the coverage topology is taken into account.  

Perhaps better to say "fast to process".   The idea is to alert users to the benefit of modelling data as a coverage, and using coverage operations.

doc/reference_coverage.xml
dr-jts: well, it's actually a bit more than that - adjacent polygons must be edge-matched as well.

I like using the term "coverage-valid" to encapsulate all this conditions.

I'm intending to write a section earlier in the manual outlining the concept of a "simple features coverage".  It will include a detailed discussion of coverage validity (similar to the one [here](http://lin-ear-th-inking.blogspot.com/2022/07/polygonal-coverages-and-operations-in.html)).

doc/reference_coverage.xml
dbaston: Yeah, the flag makes sense for GEOS where we can't have function overloads, but PostGIS doesn't have that limitation.

doc/reference_coverage.xml
dbaston: Not backported, but yes: https://github.com/libgeos/geos/blob/main/NEWS.md#changes-in-380

doc/reference_coverage.xml
dbaston: Maybe better to say "where polygons do not form a valid coverage, the problematic edges...". My issue was just with the term "clean" which means different things to different people.

doc/reference_coverage.xml
pramsey: If I make it actually available as of GEOS 3.8 I need a whole special set of ifdefs just for coverage union. I figured on just introducing all coverage functions as of a GEOS 3.12 ifdef. Is that alright?

dbaston: Don't you already have it ifdeffed on 3.8? https://github.com/postgis/postgis/pull/731/files#diff-cd95a4eccec3e850213b9d8ec7fd08a2365b44458f1e71f65aa73bbb799dc3d0R798

postgis/lwgeom_window.c
```diff
@@ -451,3 +451,423 @@ Datum ST_ClusterKMeans(PG_FUNCTION_ARGS)
...
+}
+
+/**********************************************************************
+ * ST_CoverageUnion(geometry[])
+ *
+ */
+
+Datum ST_CoverageUnion(PG_FUNCTION_ARGS);
+PG_FUNCTION_INFO_V1(ST_CoverageUnion);
+Datum ST_CoverageUnion(PG_FUNCTION_ARGS)
```
dbaston, line 796: Would this make more sense in `lwgeom_geos.c` ?

pramsey: I dunno that the extra layer of indirection is a big win? What's the argument for?

pramsey: Potato / potato. Could move it but then it would be separate from its coverage friends...

PR #732, opened 2023-05-15
Remove Outdated Doc Sections

pramsey:
There are a number of sections of documentation that should be edited down or just plain removed. Practical things like "how to do it" should go into the web site and the workshop. Reference material and theory of operation like function docs and format specifications and de9im descriptions, should stay in the docbook.

Starting this branch with the remains of the FAQ removal, as the "up to date" entries have already been copied and updated into the web site section.


pramsey:
Similarly, the entire `using_postgis` section is not worth retaining. Any MapServer usage information worth preserving should be upstreamed to the mapserver doco. I'm not sure any of it is worth saving, but will give it a quick review. The JDBC code was rarely used even when it was fresh, but it's decades old now. The empty sections on C programming speak for themselves.

pramsey:
I think the raster faq can be deleted after a pass to pull current useful advice into tips and faq entries on the web site.

robe2:
@pramsey All seems fine to me. Only concern I have is that we have no mechanism on site to translate the FAQs, and that section has been largely translated already.  But then again many of the FAQ are a bit out-dated so I do like the idea of putting in a centralized location apart from the docs, cause it's rarely reviewed as part of the docs.

The practical things stuff, yah that I feel should go in the workshop whatever left that is relevant or spinkled as tips in the website.

robe2 (approved):
You already committed this though didn't you, or you thinking of removing in the other branches?

Komzpa:
+1, FAQ entries about PostGIS 0.6 need to go.

robe2:
> Shouldn't be committed? Just on my working branch, ATM

ah okay.  Not sure where I was looking so nevermind.

PR #733, opened 2023-05-23
Proj Utilities for Accessing Metadata

pramsey:


robe2:
What's this for params->celestial_body_name = "Earth" .  Does that mean only Earth is supported?

pramsey:
Yes, I did not bother to expose that knob.

robe2:
> Yes, I did not bother to expose that knob.

I see so no Mars or Venus for now? What will I tell my people.

pramsey:
When the colonization starts, we can revisit.

robe2:
@pramsey are you waiting on something to commit this.  Hurry up before @strk has an idea.

PR #735, opened 2023-06-21
ST_LargestEmptyCircle

pramsey:
See https://trac.osgeo.org/postgis/ticket/5373


pramsey:
```
SELECT 
  st_astext(center) as center, 
  st_astext(nearest) as nearest,  
  radius 
FROM st_largestemptycircle('MULTIPOINT ((100 100), (100 200), (200 200), (200 100))', 0.001);
```

Komzpa:
Why does this need to be a separate function from https://postgis.net/docs/ST_MaximumInscribedCircle.html ? The subtle semantic difference between the last example from ST_MaximumInscribedCircle and this escapes me unfortunately.

dr-jts:
> Why does this need to be a separate function from https://postgis.net/docs/ST_MaximumInscribedCircle.html ? 

The difference between **Maximum Inscribed Circle** and **Largest Empty Circle** is that MIC determines circles lying _inside_ polygons, whereas LEC determines circles lying in the empty space between obstacle geometries (points, line or polygons).  Also, LEC supports specifying a custom polygonal boundary, whereas MIC does not because the boundary is the input polygon(s) by definition.

Currently `ST_MaximumInscribedCircle` is overloaded to accept lines and points and compute their LargestEmptyCircle.  But the different signature for LEC  (i.e. the optional boundary polygon) and the difference in polygon handling requires a separate function.

(Note: at the moment the LEC code doesn't actually support using polygons as obstacles, but this should be a fairly  easy enhancement.)

We could think about removing the point and line support from `ST_MaximumInscribedCircle`, unless that is too much of a change.

PR #737, opened 2023-07-06
Allow polygonization in ST_Contour

pramsey:
Also add interruptability to the GDAL contour operation, since polygonization takes way way longer than linear mode. References #604


Developer66:
Hello!
I have now used the new attribute `polygonize`. With the Gdal call, however, a `min_elev` and `max_elev` should come out.
But with the sql only `value` comes back, which is always 0.
Do I have an error in my SQL or is this a bug?

Thanks for the feedback and implementation.

My Sql:
`SELECT
    (st_contour(rast, level_interval := 0.25, polygonize := true)).*
FROM raster
         WHERE id = 1`

pramsey:
Well, the return type for the function is `RETURNS table(geom geometry, id integer, value float8)` so there isn't room there for a min and a max. I think I need to pass the `ELEV_FIELD_MIN` option into the [call](https://gdal.org/doxygen/gdal__alg_8h.html#a03b3ec0026fe8b23d85c7c4d920613d9) in order to even get a decent value back.

Open question is whether to change the return value to have the min/max or to just return the min in the `value` slot. (I am resisting calling it "elevation" because there's no guarantee people are contouring elevation.) Changing the return signature would be a backwards incompatibility so I'm not super stoked to do that.

PR #744, opened 2023-09-16
Don't create docbook.css for the HTML manual

cjmayo:
We have our own style.css.

---

The single page manual includes the style link, but the Makefile does not install docbook.css.
The chunked manual does install docbook.css  -  and indeed it is used as a target in the Makefile - but does style.css contain all the used styles?

PR as draft because maybe it is OK for the single page manual but the chunked...

### DIFF
--- a/doc/Makefile.in
+++ b/doc/Makefile.in
@@ -106,6 +106,7 @@ XSLTPROC_COMMONOPTS= \
 	--param header.rule 0 \
 	--param footer.rule 0 \
 	--param table.borders.with.css 1 \
+	--stringparam docbook.css.source "" \
 	--stringparam html.ext .html \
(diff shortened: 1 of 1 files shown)


robe2:
@strk I leave it up to you decide.  I'm not quite sure what this pull request is doing

strk:
it's dropping some CSS, we need to build and install the manual in a clean space to really see the effects of that drop. Drops are generally good to me, but @cjmayo has this PR marked as draft so I guess we shouldn't bother until it's marked ready for review

robe2:
@strk So you okay with this now?

strk:
Yes, go for it, we'll find out if anything breaks

robe2:
Committed at 305c7a8

PR #746, opened 2023-09-28
Reduce the size of the chunked HTML manual main table of contents

cjmayo:
Since [1] reference entries have been included in the main table of contents as level 2 information. Stop this by setting the TOC depth to 1.

[1]
acdfda8ec ("Convert PostGIS manual to DocBook 5.0", 2023-07-14)

### DIFF
--- a/doc/Makefile.in
+++ b/doc/Makefile.in
@@ -95,6 +95,7 @@ XSLTPROC_COMMONOPTS= \
 	--param section.label.includes.component.label 1 \
 	--param chunk.section.depth 0 \
 	--param generate.section.toc.level 2 \
+	--param toc.section.depth 1 \
 	--param funcsynopsis.style kr \
 	--param admon.graphics 1 \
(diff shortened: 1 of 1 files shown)


strk:
As part as the docbook version upgrade all "sect#" elements (numbered) were converted to "section" (unnumbered) so now depth level is the only indication. I'm happy with less noise in the TOC but would let @robe2 decide :)

robe2:
> As part as the docbook version upgrade all "sect#" elements (numbered) were converted to "section" (unnumbered) so now depth level is the only indication. I'm happy with less noise in the TOC but would let @robe2 decide :)

I'm fine with that too as long as it doesn't break anything

strk:
@cjmayo how does generate.section.toc.level differ from toc.section.depth ?

strk:
Pushed as 0750de748f0510e208854f1b931bf3afe9fc9f8a thank you

PR #748, opened 2023-10-11
Fix tar command in short version of installation docs

Spone:
The command in the "Short Version" section did not match the one below.


robe2:
Thanks for the catch.  Confirmed this is an issue.

robe2:
committed at https://github.com/postgis/postgis/commit/b178091a9833febf41d6d9f388804a414a26421a

PR #749, opened 2023-10-28
Allow promoting column as an id in ST_AsGeoJsonRow

jtojnar:
As per GeoJSON RFC, the id should go directly to the feature object rather than to properties:

> If a Feature has a commonly used identifier, that identifier SHOULD be included as a member of the Feature object with the name "id"

Let’s add an argument that will allow designating a column as an identifier so that people do not have to tediously build the JSON object manually.

References https://trac.osgeo.org/postgis/ticket/5596

### DIFF
--- a/doc/reference_output.xml
+++ b/doc/reference_output.xml
@@ -623,6 +623,7 @@ F000000000000000000000000000000000000000000000000');
 				<paramdef choice="opt"><type>text </type> <parameter>geom_column=""</parameter></paramdef>
 				<paramdef choice="opt"><type>integer </type> <parameter>maxdecimaldigits=9</parameter></paramdef>
 				<paramdef choice="opt"><type>boolean </type> <parameter>pretty_bool=false</parameter></paramdef>
+                <paramdef choice="opt"><type>text </type> <parameter>id_column=''</parameter></paramdef>
 			</funcprototype>
 			<funcprototype>
 				<funcdef>text <function>ST_AsGeoJSON</function></funcdef>
@@ -703,6 +704,8 @@ Conversely, passing the parameter will save column type lookups.
 			  </itemizedlist>
 			</para>
 
+            <para>The <varname>id_column</varname> parameter is used to set the "id" member of the returned GeoJSON features. As per GeoJSON RFC, this SHOULD be used whenever a feature has a commonly used identifier, such as a primary key. When not specified, the produced features will not get an "id" member and any columns other than the geometry, including any potential keys, will just end up inside the feature’s "properties" member.</para>
+
             <para>The GeoJSON specification states that polygons are oriented using the Right-Hand Rule,
             and some clients require this orientation.
             This can be ensured by using <xref linkend="ST_ForcePolygonCCW"/>.
@@ -727,6 +730,7 @@ Conversely, passing the parameter will save column type lookups.
 			<para role="changed" conformance="2.0.0">Changed: 2.0.0 support default args and named args.</para>
 			<para role="changed" conformance="3.0.0">Changed: 3.0.0 support records as input</para>
 			<para role="changed" conformance="3.0.0">Changed: 3.0.0 output SRID if not EPSG:4326.</para>
+            <para role="changed" conformance="3.5.0">Changed: 3.5.0 allow specifying the column containing the feature id</para>
 			<para>&Z_support;</para>
 	  </refsection>
 
@@ -736,34 +740,20 @@ Conversely, passing the parameter will save column type lookups.
 <para>Generate a FeatureCollection:</para>
 <programlisting>SELECT json_build_object(
     'type', 'FeatureCollection',
-    'features', json_agg(ST_AsGeoJSON(t.*)::json)
+    'features', json_agg(ST_AsGeoJSON(t.*, id_column => 'id')::json)
     )
 FROM ( VALUES (1, 'one', 'POINT(1 1)'::geometry),
               (2, 'two', 'POINT(2 2)'),
               (3, 'three', 'POINT(3 3)')
      ) as t(id, name, geom);</programlisting>
-<screen>{"type" : "FeatureCollection", "features" : [{"type": "Feature", "geometry": {"type":"Point","coordinates":[1,1]}, "properties": {"id": 1, "name": "one"}}, {"type": "Feature", "geometry": {"type":"Point","coordinates":[2,2]}, "properties": {"id": 2, "name": "two"}}, {"type": "Feature", "geometry": {"type":"Point","coordinates":[3,3]}, "properties": {"id": 3, "name": "three"}}]}</screen>
+<screen>{"type" : "FeatureCollection", "features" : [{"type": "Feature", "geometry": {"type":"Point","coordinates":[1,1]}, "id": 1, "properties": {"name": "one"}}, {"type": "Feature", "geometry": {"type":"Point","coordinates":[2,2]}, "id": 2, "properties": {"name": "two"}}, {"type": "Feature", "geometry": {"type":"Point","coordinates":[3,3]}, "id": 3, "properties": {"name": "three"}}]}</screen>
 
 <para>Generate a Feature:</para>
-		<programlisting>SELECT ST_AsGeoJSON(t.*)
+		<programlisting>SELECT ST_AsGeoJSON(t.*, id_column => 'id')
 FROM (VALUES (1, 'one', 'POINT(1 1)'::geometry)) AS t(id, name, geom);</programlisting>
 <screen>                                                  st_asgeojson
 -----------------------------------------------------------------------------------------------------------------
- {"type": "Feature", "geometry": {"type":"Point","coordinates":[1,1]}, "properties": {"id": 1, "name": "one"}}
-</screen>
-
-<para>An alternate way to generate Features with an <varname>id</varname> property
-is to use JSONB functions and operators:</para>
-		<programlisting>SELECT jsonb_build_object(
-    'type',       'Feature',
-    'id',         id,
-    'geometry',   ST_AsGeoJSON(geom)::jsonb,
-    'properties', to_jsonb( t.* ) - 'id' - 'geom'
-    ) AS json
-FROM (VALUES (1, 'one', 'POINT(1 1)'::geometry)) AS t(id, name, geom);</programlisting>
-<screen>                                                  json
------------------------------------------------------------------------------------------------------------------
- {"id": 1, "type": "Feature", "geometry": {"type": "Point", "coordinates": [1, 1]}, "properties": {"name": "one"}}
+ {"type": "Feature", "geometry": {"type":"Point","coordinates":[1,1]}, "id": 1, "properties": {"name": "one"}}
 </screen>
 
 <para>Don't forget to transform your data to WGS84 longitude, latitude to conform with the GeoJSON specification:</para>
--- a/postgis/postgis_before_upgrade.sql
+++ b/postgis/postgis_before_upgrade.sql
@@ -173,3 +173,9 @@ BEGIN
 END;
 $$;
 
+-- FUNCTION ST_AsGeoJson added `id_column` optional argument in 3.5.0.
+SELECT _postgis_drop_function_by_identity
+	(
+	'ST_AsGeoJson',
+	'r record, geom_column text, maxdecimaldigits integer, pretty_bool boolean'
+	);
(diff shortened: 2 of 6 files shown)


jtojnar:
I have not opened trac issue since I have not received mantra for osgeo account yet.

Tests pass (`PGUSER=postgres make check`) for me locally.

regress/core/out_geojson_expected
```diff
@@ -27,4 +27,11 @@ gj04|4|{"d": "2004-04-04", "f": 4.4, "g": {"type": "GeometryCollection", "geomet
...
 gj04|7|{"d": "2007-07-07", "f": 7.7, "g": {"type": "GeometryCollection", "geometries": [{"type": "Point", "coordinates": []}, {"type": "Point", "coordinates": [1, 2]}]}, "i": 7, "t": "seven"}
+gj05|1|{"type": "Feature", "geometry": {"type":"Point","coordinates":[42,42]}, "id": 1, "properties": {"f": 1.1, "t": "one", "d": "2001-01-01"}}
+gj05|2|{"type": "Feature", "geometry": {"type":"LineString","coordinates":[[42,42],[45,45]]}, "id": 2, "properties": {"f": 2.2, "t": "two", "d": "2002-02-02"}}
+gj05|3|{"type": "Feature", "geometry": {"type":"Polygon","coordinates":[[[42,42],[45,45],[45,42],[42,42]]]}, "id": 3, "properties": {"f": 3.3, "t": "three", "d": "2003-03-03"}}
+gj05|4|{"type": "Feature", "geometry": {"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[42,42]}]}, "id": 4, "properties": {"f": 4.4, "t": "four", "d": "2004-04-04"}}
+gj05|5|{"type": "Feature", "geometry": {"type":"Point","coordinates":[]}, "id": 5, "properties": {"f": 5.5, "t": "five", "d": "2005-05-05"}}
+gj05|6|{"type": "Feature", "geometry": {"type": null}, "id": 6, "properties": {"f": 6.6, "t": "six", "d": "2006-06-06"}}
```
jtojnar, line 35: I also noticed that null geometry does not conform to RFC 7946 – [the `type` member must be one of seven geometry type names](https://www.rfc-editor.org/rfc/rfc7946#section-3.1). The proper designation is [`"geometry": null`](https://www.rfc-editor.org/rfc/rfc7946#section-3.2) but I am not sure it would not be a BC break. And it is not related to this patch anyway.

doc/reference_output.xml
```diff
@@ -623,6 +623,7 @@ F000000000000000000000000000000000000000000000000');
 				<paramdef choice="opt"><type>text </type> <parameter>geomcolumnname</parameter></paramdef>
```
jtojnar, line 623: On another unrelated note, this is wrong and there is patch fixing it on Trac: https://trac.osgeo.org/postgis/ticket/4764

strk:
> I have not opened trac issue since I have not received mantra for osgeo account yet.

You should have received the mantra by now, could you file that ticket and reference it here in comments and in commit log please ?

The mail was sent to you on `Sat, 28 Oct 2023 18:49:23 -0400` by `Vicky Vergara`. Check also in the spam folder.

jtojnar:
I have opened https://trac.osgeo.org/postgis/ticket/5596

regress/core/out_geojson_expected
jtojnar: Opened https://trac.osgeo.org/postgis/ticket/5597 for that one

robe2:
@pramsey have any issues with this pull request?

robe2:
@strk  at a glance, most of this is in order, @jtojnar  did a good job of updating the documentation and adding tests.  My only concern is I don't think our plumbing is set up to handle a change in args of function correct?  So adding to drop the old function and create a new is required.  I'm going to shove this thru one of our upgrade testers to see if it breaks.  I suspect it will.

robe2:
As I feared it fails the upgrade test - https://woodie.osgeo.org/repos/30/pipeline/1540/20


@jtojnar  this will cause some upgrade burdens.  There are two approaches to go with this

1) Don't remove the old signature, just add a new signature marking it as Availability: 3.5.0

approach 2) a) overwrite the old signature as you are doing now
b) Add a drop line to https://github.com/postgis/postgis/blob/master/postgis/postgis_before_upgrade.sql

The benefit of option one is I think it will cause less upgrade pain for folks who are using ST_AsGeoJSON row variant in their views and materialized views, but does cause some extra baggage for us to carry around

My vote is for option 1 even though it will cause more pain for folks.

@strk,  @pramsey what's your thought on the matter

robe2 (changes requested):
Needs to add a drop statement here - https://github.com/postgis/postgis/blob/master/postgis/postgis_before_upgrade.sql  or define a whole new function leaving the old signature intact.

jtojnar:
> but does cause some extra baggage for us to carry around

Would not the `drop` be extra baggage to carry around just as well?

pramsey (approved):
Either approach to the new SQL signature is fine with me

jtojnar:
> Don't remove the old signature, just add a new signature marking it as Availability: 3.5.0

Actually, it looks like just adding a new signature is what is currently happening. I tried the following:

```postgresql
CREATE OR REPLACE FUNCTION dup(int) RETURNS TABLE(f1 int, f2 text)
    AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
    LANGUAGE SQL;
CREATE OR REPLACE FUNCTION dup(int, int=4) RETURNS TABLE(f1 int, f2 text)
    AS $$ SELECT $1 + $2, CAST($1 AS text) || ' is text' $$
    LANGUAGE SQL;
select dup(1);
```

And got the same error as in the upgrade check:

```
ERROR: function dup(integer) is not unique
```

---

Pushed the suggested removal.

robe2:
> > Don't remove the old signature, just add a new signature marking it as Availability: 3.5.0
> 
> Actually, it looks like just adding a new signature is what is currently happening. I tried the following:
> 
> ```sql
> CREATE OR REPLACE FUNCTION dup(int) RETURNS TABLE(f1 int, f2 text)
>     AS $$ SELECT $1, CAST($1 AS text) || ' is text' $$
>     LANGUAGE SQL;
> CREATE OR REPLACE FUNCTION dup(int, int=4) RETURNS TABLE(f1 int, f2 text)
>     AS $$ SELECT $1 + $2, CAST($1 AS text) || ' is text' $$
>     LANGUAGE SQL;
> select dup(1);
> ```
> 
> And got the same error as in the upgrade check:
> 
> ```
> ERROR: function dup(integer) is not unique
> ```
> 
> Pushed the suggested removal.

I forgot to mention for the new signature, you can't have it have a default.  That default needs to be removed otherwise you do get a duplicate.

robe2:
I see you went with the drop approach.  At a glance all looks in order to me.  I'll do one final upgrade check and accept unless @strk  has any opinions

robe2:
still failing with not unique during upgrade  https://woodie.osgeo.org/repos/30/pipeline/1540/20

The postgis_before_drop should have taken care of this.  I'm investigating now.

```
regress/core/out_geojson .. failed (diff expected obtained: /tmp/pgis_reg/test_114_diff)
--- /woodpecker/src/git.osgeo.org/gitea/postgis/postgis/regress/core/out_geojson_expected	2023-11-27 18:07:13.029181299 +0000
+++ /tmp/pgis_reg/test_114_out	2023-11-27 19:02:38.189018095 +0000
@@ -6,13 +6,7 @@
 gj01|5|{"type":"Point","coordinates":[]}
 gj01|6|
 gj01|7|{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[]},{"type":"Point","coordinates":[1,2]}]}
-gj02|1|{"type": "Feature", "geometry": {"type":"Point","coordinates":[42,42]}, "properties": {"i": 1, "f": 1.1, "t": "one", "d": "2001-01-01"}}
-gj02|2|{"type": "Feature", "geometry": {"type":"LineString","coordinates":[[42,42],[45,45]]}, "properties": {"i": 2, "f": 2.2, "t": "two", "d": "2002-02-02"}}
-gj02|3|{"type": "Feature", "geometry": {"type":"Polygon","coordinates":[[[42,42],[45,45],[45,42],[42,42]]]}, "properties": {"i": 3, "f": 3.3, "t": "three", "d": "2003-03-03"}}
-gj02|4|{"type": "Feature", "geometry": {"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[42,42]}]}, "properties": {"i": 4, "f": 4.4, "t": "four", "d": "2004-04-04"}}
-gj02|5|{"type": "Feature", "geometry": {"type":"Point","coordinates":[]}, "properties": {"i": 5, "f": 5.5, "t": "five", "d": "2005-05-05"}}
-gj02|6|{"type": "Feature", "geometry": {"type": null}, "properties": {"i": 6, "f": 6.6, "t": "six", "d": "2006-06-06"}}
-gj02|7|{"type": "Feature", "geometry": {"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[]},{"type":"Point","coordinates":[1,2]}]}, "properties": {"i": 7, "f": 7.7, "t": "seven", "d": "2007-07-07"}}
+ERROR:  function st_asgeojson(g) is not unique at character 19
 gj03|1|{"g":{"type":"Point","coordinates":[42,42]},"i":1,"f":1.1,"t":"one","d":"2001-01-01"}
 gj03|2|{"g":{"type":"LineString","coordinates":[[42,42],[45,45]]},"i":2,"f":2.2,"t":"two","d":"2002-02-02"}
 gj03|3|{"g":{"type":"Polygon","coordinates":[[[42,42],[45,45],[45,42],[42,42]]]},"i":3,"f":3.3,"t":"three","d":"2003-03-03"}
@@ -34,4 +28,4 @@
 gj05|5|{"type": "Feature", "geometry": {"type":"Point","coordinates":[]}, "id": 5, "properties": {"f": 5.5, "t": "five", "d": "2005-05-05"}}
 gj05|6|{"type": "Feature", "geometry": {"type": null}, "id": 6, "properties": {"f": 6.6, "t": "six", "d": "2006-06-06"}}
 gj05|7|{"type": "Feature", "geometry": {"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[]},{"type":"Point","coordinates":[1,2]}]}, "id": 7, "properties": {"f": 7.7, "t": "seven", "d": "2007-07-07"}}
-4695|{"type": "Feature", "geometry": {"type":"Point","crs":{"type":"name","properties":{"name":"EPSG:2227"}},"coordinates":[0,1]}, "properties": {"v": 1}}
+ERROR:  function st_asgeojson(record) is not unique at character 16
```

postgis/postgis_before_upgrade.sql
```diff
@@ -173,3 +173,9 @@ BEGIN
 END;
 $$;
 
+-- FUNCTION ST_AsGeoJson added `id_column` optional argument in 3.5.0.
+SELECT _postgis_drop_function_by_identity
+	(
+	'ST_AsGeoJson',
+	'r record, geom_column text, maxdecimaldigits integer, pretty_bool bool'
```
robe2, line 180: 'r record, geom_column text, maxdecimaldigits integer, pretty_bool boolean'

change bool to boolean.  It's not dropping cause the postgresql canonical signature standardizes on boolean

robe2 (changes requested):

postgis/postgis_before_upgrade.sql
jtojnar: Fixed.

postgis/postgis_before_upgrade.sql
```diff
@@ -173,3 +173,9 @@ BEGIN
 END;
 $$;
 
+-- FUNCTION ST_AsGeoJson added `id_column` optional argument in 3.5.0.
+SELECT _postgis_drop_function_by_identity
```
strk, line 177: For the record:  by adding a new parameter we actually created a NEW SIGNATURE (rather than changing the identity of an existing signature). So we'd have needed a different function (drop by signature) BUT the simplest way to deal with upgrades is the use a `Replaces` comment, like it was done to fix the bug which resulted by this misues of _postgis_drop_function_by_identity, see

https://trac.osgeo.org/postgis/ticket/5623#comment:5

https://git.osgeo.org/gitea/postgis/postgis/commit/a6f4fa8282d3fef7e298f66cb7db96674486f028

PR #750, opened 2023-10-28
Fix shebangs in utils

jtojnar:
Shebang need to be on the first line or the script will fail to execute on some systems:

    make: ./utils/check_tests_enabled.sh: No such file or directory

### DIFF
--- a/utils/check_lang_enabled.sh
+++ b/utils/check_lang_enabled.sh
@@ -1,4 +1,3 @@
-
 #!/bin/sh
 
 usage() {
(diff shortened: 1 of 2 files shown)


strk (commented):
Thanks, I don't know how I missed this (not approving to avoid the temptation of clicking the green button, as we're using git.osgeo.org/gitea/postgis as the official forge)

jtojnar:
Should I reopen this there now that I have OSGeo account?

robe2:
> Should I reopen this there now that I have OSGeo account?

Not necessary.  I'll commit shortly to 3.3 - 3.5 branches

robe2:
I guess these were new so no need to add to prior branches.

PR #756, opened 2024-01-20
lwcurvepoly_from_wkb_state(): avoid potentially very deeply neested geometries causing stack call overflow

rouault:
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65973


pramsey:
Merci!

PR #762, opened 2024-02-20
ST_CurveN and ST_NumCurves plus consistency

pramsey:
Accessors for the components of CompoundCurve, and changes to geometry accessors relative to CompoundCurve to respect that it is a unitary object, just like LineString and CircularString.


regress/core/sql-mm-compoundcurve.sql
```diff
@@ -240,3 +240,25 @@ SELECT 'minpoints issues - pass', encode(ST_AsBinary(ST_GeomFromText('COMPOUNDCU
...
+SELECT 'curve accessors 01',
+  ST_NumGeometries(geom) as numgeometries,
+  ST_GeometryType(ST_GeometryN(geom,1)) as geometryn_1,
+  ST_GeometryType(ST_CurveN(geom,1)) as curven_1,
+  ST_GeometryN(geom,2) AS geometryn_2,
+  ST_NumCurves(geom) AS numcurves,
+  ST_GeometryType(ST_CurveN(geom,1)) as curven_1,
+  ST_GeometryType(ST_CurveN(geom,2)) as curven_2,
+  ST_GeometryType(ST_CurveN(geom,3)) as curven_3,
+  ST_GeometryType(ST_CurveN(geom,3)) as curven_4,
```
rouault, line 262: this repeats the curven_3 query. Is that indented, or was it rather ST_CurveN(geom,4) ?

doc/reference_accessor.xml
```diff
@@ -1503,6 +1500,103 @@ FROM (SELECT ST_BuildArea(
...
+        <para>&Z_support;</para>
+      </refsection>
+
+      <refsection>
+        <title>Examples</title>
+
+        <programlisting>
+-- Returns 3
+SELECT ST_NumCurves('COMPOUNDCURVE(
+    LINESTRING(2 2, 2.5 2.5),
```
rouault, line 1531: My understanding of https://github.com/postgis/postgis/blob/8ed84517a9b86c86724504d1b6a0f0c1ccd86cf6/doc/bnf-wkt.txt is that the explicit LINESTRING() keyword within a COMPOUNDCURVE is not allowed, and a PostGIS, OGR, etc. extension. It should probably not be advertized here (and below)

PR #765, opened 2024-02-26
Make sure our vendored flatbuffers copy has a unique namespace

rouault:
Cf https://lists.osgeo.org/pipermail/gdal-dev/2024-February/058544.html and https://github.com/OSGeo/gdal/pull/9313


pramsey:
Any reason not to back-port this a couple versions?

rouault:
> Any reason not to back-port this a couple versions?

Should be safe to do

PR #771, opened 2024-04-24
Fix typo in ST_Touches math expression

rootjalex:
I believe this fixes a typo, as the current expression says that the interiors _do_ intersect.


dr-jts:
Fixed by https://github.com/postgis/postgis/commit/8bb45e3552de3bf93d3cc064d206d10379fbfa97

PR #772, opened 2024-05-20
[CI][macOS]: disable protobuf

lbartoletti:
On macOS, [protobuf-c](https://formulae.brew.sh/formula/protobuf-c) uses [protobuf@26](https://formulae.brew.sh/formula/protobuf) whereas [protobuf@3](https://formulae.brew.sh/formula/protobuf@3) should be used.

Until a proper fix on *.proto file is applied, let's disable protobuf on macOS CI.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -79,10 +79,9 @@ task:
         pkg-config
         postgresql@14
         proj
-        protobuf-c
         sfcgal
     - ./autogen.sh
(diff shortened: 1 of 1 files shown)


strk:
Merged as 1cfeb88f9a4786ded15db1f4ede6821f3184316e after rewording the commit to include the reference to the trac ticket https://trac.osgeo.org/postgis/ticket/5730

PR #775, opened 2024-06-13
[3.4 backport] ST_EstimatedExtent(geography)

pramsey:


strk:
Let's first make sure stats are actually right in master branch first ? 
https://trac.osgeo.org/postgis/ticket/5750

I'm still thinking we could return NULL and call it a bugfix, in anything but master (3.5.0dev)

PR #777, opened 2024-07-18
MSVC Compat 1: Reconcile attributes and math macros

ralian:
- Define _USE_MATH_DEFINES on MSVC wherever constants like M_PI are needed
- Do not allow use of format attribute outside of GCC

### DIFF
--- a/liblwgeom/lwgeom_log.h
+++ b/liblwgeom/lwgeom_log.h
@@ -126,7 +126,11 @@
  * For debugging, use LWDEBUG() or LWDEBUGF().
  * @ingroup logging
  */
-void lwnotice(const char *fmt, ...) __attribute__ ((format (printf, 1, 2)));
+void lwnotice(const char *fmt, ...)
+#ifdef __GNUC__
+  __attribute__ ((format (printf, 1, 2)))
+#endif
+;
 
 /**
  * Write a notice out to the error handler.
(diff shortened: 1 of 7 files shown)


liblwgeom/lwgeom_log.h
```diff
@@ -126,7 +126,11 @@
  * For debugging, use LWDEBUG() or LWDEBUGF().
  * @ingroup logging
  */
-void lwnotice(const char *fmt, ...) __attribute__ ((format (printf, 1, 2)));
+void lwnotice(const char *fmt, ...)
+#ifdef __GNUC__
```
lbartoletti, line 130: Also available with CLANG https://clang.llvm.org/docs/AttributeReference.html#format (but with __printf__ in the doc). I cannot test this week.

ralian: I could be wrong for specific cases, but I believe `__GNUC__` is also defined on clang - as it is meant to be compatible with GCC in almost every way. See this rather humorous SO post: https://stackoverflow.com/questions/38499462/how-to-tell-clang-to-stop-pretending-to-be-other-compilers

I still need to do some more testing as well, though. Let me know if you uncover anything I missed in your own testing.

strk:
See https://git.osgeo.org/gitea/postgis/postgis/pulls/208 for more comments (or best file a ticket so discussion remains there) - https://trac.osgeo.org/postgis

PR #779, opened 2024-09-24
doc: unify gdal_enabled_drivers and enable_outdb_rasters

nikolai-b:
I tried the 

```sql
ALTER SYSTEM postgis.enable_outdb_rasters = true;
```
from the docs but this errors as it is missing `SET`.  I've added that and tried to make `gdal_enabled_drivers` and `enable_outdb_rasters` docs use similar language.

### DIFF
--- a/doc/reference_guc.xml
+++ b/doc/reference_guc.xml
@@ -149,30 +149,25 @@ SET postgis.gdal_datapath = 'C:/Program Files/PostgreSQL/9.3/gdal-data';</progra
 
             <refsection>
                 <title>Examples</title>
-                <para>Set and reset <varname>postgis.gdal_enabled_drivers</varname></para>
-
-                <para>Sets backend for all new connections to database</para>
-                <programlisting>ALTER DATABASE mygisdb SET postgis.gdal_enabled_drivers TO 'GTiff PNG JPEG';</programlisting>
-
-                <para>Sets default enabled drivers for all new connections to server. Requires super user access and PostgreSQL 9.4+.
-                    Also note that database, session, and user settings override this.</para>
-                <programlisting>ALTER SYSTEM SET postgis.gdal_enabled_drivers TO 'GTiff PNG JPEG';
-SELECT pg_reload_conf();
-                </programlisting>
+                <para>To set and reset <varname>postgis.gdal_enabled_drivers</varname> for current session</para>
 
                 <programlisting>
-SET postgis.gdal_enabled_drivers TO 'GTiff PNG JPEG';
+SET postgis.gdal_enabled_drivers = 'ENABLE_ALL';
 SET postgis.gdal_enabled_drivers = default;
                 </programlisting>
 
-                <para>Enable all GDAL Drivers</para>
-                <programlisting>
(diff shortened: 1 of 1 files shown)


robe2:
I'm going to hold off on pushing this until after 3.5.0 release which I am in the middle of. I agree with most of the changes except possibly for one.  Looks like your change wiped out setting for just the current session.

doc/reference_guc.xml
```diff
@@ -151,18 +151,16 @@ SET postgis.gdal_datapath = 'C:/Program Files/PostgreSQL/9.3/gdal-data';</progra
...
-                <para>Sets default enabled drivers for all new connections to server. Requires super user access and PostgreSQL 9.4+.
+                <para>Sets default enabled drivers for all new connections to whole database cluster. Requires super user access and PostgreSQL 9.4+.
                     Also note that database, session, and user settings override this.</para>
-                <programlisting>ALTER SYSTEM SET postgis.gdal_enabled_drivers TO 'GTiff PNG JPEG';
-SELECT pg_reload_conf();
-                </programlisting>
-
                 <programlisting>
-SET postgis.gdal_enabled_drivers TO 'GTiff PNG JPEG';
-SET postgis.gdal_enabled_drivers = default;
```
robe2, line 165: These need to stay cause they represent change drivers just for the running session which is important.  e.g. you don't want to reconnect for changes to take effect and you are using GDAL raster functionality for intermediary - one time use to convert to postgis raster at which point you might not need these.

nikolai-b: Thank you, that makes sense. It was slightly confusing to me as it was below the `ALTER SYSTEM` but with no text / explanation in between. Do you want me to copy the format (session, db, system) used in `enable_outdb_rasters`?

robe2: Yes that would be great.

doc/reference_guc.xml
```diff
@@ -234,16 +228,17 @@ SET postgis.enable_outdb_rasters = True;
...
-                <para>Setting for whole database cluster. You need to reconnect to the database for changes to take effect.</para>
+                <para>Setting for whole database cluster. Requires super user access.
+                    Also note that database, session, and user settings override this.</para>
                 <programlisting>
  --writes to postgres.auto.conf
-ALTER SYSTEM postgis.enable_outdb_rasters = true;
+ALTER SYSTEM SET postgis.enable_outdb_rasters = true;
  --Reloads postgres conf
 SELECT pg_reload_conf();
                 </programlisting>
```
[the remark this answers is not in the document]
nikolai-b: I think the postgis docs would do well to reference https://www.postgresql.org/docs/current/sql-set.html and https://www.postgresql.org/docs/current/sql-altersystem.html as there isn't anything postgis specific about these.  There the postgres docs have `{ TO | = } { value [, ...] | DEFAULT }` which would be be helpful.

robe2:
Committed at d0e53a5 -- 56622d9 for PostGIS 3.6.0

and d0e53a5 for PostGIS 3.5.1

PR #780, opened 2024-10-10
Fix ST_NewEdgesSplit can cause invalid topology

bjornharrtell:
ref https://trac.osgeo.org/postgis/ticket/5795


strk (changes requested):
Could you please include the ticket reference as a label in all possible outputs of the test case ? It will be easier to debug in case of failures

bjornharrtell:
@strk something like I've done now?

I've also now expanded the test so it does more operations causing more invalidities until it crashes with "ERROR:  [/projects/postgis/liblwgeom/topo/lwgeom_topo.c:_lwt_AddFaceSplit:1918] Backend error: Corrupted topology: ring of edge -12 is topologically non-closed". Not sure how to test for that without env specific path.

bjornharrtell:
@strk I am hopeful I have found the issue and fix, AFAIK it seems the issue was that current code mixed up the assignment of new next right on connected edges.

strk:
Great you found the fix. We might want to backport this as I suspect all supported versions are affected.

bjornharrtell:
Merged elsewhere.

PR #783, opened 2024-10-24
The example of the ST_LineFromMultiPoint function contains an error.

w345731923:
Other branches have the same issue, along with the translation file.


w345731923:
@strk Why is my CI showing an error?I only submitted a sample modification.

PR #785, opened 2024-11-27
reference ST_ClipByBox2 in ST_Intersection manual

arenevier:
I have an app using postgis. I was using ST_Intersection to compute the intersection of geometries with a bounding box and it was super slow. What I realled needed was ST_ClipByBox2, but I was not aware of that method. I would have found it easier it had been referenced in the "See Also" of ST_ClipByBox2


strk:
Merged as d154235a363de2014c9eb6176a95820dcccd092b, thanks

PR #786, opened 2024-12-18
Update installation.xml.Fix the git clone address.

w345731923:


w345731923:
@strk hi,I fixed #5824.

w345731923:
@robe2 Got it, thanks!

PR #789, opened 2025-02-13
Fix typos

ProjectMutilation:
The variable "result" is used without overwriting the value. 
The variable "i" was set instead of the variable "result". 
Fixes: 61ca419a9 ("Implement ST_ModEdgeHeal and ST_NewEdgeHeal in C") 
Found by Postgres Pro


pramsey:
Thanks, applied in c30f3237b30c19278b534f654bf211144868df39

PR #793, opened 2025-04-04
Fix ST_AddPoint with empty point argument

pramsey:


liblwgeom/lwgeom_api.c
```diff
@@ -134,6 +135,7 @@ getPoint4d_p(const POINTARRAY *pa, uint32_t n, POINT4D *op)
 
 	if ( n>=pa->npoints )
 	{
+		lwerror("%s [%d] called with n=%d and npoints=%d", __FILE__, __LINE__, n, pa->npoints);
```
lnicola, line 138: Do we still need the `LWDEBUGF` below?

lnicola (approved):

liblwgeom/lwgeom_api.c
pramsey: No, we don't

regress/core/tickets.sql
```diff
@@ -1575,3 +1575,7 @@ WHERE ST_DFullyWithin(
     extent,
     1700
 );
+
+SELECT '#####', ST_AsText(ST_AddPoint(
```
lbartoletti, line 1579: We have the same problem with `ST_SetPoint`:

SELECT ST_AsText(ST_SetPoint('LINESTRING(1 1, 2 2, 3 3)', 1, 'POINT EMPTY'));
                           st_astext                            
----------------------------------------------------------------
 LINESTRING(1 1,3.3425537346456e-310 3.34255373464587e-310,3 3)
(1 ligne)

PR #803, opened 2025-05-14
SFCGAL: Add CG_Simplify

lbartoletti:
This introduces CG_Simplify, a new function for SFCGAL
backend based on sfcgal_geometry_simplify. 
The algorithm is based on [CGAL Simplification package](https://doc.cgal.org/latest/Polyline_simplification_2/index.html) and uses CDT.

The documentation and unit tests are updated accordingly.

Since our algorithm supports both Z and M coordinates, I have also enhanced the ptarray to SFCGAL conversion with proper M dimension support. This enhancement requires SFCGAL 1.5.0 or higher for M coordinate handling. This requirement is not an issue in practice, as the CG_Simplify function itself requires SFCGAL 2.1.0 or higher, which already satisfies the dependency.

SFCGAL 2.1 or above is needed.

cc @ptitjano


lbartoletti:
Merged

PR #808, opened 2025-05-26
sfcgal: Adapt CG_ExtrudeStraightSkeleton test to validity fix

ptitjano:
The geometry returned was invalid because the bottom/base of the roof was present. So, the geometry contained inner surface inside the building (wall + roof).

The unit test is updated by counting the number of patches instead of checking the result. This makes it easy to have a unit test which works for all SFCGAL versions.

See: https://gitlab.com/sfcgal/SFCGAL/-/merge_requests/511


lbartoletti (approved):

lbartoletti:
Merged. Thanks @ptitjano

PR #810, opened 2025-05-31
Doc: replace outdated references to trac.osgeo.org/gdal by current gdal.org ones

rouault:


robe2:
@rouault  I'll commit this shortly.  For future reference, no need to update the .po.  Weblate does that automatically when I push the latest .pot file

robe2:
Committed at f02bf29814

PR #811, opened 2025-06-02
Coverage Cleaning with GEOS 3.14

pramsey:
First draft of PostGIS exposure of new coverage cleaner.

API is a window function:
```
geometry ST_CoverageClean(
          geometry winset geom,
          float8  snappingDistance = -1,
          float8  gapMaximumWidth = 0,
          text  overlapMergeStrategy = 'MERGE_LONGEST_BORDER'
)
```

### DIFF
--- a/doc/reference_coverage.xml
+++ b/doc/reference_coverage.xml
@@ -109,6 +109,7 @@ SELECT true = ALL (
       <para>
         <xref linkend="ST_IsValid"/>,
         <xref linkend="ST_CoverageUnion"/>,
+        <xref linkend="ST_CoverageClean"/>,
         <xref linkend="ST_CoverageSimplify"/>
       </para>
     </refsection>
@@ -212,7 +213,9 @@ SELECT id, ST_AsText(ST_CoverageSimplify(geom, 30) OVER ())
     <refsection>
       <title>See Also</title>
       <para>
-        <xref linkend="ST_CoverageInvalidEdges"/>
+        <xref linkend="ST_CoverageInvalidEdges"/>,
+        <xref linkend="ST_CoverageUnion"/>,
+        <xref linkend="ST_CoverageClean"/>
       </para>
     </refsection>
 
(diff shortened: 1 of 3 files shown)


dr-jts:
Throwing out another idea for API: 

Provide only `gapWidth` as a separate (optional) parameter, and put other parameters into a single text string (as in [`ST_Buffer`](https://postgis.net/docs/manual-3.5/ST_Buffer.html)). Reasons:
* The most likely call is to specify `gapWidth` only.  snapping tolerance and merge strategy are likely to be fine with defaults for most uses
* That is extensible if parameters are added to support more functionality (for example, [`mapshaper -clean`](https://github.com/mbloch/mapshaper/wiki/Command-Reference#-clean) provides quite a few options, and my expectation is that there will be ideas for further cleaning techniques which emerge)

robe2:
> Throwing out another idea for API:
> 
> Provide only `gapWidth` as a separate (optional) parameter, and put other parameters into a single text string (as in [`ST_Buffer`](https://postgis.net/docs/manual-3.5/ST_Buffer.html)). Reasons:
> 
>     * The most likely call is to specify `gapWidth` only.  snapping tolerance and merge strategy are likely to be fine with defaults for most uses
> 
>     * That is extensible if parameters are added to support more functionality (for example, [`mapshaper -clean`](https://github.com/mbloch/mapshaper/wiki/Command-Reference#-clean) provides quite a few options, and my expectation is that there will be ideas for further cleaning techniques which emerge)

I like that idea.  Fewer arguments that are not commonly used the better.

dr-jts:
> Throwing out another idea for API:
> 
> Provide only `gapWidth` as a separate (optional) parameter, and put other parameters into a single text string (as in [`ST_Buffer`](https://postgis.net/docs/manual-3.5/ST_Buffer.html)). Reasons:

Okay, Paul points out that named parameters with defaults are just as flexible and more structured than a params string.  So I'm good with that.

PR #814, opened 2025-06-10
liblwgeom: avoid NULL pointer dereference

ProjectMutilation:


robe2:
committed at https://github.com/postgis/postgis/commit/c932604514651d454543b66fdb18055fd5ae40a2

PR #815, opened 2025-06-10
replace input arguments check for avoiding null dereference

ProjectMutilation:


ProjectMutilation:
Applied 0f4a868ef891e6211510e

PR #817, opened 2025-06-10
Added missing curly braces for the conditional block

ProjectMutilation:


strk:
@rouault could you please take a look at this ? It looks it was your 4f653e76d48e93a81b6aa2a9d399154503e3e340 that introduced the need for brackets in there, and I don't know if the va_copy is needed again after it was done on line 209

rouault:
> could you please take a look at this ?

the PR looks fine and needed.

> and I don't know if the va_copy is needed again after it was done on line 209

I'd say it is either need in both places, or not needed at all. I can't find a reference that explains that clarly, so out of caution I'd use va_copy in both places.

strk:
@pramsey @robe2 any reason why we don't use vasprintf from GNU instead ? We are already using it in loader/pgsql2shp-core.c - beside we are also checking for it in ./configure.ac but not using the result of that check at the moment...

strk:
@ProjectMutilation would you consider  adding a test that triggers the problem in liblwgeom/cunit/cu_stringbuffer.c ?

strk:
merged as 7a420853f4f810a66cd7361615df0411cfb03c11

PR #824, opened 2025-07-28
Improve clarity of docs for ST_Buffer's side parameter

petern48:
It was very unclear to me what the default of ST_Buffer's side parameter was and what case this line applied to.
```
By default end caps are square.
```

To my understanding, this line intended to mean that when 'left' or 'right' are specified (which aren't default options), the endcaps are square instead of the usual default of 'round'.


pramsey:
Thanks, applied in 56c2cc850

PR #825, opened 2025-08-23
.gitignore .deps/ .Po build artifacts

polobo:
Saw these artifacts show up as untracked after performing a simple make.  Doesn't seem like it is something only my setup would be generating so put it in the project .gitignore

https://trac.osgeo.org/postgis/ticket/5972#ticket


strk:
All merged, thanks

PR #826, opened 2025-08-24
Enable MobilityDB to call some PostGIS functions

estebanzimanyi:


Komzpa:
This PR is now carried forward in #947, which keeps the MobilityDB helper-export intent, includes Esteban's authorship via co-authorship, and has the additional export shims/tests needed for the current PostGIS tree.

I am leaving this open for maintainers/contributor preference rather than closing it directly; technically, #947 is the maintained successor and is currently green.

Komzpa:
Replaced with https://github.com/postgis/postgis/pull/947

PR #827, opened 2025-08-29
doc(sfcgal): Fix CG_ApproximateMedialAxis regarding 3d geometries

ptitjano:
CG_ApproximateMedialAxis ignores the Z dimension and always gives a 2D result.

cc @lbartoletti


pramsey:
Needs to be rebased to master.

PR #830, opened 2025-09-22
init nurbs

lbartoletti:


Komzpa (commented):
I'd love to see some support in ST_Intersects and ST_Distance, even if it will be similar to the curves where they will get linearized before that

lbartoletti:
@Komzpa Sorry, I’m closing this for now, it’s not ready yet, but I’m already taking your feedback into account. Thanks

PR #834, opened 2025-10-02
Fix NULL-pointer checks for gaztab and rultab in GetStdFromPortalCache

ProjectMutilation:
Bad copy paste was potential reason of NULL pointer dereference

Fixes: c6091a4bb ("Prep to move address_standardizer into extensions folder")


pramsey:
Thanks!

PR #854, opened 2026-04-06
liblwgeom: topo: remove dead code

ProjectMutilation:
The condition is always false.


pramsey:
Yikes? https://github.com/postgis/postgis/actions/runs/24030800655/job/70078956082#step:4:2308

pramsey:
Done at 88169d2c58bd48552512eeeeae5f6c8ecc71e22f

strk:
Excuse me for the late review, but if checkFace is false foundInFace can very well be -1, as initialized at the top of the function. It's not very readable code, I know, but did that line cause any trouble ?

pramsey:
It's already committed, feel free to put it back (with a comment). The removable looked harmless to me.

PR #855, opened 2026-04-06
liblwgeom: Add NULL geometry check in lwgeom_simplify_in_place

ProjectMutilation:
Prevent crash by checking for NULL geometry parameter and returning appropriate error message instead of dereferencing NULL pointer.


pramsey:
Done at 88169d2c58bd48552512eeeeae5f6c8ecc71e22f

PR #858, opened 2026-05-04
docs: split PostGIS skill by scope

Komzpa:
## Summary
- move the PostGIS skill from `doc/SKILL.md` to `skills/postgis/SKILL.md` so skill importers see a real `postgis` skill instead of a `doc` skill
- split the old mixed guidance into two skills: PostGIS-related spatial/data-modeling guidance and general SQL style/maintenance guidance
- keep the spatial-adjacent operational advice in the PostGIS skill, including BRIN and `INCLUDE` indexing, absolute stored values, missing coordinate handling, SQL query-shaping gotchas, version checks, raster advice, geometry/geography naming, projection advice, and KNN gotchas
- keep only generic SQL documentation, formatting, idempotent-file, migration, error-message, and include-file-path advice in `skills/sql-style/SKILL.md`

## Context
Downstream skill collections want to import the upstream skill as-is, but the old single file mixed PostGIS-related guidance with general documentation/style/debugging rules. Some importers also derive skill identity from the containing directory, so `doc/SKILL.md` could be imported as a `doc` skill rather than a `postgis` skill.

This split keeps advice that commonly matters while building PostGIS pipelines near the PostGIS skill, even when the advice is not a function-specific PostGIS gotcha. The non-spatial remainder is named `sql-style` because it is about documentation, formatting, and maintenance hygiene rather than SQL programming techniques.

## Validation
- `git diff --check HEAD~1..HEAD`
- checked both skill files have `name` and `description` frontmatter


robe2:
@Komzpa Any reason you haven't committed this yet?  PG19 beta will be coming out soon, so we'll want to release some sort of tagged release alongside that.

lnicola:
FWIW, I think harnesses look for those under `.agents/skills` so they won't find them. That is, assuming anyone wants to use them while working on PostGIS itself, which is questionable.

PR #859, opened 2026-05-06
sfcgal: Adapt CG_ExtrudeStraightSkeleton test to triangulation improvements

ptitjano:
The geometry returned by CG_ExtrudeStraightSkeleton, when extrude and roof parameters are set was triangulated. Since SFCGAL 2.3, the triangles can be merged when the underlying surface is coplanar. This gives a smoother result and a polyhedralsurface which has less patches.

Adapt the unit test to work with all SFCGAL versions by comparing the number of generated patches.

See: https://gitlab.com/sfcgal/SFCGAL/-/merge_requests/730

cc @lbartoletti


lbartoletti:
Committed, thanks

PR #862, opened 2026-05-25
sfcgal: Adapt to deprecated sfcgal_geometry_tesselate

ptitjano:
`sfcgal_geometry_tessellate` is now used instead.

cc @lbartoletti


pramsey (approved):
Good catch.

PR #863, opened 2026-06-01
Hex9 Extension Draft PR

MrBenGriffin:
- **Runtime dependency:** liblwgeom only (resolved at load via the standard
  PostGIS module — no extra libraries, no GEOS/PROJ calls).
- **Payload:** the equal-area warp needs a precomputed displacement field.
  «CURRENT STATE: the source carries a ~13 MB generated C header
  (`h9_warp_embedded.h`) that compiles to a ~2.1 MB `.o`. PLANNED: ship the
  ~2.14 MB binary warp file and `.incbin` it, so the in-tree source stays ~2 MB.»
  The warp field is reconstructed at load (~1 s, once per backend).
- **Init cost:** ~1 s per backend to build the interpolation state (down from
  ~13 s after fixing an O(n·grid) point-location step).

## Build & integration

- Builds today via a standalone PGXS `Makefile`.
- **TODO before merge:** convert to in-tree `Makefile.in` + wire into
  `extensions/Makefile` and `configure` so it builds with a normal PostGIS
  `make`; remove the hardcoded `POSTGIS_SRC` header path.

## Tests

- `make installcheck` regression test (`sql/hex9.sql` + `expected/hex9.out`)
  covering encode/decode round-trip, binning, cell/grid geometry.
- «Note any coverage you want to highlight or add.»

## Licensing
The extension is currently **Apache-2.0**; PostGIS is **GPL-2.0-or-later**.
Flagging this for guidance — happy to **relicense the extension to GPL-2.0+** to match the tree if that's preferred.

## Open items (draft → ready)

- [ ] In-tree build wiring (`Makefile.in`, `extensions/Makefile`, `configure`)
- [ ] Remove hardcoded `POSTGIS_SRC`; resolve liblwgeom headers from the tree
- [ ] PostGIS version-compat guard in `_PG_init`
- [ ] Switch warp payload to `.incbin` of the 2 MB binary
- [ ] `h9_version()` string (still reads `1.0.0`)
- [ ] Licensing decision
- [ ] C++17 / `-fno-exceptions` baseline confirmation with maintainers

## Build context
- Language: C++17 (`-fno-exceptions`), C ABI at the PG boundary.
- Tested on: PG 18 / PostGIS 3.6.2 / macOS arm64+x86_64


MrBenGriffin:
I was too eager.. will try again shortly.

robe2:
> I was too eager.. will try again shortly.

@MrBenGriffin  As mentioned in https://discourse.osgeo.org/t/requesting-comments-re-recent-research-on-hierarchical-hexagonal-grids/147709/17 we don't like having any C++ code in our extensions directly.  it just causes all sorts of problems.

Also I'm a little concerned about how huge this extension is, it really should be it's own extension similar to the https://github.com/postgis/h3-pg  

We've been trying to avoid adding more extensions into PostGIS code as it just causes hell for packagers.  
We've been hating it so much we even removed address_standardizer extension and postgis_tiger_geocoder extension out of the PostGIS core code base in PostGIS 3.7.  

Is it possible for you to have this as a separate extension.  It really is very much like https://github.com/postgis/h3-pg in flavor.

Komzpa:
@MrBenGriffin it will also help to have a link to the spec of the standard and/or reference library.

This is definitely interesting as in "hex grid that continues deeper than h3" - is the thing an ongoing research or finalized? I'd be happy to provide some input based on our usage of h3  if there is a proper place for it (email works for me too, me@komzpa.net)

PR #883, opened 2026-06-16
Make ST_Relate DE-9IM inversion length-aware to prevent OOB access

Komzpa:
### Motivation
- A new `imInvert()` helper unconditionally indexed fixed DE-9IM positions, which can read/write past the end of short attacker-supplied pattern strings and cause out-of-bounds access in the GEOS prepared `ST_Relate` path. 
- The SQL input was only truncated to a maximum of 9 characters with `text_left(..., 9)` and no exact-length validation was performed, so short patterns could reach the inversion logic.

### Description
- Changed `imInvert(char *im)` to `imInvert(char *im, size_t len)` and made the three swaps conditional on `len` so only existing positions are accessed.  
- Cached the pattern length with `size_t im_len = strlen(im);` and passed it to `imInvert(im, im_len)`.  
- Normalized the pattern using `toupper((unsigned char)im[i])` to avoid undefined behavior for signed `char` values.  
- Modifications are confined to `postgis/lwgeom_geos_predicates.c` around the `relate_pattern` implementation and the inversion helper.

### Testing
- Ran `git diff --check` which reported no whitespace or diff-check errors.  
- Ran `git clang-format` / `git clang-format --diff --cached` and `clang-format` did not report further changes.  
- Attempted the full build bootstrap with `./autogen.sh` but it failed in this environment due to missing `aclocal`, so a full `./configure`/`make`/`make check` run could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2af52c323c8324a56044d42262cc31)

### DIFF
--- a/postgis/lwgeom_geos_predicates.c
+++ b/postgis/lwgeom_geos_predicates.c
@@ -888,12 +888,27 @@ Datum coveredby(PG_FUNCTION_ARGS)
  *  to the GEOSPreparedRelatePattern function.
  */
 static void
-imInvert(char *im)
+imInvert(char *im, size_t len)
 {
 	char t;
-	t = im[1]; im[1] = im[3]; im[3] = t;
-	t = im[2]; im[2] = im[6]; im[6] = t;
-	t = im[5]; im[5] = im[7]; im[7] = t;
+	if (len > 3)
+	{
+		t = im[1];
+		im[1] = im[3];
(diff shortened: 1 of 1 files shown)


Komzpa:
Recreated from the Komzpa fork as #889; closing this mirror-head duplicate.

PR #884, opened 2026-06-16
Prevent OOB reads in GIDX BRIN merge by checking dimensionality

Komzpa:
### Motivation
- The new BRIN merge path for GIDX-backed summaries could read past a shorter GIDX when merging mixed-dimension summaries, causing out-of-bounds reads or corrupted BRIN summaries. 
- The existing `add_value` path already treated differing GIDX dimensions as unmergeable, but the merge helper lacked an equivalent check. 
- The change aims to make BRIN merge behavior defensively consistent with `add_value` and avoid native memory-safety issues reachable from BRIN operations.

### Description
- Add a `gidx_brin_inclusion_mergeable(GIDX *gidx_key, GIDX *gidx_geom)` helper that returns whether the two GIDX objects have the same number of dimensions. 
- Guard `gidx_brin_inclusion_merge()` with the new helper to return early when dimensionality differs so the merge loop never indexes non-existent dimensions. 
- Expose boolean wrappers for the helper as BRIN support functions and register them as support proc 12 in the GIDX-backed opclass SQL for geography, 3D geometry, and 4D geometry. 
- Changes touch `postgis/brin_nd.c`, `postgis/postgis_brin.sql.in`, and `postgis/geography_brin.sql.in` and otherwise preserve existing behavior for matching-dimension merges.

### Testing
- Ran `git diff --check` which reported no whitespace or patch issues and succeeded. 
- Attempted a full build with `make -j"$(nproc)"` but the tree requires a prior `./configure` run so the build did not proceed. 
- Attempted `./autogen.sh && ./configure --without-raster` but the environment lacked `aclocal`/automake and installing build tools failed due to network/proxy restrictions, so compilation and runtime regression tests could not be executed in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2b463d53408324acd16489e33a932d)

### DIFF
--- a/postgis/brin_nd.c
+++ b/postgis/brin_nd.c
@@ -17,6 +17,8 @@ static Datum gidx_brin_inclusion_add_value(
 
 static GIDX * gidx_brin_inclusion_merge(
 	GIDX *gidx_key, GIDX *gidx_geom);
+static bool gidx_brin_inclusion_mergeable(
+	GIDX *gidx_key, GIDX *gidx_geom);
 
 /*
  * As for the GiST case, geographies are converted into GIDX before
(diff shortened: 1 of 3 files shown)


Komzpa:
Recreated from the Komzpa fork as #890; closing this mirror-head duplicate.

PR #900, opened 2026-06-16
sfcgal: fix regression gating and roof checks

Komzpa:
Covers related hardening follow-ups:

- row 18: polygon repair regression tests are only selected when CGAL 6.0+ is available, matching the C guard
- row 19: generic CG_GenerateRoof is regression-checked against specialized roof wrappers with distinct height/slope values

Verification:
- git diff --check
- ./configure --without-raster --without-topology --with-sfcgal --without-protobuf --without-json
- perl ./utils/repo_revision.pl
- make -C liblwgeom -j32
- make -C libpgcommon -j32
- make -C sfcgal -j32
- ./config.status --file=sfcgal/regress/tests.mk:sfcgal/regress/tests.mk.in and inspected generated gates


lbartoletti (approved):
thanks!

lbartoletti:
unrelated macos failure, already fixed in antoher PR

PR #927, opened 2026-06-17
Add foreign table coverage for Populate_Geometry_Columns

Komzpa:
## Summary

- add regression coverage for `Populate_Geometry_Columns` on foreign tables
- cover typmod metadata and constraint-style `use_typmod=false` using a `file_fdw` table so the populate call visibly changes foreign-table metadata without depending on loopback PostgreSQL authentication
- gate the superuser-only FDW DDL so hardened non-superuser regression runs stay deterministic
- use a per-run server-side CSV fixture path for repeatable local reruns
- add a NEWS entry for Trac ticket 3340
- rebase the test-only branch onto current `upstream/master` (`4e470f153`)

## Validation

- `./autogen.sh`
- `./configure --with-jsondir=/usr --with-projdir=/usr --with-raster --with-topology --with-sfcgal`
- `make -C postgis -j32`
- `make -C extensions postgis_extension_helper.sql`
- `make -C extensions/postgis -j1`
- `sudo -n make -C postgis install`
- `sudo -n make -C extensions/postgis install`
- `make -C regress check RUNTESTFLAGS='--extension --verbose' TESTS="$(pwd)/regress/core/regress_management"`
  - includes the harness's automatic upgrade rerun
- `./utils/check_news.sh .`
- `git diff --check`
- `git diff --cached --check`
- `git clang-format --diff upstream/master -- regress/core/regress_management.sql regress/core/regress_management_expected NEWS`
  - reported no modified files to format

Closes https://trac.osgeo.org/postgis/ticket/3340

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -73,6 +73,8 @@ To take advantage of all postgis_sfcgal extension features SFCGAL 2.3+ is needed
           and pass callback call data as actual arguments (Darafei Praliaskouski)
  - #2898, Document SQL function cost tiers for contributors
           (Darafei Praliaskouski)
+ - #3340, Add regression coverage for Populate_Geometry_Columns on
+          foreign tables (Darafei Praliaskouski)
  - #6062, [topology] Stop using recursive snapping, for improved robustness (Sandro Santilli)
  - #6065, Improved winding order computation robustness for rings having collapsed elements
           (Sandro Santilli)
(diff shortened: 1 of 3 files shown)


robe2 (changes requested):
This doesn't need to be mentioned in the news.  It will just confuse people. Regression tests should never be mentioned in the news.

PR #928, opened 2026-06-17
doc: fix Japanese PDF listing arrows

Komzpa:
## Summary
- replace Unicode arrows in the CG_GenerateSkillionRoof SQL program listing with ASCII arrows
- avoid XeTeX/listings treating those arrows as undefined control sequences in the Japanese PDF build
- add a NEWS entry for Trac #6082

## Testing
- `make -j32`
- reproduced failure before the patch with `make -C doc/po/ja local-pdf V=1`: `postgis-out.tex:30066: Undefined control sequence`
- `make -C doc/po/ja local-pdf V=1`
- `git diff --check`

Closes #6082


Komzpa:
Landed upstream as https://github.com/postgis/postgis/commit/746769bc9603c34262f2a704369a5efda9ac4ebe, which carries the same #6082 fix and closes this PR in the commit message. Closing the leftover mirror PR.

PR #944, opened 2026-06-17
Fix tiger shell loader schema creation (#6072)

Komzpa:
## Summary
- create the tiger_data schema from generated Unix shell tiger loader scripts
- use `CREATE SCHEMA IF NOT EXISTS` directly so shell loaders do not expand PL/pgSQL dollar quotes
- add a tiger geocoder upgrade regression that checks the sh loader script contains the schema creation command
- keep the upgrade regression portable across supported PostgreSQL versions by using `position(...)` instead of PostgreSQL 15+ `regexp_count`
- add a 3.6.5 NEWS entry

## Tests
- `git diff --check`
- `make -j32`
- `sudo -n make install`
- `make -C extensions/postgis_tiger_geocoder -j1`
- `make -C extensions/postgis_tiger_geocoder installcheck`

Closes #6072 for PostGIS 3.6

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,8 @@ PostGIS 3.6.5
 
 * Fixes *
 
+- #6072, Create tiger_data schema from shell tiger loader scripts
+         (Darafei Praliaskouski)
 - #5989, CurvePolygon distance corner case (Paul Ramsey)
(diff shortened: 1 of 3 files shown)


Komzpa:
Moved to the split tiger geocoder repository: https://gitea.osgeo.org/postgis/postgis_tiger_geocoder/pulls/25

PR #956, opened 2026-06-18
Clarify ST_LineSubstring 2D length fractions

Komzpa:
References [Trac #5776](https://trac.osgeo.org/postgis/ticket/5776).

This clarifies that `ST_LineSubstring(geometry, ...)` uses fractions of 2D line length even for Z-enabled inputs, while Z values are interpolated at the selected 2D locations. The added example uses the ticket's 3D linestring shape to show why half of the 2D length is not necessarily half of `ST_3DLength`.

Validation run locally:
- `make -C doc check-xml`
- `git diff --check`

### DIFF
--- a/doc/reference_lrs.xml
+++ b/doc/reference_lrs.xml
@@ -381,8 +381,8 @@ FROM (SELECT ST_GeomFromText('LINESTRING(1 2, 4 5, 6 7)') As the_line) As foo;
 			starting and ending at the given fractional locations.
             The first argument must be a LINESTRING.
 			The second and third arguments are values in the range [0, 1]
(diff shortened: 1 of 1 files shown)


Komzpa:
Updated the wording for the geography-overload review thread.

On head `7c1a7aab06496d296b1294349354fb3f75c40285`, the generic fraction sentence now only says the arguments are start/end locations, and the planar-length statement is explicitly qualified to geometry inputs. That avoids implying that the geography overload uses 2D planar length.

Validation:

- `make -C doc check-xml`
- `git diff --check && git diff --cached --check`

PR #965, opened 2026-06-18
Preserve raster_columns ACLs during upgrade

Komzpa:
Closes [Trac #2823](https://trac.osgeo.org/postgis/ticket/2823).

## Summary

- preserve existing `raster_columns` privileges across raster upgrade scripts even though the view is dropped and recreated
- save relation-level and column-level ACL payloads before upgrade cleanup drops the old view
- replay saved relation and column grants under their original grantors after recreating `raster_columns`
- let PostgreSQL maintain `pg_shdepend` ACL dependencies through normal `GRANT` execution instead of directly updating catalog ACL arrays
- keep the generated public grant as the install default, but replace it with the saved ACL when an upgrade captured one
- add raster upgrade hook coverage for revoked `PUBLIC`, delegated table grantor, column-level ACL preservation, and restored ACL dependency tracking

## Maintenance

- rebased onto `a4ecc46176c3fc0ab3fa69cf117d408bc622e592`
- no Gitea mirror PR was found for this branch
- the previous broad CI failures were stale loader expectation failures (`LongOptions` / `TestANALYZE`) on the old base; the branch-specific `check_raster_columns` regression was passing there and passes after rebase

## Validation

- `make -C postgis -j32`
- `make -C raster/rt_pg -j32`
- `make -C extensions postgis_extension_helper.sql`
- `make -C extensions/postgis -j1`
- `make -C extensions/postgis_raster -j1`
- `sudo -n make -C postgis install`
- `sudo -n make -C raster/rt_pg install`
- `sudo -n make -C extensions/postgis install`
- `sudo -n make -C extensions/postgis_raster install`
- `make -C raster/test/regress check RUNTESTFLAGS='--upgrade --extension --verbose' TESTS="$(pwd)/raster/test/regress/check_raster_columns"`
- `make -C raster/test/regress check RUNTESTFLAGS="--upgrade --extension --verbose --after-create-db-script $(pwd)/regress/hooks/standard-conforming-strings-off.sql" TESTS="$(pwd)/raster/test/regress/check_raster_columns"`
- `./utils/check_news.sh .`

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -115,6 +115,8 @@ To take advantage of all postgis_sfcgal extension features SFCGAL 2.3+ is needed
  - GH-899, [raster] Honor PostgreSQL interrupts in long-running GDAL
           progress callbacks (Darafei Praliaskouski)
  - #2386, Add missing raster2pgsql man page (Darafei Praliaskouski)
+ - #2823, [raster] Preserve raster_columns privileges across upgrades
+          (Darafei Praliaskouski)
  - #5975, Initialize skipped union-find cluster ids deterministically
           (Darafei Praliaskouski)
  - #5904, [topology] Avoid MinGW format warnings when logging topology ids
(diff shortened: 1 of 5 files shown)


robe2:
I'm closing this one out as a won't do.  If we do it, it would have to be done at the level of our postgis upgrade generator scripts.  I also don't think we ever change the column defs on these anymore which would be the case where we would need to drop these tables.

PR #977, opened 2026-06-18
doc: correct ST_BandIsNoData forceChecking default

Komzpa:
## Summary

Corrects the `ST_BandIsNoData` synopsis to show `forceChecking=false`, matching the actual SQL default and the existing description text.

This helps users looking at https://trac.osgeo.org/postgis/ticket/5176 find the cheap metadata path instead of seeing the expensive forced pixel scan advertised as the default. It is intentionally documentation-only; changing raster no-data scan behavior would be a larger semantic change.

References https://trac.osgeo.org/postgis/ticket/5176

## Validation

- Rebased on `postgis:master` at `63cbf4f069afebe50cb520b7abe6440c898842e9`.
- `git diff --check`
- `xmllint --noout doc/reference_raster.xml`
- `make -C doc comments`
- `make -C doc check-xml`

SQL regression tests were not run because this only changes manual text.

### DIFF
--- a/doc/reference_raster.xml
+++ b/doc/reference_raster.xml
@@ -3976,12 +3976,12 @@ WHERE rid = 2;
                     <funcdef>boolean <function>ST_BandIsNoData</function></funcdef>
                     <paramdef><type>raster </type> <parameter>rast</parameter></paramdef>
                     <paramdef><type>integer </type> <parameter>band</parameter></paramdef>
-                    <paramdef choice="opt"><type>boolean </type> <parameter>forceChecking=true</parameter></paramdef>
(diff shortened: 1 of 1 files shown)


robe2 (approved):
Confirmed the docs are wrong and this is implemented as false

Komzpa:
Closing this because the documentation fix has already landed on canonical Gitea master:

https://gitea.osgeo.org/postgis/postgis/commit/a4ecc46176c3fc0ab3fa69cf117d408bc622e592

GitHub kept this pull request open because the accepted commit is a cherry-pick of the same patch rather than the original pull request head SHA.

PR #990, opened 2026-06-20
loader: create shp2pgsql geometry columns directly

Komzpa:
## Summary

This closes https://trac.osgeo.org/postgis/ticket/4594 on current `master`.

The direct `shp2pgsql` geometry-column DDL path is already present on the current base through the newer loader action work. After rebasing, this branch keeps the release note and adds loader SQL expected fixtures that lock the default point outputs to inline `geometry(...)` typmod columns instead of `AddGeometryColumn` output.

## Validation

- `make -C loader -j32`
- `make -C regress/loader check RUNTESTFLAGS="--verbose --extension"`
- `make check-news`
- `./utils/check_news.sh .`
- `git diff --cached --check`

## Rebase refresh

- 2026-06-21: rebased over upstream/master `c4bae99fc25aabfc512dd0c14f1474b66b3e7730` and force-pushed head `1916ea8e93c6d80ecce56966687d18c8972efefb`.


robe2 (commented):
The actual implementation of this was done and committed at 0482d073a2d174a544428664fa7cebcd36b7d96a

These are just tests which I feel don't add enough value to warrant bothering with

PR #1017, opened 2026-06-20
doc: clarify repeated ring point validity

Komzpa:
## Summary
- clarify that consecutive repeated points in polygon boundary rings are treated as redundant vertices and do not make polygons invalid
- distinguish that from repeated points that create self-intersection, self-touching, or collapse

Trac: https://trac.osgeo.org/postgis/ticket/3425

## Testing
- `psql -X -v ON_ERROR_STOP=1 -d postgres -Atc "CREATE EXTENSION IF NOT EXISTS postgis; SELECT 'consecutive_polygon', ST_IsValid('POLYGON((0 0, 1 0, 1 0, 1 1, 0 0))'::geometry), ST_IsValidReason('POLYGON((0 0, 1 0, 1 0, 1 1, 0 0))'::geometry); SELECT 'repeated_nonconsecutive_polygon', ST_IsValid('POLYGON((0 0, 1 0, 0 0, 1 1, 0 0))'::geometry), ST_IsValidReason('POLYGON((0 0, 1 0, 0 0, 1 1, 0 0))'::geometry);"`
- `make -C doc check`
- `git diff --check`

### DIFF
--- a/doc/using_postgis_dataman.xml
+++ b/doc/using_postgis_dataman.xml
@@ -1090,6 +1090,13 @@ SELECT
         </para></listitem>
         </orderedlist>
 
+      <para>
+        Consecutive repeated points in a polygon boundary ring are treated as
+        redundant vertices, and do not make the polygon invalid. Repeated points
+        that cause the boundary to self-intersect, self-touch, or collapse
+        still violate the validity rules.
+      </para>
+
 	  <informaltable frame="none">
 		<tgroup cols="1">
 		  <tbody>


doc/using_postgis_dataman.xml
```diff
@@ -1090,6 +1090,13 @@ SELECT
         </para></listitem>
         </orderedlist>
 
+      <para>
+        Consecutive repeated points in a polygon boundary ring are treated as
+        redundant vertices, and do not make the polygon invalid. Repeated points
```
dbaston, line 1095: 1) I would remove the second sentence here, which makes no sense to me. How could a repeated point cause a boundary to self-intersect?

2) My intent with the ticket was to document _why_ the behavior is what it is / what interpretation of the SFS leads to teh conclusion that repeated points are valid.

PR #1028, opened 2026-06-20
Add pg_upgrade coverage to Woodie CI

Komzpa:
## Summary
- add a Woodpecker regression step that runs `utils/check_cluster_upgrade.sh` from PostgreSQL 14 to PostgreSQL 18
- seed the upgrade with a geography table matching the case from https://trac.osgeo.org/postgis/ticket/5899
- make the upgrade helper handle newer `initdb --no-data-checksums` targets and stop the upgraded cluster during cleanup
- run the cluster-upgrade script under the unprivileged `postgres` account in CI, while leaving `make install` privileged
- reject accidental root execution in `utils/check_cluster_upgrade.sh` before reaching `initdb`

Trac ticket: https://trac.osgeo.org/postgis/ticket/5995

## Maintenance
- rebased on current `master` (`c94f62361`)
- resolved the `NEWS` conflict
- addressed the live Codex review thread about running `initdb` as root

## Validation
- `utils/check_cluster_upgrade.sh -i regress/hooks/cluster-upgrade-geography.sql /usr/lib/postgresql/14/bin/pg_config /usr/lib/postgresql/18/bin/pg_config`
- `sudo -n utils/check_cluster_upgrade.sh /usr/lib/postgresql/14/bin/pg_config /usr/lib/postgresql/18/bin/pg_config` rejects root with the new guard
- `sudo -n -u postgres id -u` returned `125`, and `sudo -n -u postgres sh -c ...` confirmed the workspace command context is non-root
- `sh -n utils/check_cluster_upgrade.sh`
- `./utils/check_news.sh .`
- `git diff --check`

### DIFF
--- a/.woodpecker/regress.yml
+++ b/.woodpecker/regress.yml
@@ -131,3 +131,11 @@ steps:
       - <<: *steps-start-postgresql
       - <<: *steps-pg-install
       - <<: *steps-pg-test-all-upgrades
+
+  cluster-upgradecheck-pg14-to-pg18:
+    image: *test-image
+    depends_on: [ installcheck-pg14, installcheck-pg18 ]
+    commands:
+      - make -C build/pg14 install
+      - make -C build/pg18 install
+      - su postgres -c 'utils/check_cluster_upgrade.sh -i regress/hooks/cluster-upgrade-geography.sql /usr/lib/postgresql/14/bin/pg_config /usr/lib/postgresql/18/bin/pg_config'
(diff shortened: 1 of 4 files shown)


robe2:
@Komzpa  We never add news about our bots.  No user cares about them.  So I would take the NEWS out and then maybe push it as a branch to gitea postgis so woodie can test it.  If it passes I think it's good to go and can go anytime since it only affects our internal infra

robe2 (changes requested):
Get rid of the NEWS for this one cause no one cares about our internal infra except us.

robe2 (approved):
Already commited d9cdac922f99e86154f7ab04392280c61c87c912

PR #1065, opened 2026-06-20
Infer geometry_columns constraints for direct views

Komzpa:
## Summary

Trac: https://trac.osgeo.org/postgis/ticket/1705

This teaches `geometry_columns` to inherit constraint-derived geometry metadata for direct view and materialized-view columns. It uses PostgreSQL rewrite-rule origin metadata for top-level bare `VAR` target entries, so expression columns such as `ST_Buffer(geom)` still require an explicit typmod cast.

The follow-up narrows rewrite-rule scanning to the top-level serialized target list so bare `VAR` entries inside nested subqueries cannot be mistaken for the outer view column. The regression now covers the nested-subquery false-positive shape from review.

The regression covers:

- base constraint metadata
- direct pass-through views
- aliased direct view columns
- materialized views
- transformed geometry expressions staying `GEOMETRY`/SRID 0
- transformed geometry expressions over nested subqueries staying `GEOMETRY`/SRID 0

Docs describe the direct-view behavior consistently and keep the explicit-cast guidance for transformed geometry expressions.

## Testing

- /usr/bin/perl ./utils/repo_revision.pl
- make postgis_revision.h
- make -C postgis -j32
- make -C extensions postgis_extension_helper.sql
- make -C extensions/postgis -j1
- sudo -n make -C postgis install
- sudo -n make -C extensions/postgis install
- make -C regress check RUNTESTFLAGS="--extension --verbose" TESTS="$(pwd)/regress/core/tickets"
- make -C regress check RUNTESTFLAGS="--upgrade --extension --verbose" TESTS="$(pwd)/regress/core/tickets"
- make -C doc check-xml
- ./utils/check_news.sh .
- git diff --check upstream/master..HEAD && git diff --check && git diff --cached --check
- git clang-format --diff upstream/master -- postgis/postgis.sql.in regress/core/tickets.sql regress/core/tickets_expected doc/using_postgis_dataman.xml NEWS
- find . \! -user $(id -u) -print | head -20

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -91,6 +91,8 @@ To take advantage of all postgis_sfcgal extension features SFCGAL 2.3+ is needed
           (Darafei Praliaskouski)
  - #3103, Add regression coverage for exact-schema find_srid and
           geometry_columns lookups (Darafei Praliaskouski)
+ - #1705, Infer constraint metadata for direct view and materialized view
+          geometry columns (Darafei Praliaskouski)
  - #6038, Avoid stale relation lookups in geometry_columns after topology
           objects are dropped (Darafei Praliaskouski)
  - #5655, [doc] Skip XML validation during `make check` when xsltproc is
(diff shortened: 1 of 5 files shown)


robe2:
I feel like this one isn't worth it.  We really want to deprecate usage of constraints cause it's slow and has caused annoying problems in the past. @pramsey  what's your feeling about this?

robe2 (commented):
I think we should just get rid of this one and close as a won't fix.  I don't think doing anything with constraint based versions if worth the effort given the extra code

robe2:
Not worth fixing

PR #1084, opened 2026-06-21
Fix raster garden SQL placeholder generation

Komzpa:
## Summary
- teach the raster garden XSL generator placeholders for documented raster-specific argument types that previously rendered as empty SQL arguments
- add callback helper functions for deprecated and current raster map algebra callback signatures
- SQL-escape generated parameter labels so defaults such as `resample='nearest'` do not break the garden script
- keep `ST_CreateOverview` garden calls on a constrained overview fixture and make `nbands` placeholder selection explicit for text vs integer-array overloads
- use separate constrained overview fixtures for schema-qualified and unqualified `AddOverviewConstraints` / `DropOverviewConstraints` overloads so the generated garden does not double-add or double-drop the same overview constraint
- tag the generated `ST_Retile` extent with the fixture SRID after `UpdateRasterSRID` mutates the shared raster table

Trac ticket: https://trac.osgeo.org/postgis/ticket/2299

## Testing
- `make -C doc raster_gardentest_37.sql`
- generated SQL spot-checks for `ST_CreateOverview`, `AddOverviewConstraints`, `DropOverviewConstraints`, `ST_Band` text/list overloads, and `ST_AsJPEG`/`ST_AsPNG`/`ST_AsTIFF` integer-array `nbands` overloads
- generated SQL spot-check that `ST_Retile` now uses `ST_MakeEnvelope(0, 0, 1, 1, 3395)` after the fixture SRID update
- `rg -n "UpdateRasterSRID\\(,|,\\s*,|,\\s*\\)|\\(\\s*," doc/raster_gardentest_37.sql | head -200` (only label/data false positives after the fix)
- `make -C doc check-xml garden`
- `make -C regress staged-install-core staged-install-raster`
- loaded staged `postgis.sql`, `spatial_ref_sys.sql`, and `rtpostgis.sql` into a throwaway database and ran `psql -X -d postgis_2299_raster_garden < doc/raster_gardentest_37.sql`; semantic garden errors remain expected, but no generated SQL syntax failures from blank arguments or unescaped parameter labels were present
- direct `ST_CreateOverview('pgis_rgarden_overview_src'::regclass, 'rast'::name, 2, 'Lanczos')` smoke in a throwaway database returned an overview table instead of the previous missing-constraints exception
- direct Add/Drop overview smoke in a throwaway database returned true for both schema-qualified and unqualified `AddOverviewConstraints` / `DropOverviewConstraints` overloads using separate generated fixture tables
- `./utils/check_news.sh .`
- `git diff --check upstream/master..HEAD && git diff --check`

### DIFF
--- a/doc/Makefile.in
+++ b/doc/Makefile.in
@@ -761,8 +761,13 @@ else
 garden: postgis_gardentest_${POSTGIS_MAJOR_VERSION}${POSTGIS_MINOR_VERSION}.sql raster_gardentest_${POSTGIS_MAJOR_VERSION}${POSTGIS_MINOR_VERSION}.sql sfcgal_gardentest_${POSTGIS_MAJOR_VERSION}${POSTGIS_MINOR_VERSION}.sql topology_gardentest_${POSTGIS_MAJOR_VERSION}${POSTGIS_MINOR_VERSION}.sql
 endif
 
+RASTER_GARDEN_OUTDBFILE=/tmp/postgis_raster_garden_outdb_${POSTGIS_MAJOR_VERSION}${POSTGIS_MINOR_VERSION}.tif
+
 %_gardentest_${POSTGIS_MAJOR_VERSION}${POSTGIS_MINOR_VERSION}.sql: xsl/%_gardentest.sql.xsl postgis-out.xml
-	$(XSLTPROC) -o $@ $(XSLTPROC_PATH_OPT) $^
+	$(INSTALL_DATA) "$(abs_top_srcdir)/raster/test/regress/loader/SourceOverview.tif" "$(RASTER_GARDEN_OUTDBFILE)"
+	$(XSLTPROC) -o $@ $(XSLTPROC_PATH_OPT) \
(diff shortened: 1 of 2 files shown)


robe2 (changes requested):
The ST_CreateOverview test takes 32 minutes on my computer to run, that's the only egregious one I see.  The rest are at most 9 secs

```
SELECT  ST_CreateOverview('pgis_rgarden_overview_src'::regclass, 'rast'::name, 2, 'Lanczos');
```

Komzpa:
Addressed the slow `ST_CreateOverview` garden case on current head `01586aabdd0db8b009bc015930083f77170036e6`:

- its source fixture is limited to one raster row;
- that case uses `NearestNeighbour` instead of `Lanczos`;
- the general algorithm placeholder remains unchanged for the other garden cases.

The [current checks are green](https://github.com/postgis/postgis/pull/1084/checks). This is ready for another review.

robe2 (approved):

PR #925, opened 2026-06-17
doc: move development workflow documentation into repo

Komzpa:
Closes https://trac.osgeo.org/postgis/ticket/5638

## Maintenance Notes

- rebased on `postgis:master` at `4e470f1534795ae4a4c52ad5ad35b8744af9d708`
- current head: `bfd16cefc25f9cae395940896b11d1fcddd3feca`
- all current GitHub review threads are resolved (`reviewThreads`: 0 unresolved current threads)
- `doc/development/SKILL.md` is now a regular file, byte-identical to `doc/development/README.md`, so the release payload does not depend on symlink mtimes
- the temporary `doc/development/doc-wip.md` migration ledger has been removed; surviving high-level ideas are now in `doc/development/todo.md`
- raw Trac wiki source imports have been removed from the maintained tree; `doc/trac-wiki/` now retains only provenance/manifest files

## Summary

Moves maintained PostGIS developer and maintainer workflow documentation into the repository under `doc/development/`, organized as named Markdown topic pages with companion subdirectories where needed.

Current structure highlights:

- `contributing.md` for contribution routing, Trac tickets, mirrors, mailing lists, chat, bug reporting, security reporting, and contributor workflow
- `environment/ubuntu.md`, `environment/macos.md`, `environment/freebsd.md`, `environment/windows.md`, and `environment/docker.md` for local package-based, CI-derived, Windows MSYS2/Winnie, and container-based development environments
- `testing.md`, `testing/ci.md`, `tools.md`, and `style.md` for regression testing, garden checks, CI ownership, coverage/profiling, formatting, and source style
- `internals.md` and `internals/*` for memory management, PostgreSQL internals, empty geometry semantics, spatial collections, raster georeferencing, raster/GDAL boundaries, distance internals, and topology internals
- `release.md`, `release/*`, and `release-process.md` for API compatibility, versioning, deprecation/removal, SQL upgrade scripting, dependency/support guards, release preparation, publishing, announcements, and next-cycle setup
- `manual.md` and `website.md` for DocBook/manual work, generated comments, translations, images, Doxygen, the `postgis.net` Hugo repository, release-pointer updates, and website validation
- `governance.md`, `governance/project-inventory.md`, and `maintenance.md` for PSC process, RFC-derived contributor rules, project inventory, source-of-truth branches, mirror PR handling, tracker trailers, `NEWS`, and CI/review readback
- `todo.md` for high-level future research ideas that survived the Trac/wiki migration but are not accepted designs, release blockers, or scheduled work
[...]

PR #1104, opened 2026-06-21
topology: let CopyTopology skip layers

Komzpa:
Trac ticket: https://trac.osgeo.org/postgis/ticket/1195

Summary:
- add an optional `copy_layers` boolean argument to `topology.CopyTopology`, defaulting to true
- when false, copy primitives only and leave `topology.layer`, relation rows, and `topogeo_*` sequences out of the target topology

Validation:
- `make -j32`
- `make -C topology -j32`
- `make -C extensions/postgis_topology -j32`
- `sudo -n make -C topology install`
- `sudo -n make -C extensions/postgis_topology install`
- `make -C topology/test check RUNTESTFLAGS="--extension --verbose" TESTS="$(pwd)/topology/test/regress/copytopology"`
- `make -C doc check-xml`
- `./utils/check_news.sh .`
- `git diff --check`

---

Gitea mirror: https://gitea.osgeo.org/postgis/postgis/pulls/303

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -67,6 +67,8 @@ To take advantage of all postgis_sfcgal extension features SFCGAL 2.3+ is needed
           public liblwgeom header (Darafei Praliaskouski)
  - #4678, [raster] Move st_sum4ma neighborhood callback to C
           (Darafei Praliaskouski)
+ - #1195, [topology] Add CopyTopology option to skip layer copies
+          (Darafei Praliaskouski)
  - #2116, [raster] Add ST_Value nearest-neighbor boundary options
           (Darafei Praliaskouski)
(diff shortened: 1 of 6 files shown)


Komzpa:
While burning down the `TopologyCopy` leftover from the developer-doc migration in https://github.com/postgis/postgis/pull/925, I checked this against the current open topology PRs.

This PR covers the actionable `CopyTopology` part I wanted to preserve from the old note: primitive-only topology copying, with layers and `TopoGeometry` definitions left out when `copy_layers` is false.

The broader "detached layer attach/detach/drop lifecycle" idea is not covered here, but it is also not a blocker for the documentation migration. If that idea becomes real work later, it should get its own design/ticket rather than stay as a vague `doc-wip.md` item.

PR #1116, opened 2026-06-21
Rasterize curved geometry inputs

Komzpa:
## Summary

This is a bounded slice of https://trac.osgeo.org/postgis/ticket/1168.

`ST_AsRaster` now supports curved geometry inputs while using the curve handling available in the linked GDAL version:

- with GDAL versions before 3.14, PostGIS linearizes curved inputs before serializing them to WKB for rasterization
- with GDAL 3.14 and newer, standard OGR curve types use the native curve linearization added to `GDALRasterizeGeometries()` by https://github.com/OSGeo/gdal/pull/14509
- `NURBSCurve` inputs continue to use the PostGIS curve linearizer on every GDAL version because OGR has no corresponding geometry type

TIN and PolyhedralSurface rasterization remain unsupported, so this PR intentionally references rather than closes the broader Trac ticket.

## Review Follow-Up

- preserve the PostGIS fallback through GDAL 3.13
- defer standard OGR curve handling to GDAL 3.14 and newer
- keep direct and nested `NURBSCurve` inputs on the PostGIS linearizer path
- avoid cloning already-linear inputs
- test curved inputs, straight curve containers, nested collections, and `NURBSCurve` parity with `ST_CurveToLine`

References https://trac.osgeo.org/postgis/ticket/1168

## Validation

- `make -C raster/rt_pg -j32`
- `make staged-install`
- `make -C raster/test/regress check TESTS=rt_asraster RUNTESTFLAGS='--raster --verbose'`
- `make check-news`
- `make -C doc check`
- `git diff --check`

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -62,8 +62,8 @@ To take advantage of all postgis_sfcgal extension features SFCGAL 2.3+ is needed
           rasterization (Darafei Praliaskouski)
 
 
-PostGIS 3.7.0alpha1
-2026/07/05
+PostGIS 3.7.0dev
+2026/xx/xx
 
 This version requires GEOS 3.10+, PostgreSQL 14-19beta1, Proj 6.1+, libgmp.
 To take advantage of all features postgis extension features, GEOS 3.15+ is needed.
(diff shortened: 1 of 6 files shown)


dbaston:
FWIW, this is already done by GDAL.

dbaston:
> FWIW, this is already done by GDAL.

I mean that current GDAL linearizes curves internally, so no action from PostGIS is necessary. I'm surprised that whatever wrote this didn't check?

Komzpa:
@dbaston -- checked. It does not seem to linearize it by itself on my machine.

GDAL="GDAL 3.13.1 "Iowa City", released 2026/06/01

dbaston:
Looks like it's a 3.14 addition, though possibly could be backported to 3.13.2.

PR #1121, opened 2026-06-25
Fix typo doc ref overlay

jmonticolo:
In the page of [`ST_SymDifference`](https://postgis.net/docs/ST_SymDifference.html), there is a typo :

> Returns a geometry representing the portions of geonetries A and B that do not intersect.

This PR fixes the typo (geometries instead of geonetries) in 2 commits:
- One for the XML : reference_overlay.xml
- an other one for `.po` files.


Komzpa:
Thank you! Merged.

PR #1125, opened 2026-06-27
regress: stabilize GEOS 3.9 MIC output

Komzpa:
Stabilize the GEOS 3.9 maximum-inscribed-circle regression output on stable-3.3.

Jenkins Bessie currently fails `regress/core/geos39` on stable-3.3 because GEOS returns slightly different MIC coordinates/radius for `mic-box` and `mic-cte` than the historical expected file. Newer stable branches already avoid this exact precision sensitivity by checking the noisy MIC case at one decimal place.

This backports that stabilization pattern for the affected stable-3.3 checks only.

Validation:
- `git diff --check`
- `perl -c regress/run_test.pl`


robe2 (approved):

Komzpa:
Closing this because the patch has already landed on the canonical Gitea stable branches.

Accepted commits:

- stable-3.3: https://gitea.osgeo.org/postgis/postgis/commit/3794c40f33e773396b42d3b6f4a368f0e1b00903
- stable-3.2: https://gitea.osgeo.org/postgis/postgis/commit/b34a72a26ecadaeb73865ca4fcfec0245ec1bca9

GitHub kept this pull request open because the accepted commits are cherry-picks of the same patch rather than the original pull request head SHA.

PR #1117, opened 2026-06-22
Backport SFCGAL 2.3 regression test fixes

Komzpa:
## Summary

Backport the SFCGAL 2.3 regression-test fixes to `stable-3.6`:

- normalize and tolerate the SFCGAL 2.3 `CG_AlphaShape` / `CG_OptimalAlphaShape` output differences
- accept the reduced `CG_ExtrudeStraightSkeleton` patch counts produced by SFCGAL 2.3
- carry the upstream example output update and add the 3.6.5 NEWS entry

Trac ticket: https://trac.osgeo.org/postgis/ticket/6081

## Validation

Rebased on `stable-3.6` at `6e3bc696c` and tested locally with SFCGAL 2.3.0:

```bash
./autogen.sh
./configure --with-sfcgal --with-jsondir=/usr --with-projdir=/usr --with-raster --with-topology
make -j32
sudo -n make install
git diff --check upstream/stable-3.6..HEAD
./utils/check_news.sh .
PGPASSWORD=postgres PGHOST=127.0.0.1 PGPORT=5432 PGUSER=kom PGDATABASE=postgres \
  make -C sfcgal/regress check RUNTESTFLAGS="--extension --verbose" \
  TESTS="$(pwd)/sfcgal/regress/alphashape $(pwd)/sfcgal/regress/extrudestraigthskeleton"
```

Focused SFCGAL regressions passed both fresh install and automatic upgrade runs with `Postgis 3.6.5dev` and `SFCGAL 2.3.0`.

PR #945, opened 2026-06-17
sfcgal: stabilize SFCGAL 2.3 regression output (#6081)

Komzpa:
Closes #6081

This stabilizes the SFCGAL regression expectations for SFCGAL 2.3.0 on the
stable-3.6 branch.

Changes:
- normalize `alphashape` and `optimalalphashape` results through `ST_Multi` and
  `ST_Normalize` so equivalent polygon/multipolygon output and ring ordering do
  not fail the test
- replace the exact `CG_ExtrudeStraightSkeleton` roof WKT comparison with
  version-aware semantic checks for polyhedral output, coordinate dimension,
  non-empty output, and patch count
- accept the SFCGAL 2.3 building+roof patch count while preserving the older
  version branches
- add a 3.6.5 NEWS entry

Release / upgrade notes:
- NEWS is updated for PostGIS 3.6.5.
- No database extension upgrade SQL changes are needed; this only changes
  regression expectations and test SQL.

Validation:
- `./autogen.sh`
- `./configure --with-jsondir=/usr --with-projdir=/usr --without-raster --without-topology --with-sfcgal`
- `make -j32`
- `sudo -n make install`
- `make -C sfcgal/regress check RUNTESTFLAGS="--extension --verbose" TESTS="$(pwd)/sfcgal/regress/alphashape $(pwd)/sfcgal/regress/extrudestraigthskeleton"`

The focused SFCGAL regression run passed on SFCGAL 2.3.0, including the
automatic upgrade rerun.

PR #895, opened 2026-06-16
libpgcommon: qualify pg_class in SRID lookup

Komzpa:
## Summary

Qualify the `pg_class` catalog lookup used before querying `spatial_ref_sys` during SRID/PROJ lookup. This keeps the preflight check on the real PostgreSQL catalog even if a caller changes `search_path` so that another schema precedes `pg_catalog`.

## CSV finding covered

- Unqualified pg_class lookup enables SRID validation bypass.

## Release/Upgrade Notes

- Added a PostGIS 3.7.0 NEWS entry for https://github.com/postgis/postgis/pull/895.
- No extension upgrade SQL is needed: SQL definitions and catalog-visible objects are unchanged; this is libpgcommon runtime behavior.

## Current state

- Rebased on `postgis:master` at `4e470f1534795ae4a4c52ad5ad35b8744af9d708`.
- Current head: `f46ed7d7a8be7bc25d2338d18d2c393d33125b63`.
- GitHub readback after push: draft PR, `MERGEABLE`, approved, no review threads; CI checks are queued/pending.
- Gitea search found no matching active mirror PR for `895` or `fix-srid-pg-class-lookup`.

## Tests

- `git diff --check upstream/master..HEAD`
- `git clang-format --diff upstream/master -- libpgcommon/lwgeom_transform.c NEWS`
- `./utils/check_news.sh .`
- `make -C libpgcommon -j32`

PR #893, opened 2026-06-16
ci: harden utility scripts and doc version list

Komzpa:
## Summary

This collects small maintenance fixes that are not runtime PostGIS vulnerabilities but still improve hardening and correctness:

- pin the Codespell GitHub Actions install and install it without `sudo pip`
- make `check_all_upgrades.sh` use a private `mktemp -d` directory and fall back when `unbuffer` is unavailable
- remove the duplicate `3.6` entry from the generated documentation version list
- make the Winnie regression script tolerate unset optional environment flags

## Scope

Covers related CI, utility-script, and documentation maintenance follow-ups.

Rebased onto current `master` at `6f291bf547cf9b1834cfa63a779899f6ede4aeba` to pick up the shared SFCGAL NURBS expected-output stabilization that fixed the previous macOS-only regression failure. Final head is `e06ab2f4f307f80a489d56a9c6ae4c642d480537`.

## Validation

- `git diff --check origin/codex/fix-ci-doc-utility-findings..HEAD` for the new Winnie follow-up commit
- `bash -n utils/check_all_upgrades.sh`
- `bash -n ci/winnie/regress_postgis.sh`
- `xmllint --noout doc/xsl-config.xml`
- `shellcheck utils/check_all_upgrades.sh` still reports existing broad script warnings; the reviewed downgrade `SC2015` instance is gone, and `${MAKE_ARGS}` remains intentionally unquoted so callers can pass multiple make arguments.
- workflow text sanity check for the venv paths and pinned `codespell==2.4.1`

Fresh GitHub CI and CodeRabbit checks are queued on the final head.

PR #1127, opened 2026-06-27
Backpatch SFCGAL 2.3, tiger_geocoder, and address_standardizer fixes to stable-3.6

Komzpa:
## Summary

Roll up the stable-3.6 backpatch work for PostGIS 3.6.5.

This branch carries:

- SFCGAL 2.3 regression-test fixes from #1117 and #945
- split-repository `address_standardizer` hardening that still applies to pre-split stable branches
- split-repository `tiger_geocoder` loader/schema hardening that still applies to pre-split stable branches
- stable-3.6-applicable fixes from #887, #888, #889, #890, #891, #893, #894, #895, #897, and #898

## Scope cleanup

The earlier branch revision listed and carried #896 refresh leftovers. Those have been removed from this head after re-checking Paul Ramsey's #896 split. The stable branch already has the accepted geohash finite-loop fix from Paul; this PR no longer carries the later #896 exact-geohash/closed-TIN/patch-accessor follow-up chain.

#1132 is now only the master fuzzer smoke-link fix and is not a common #896 follow-up for this stable branch.

Current head: `c1168f6f4300bdc75174670aeb1b8c69e2585042`.

## Validation

- `git diff --check`
- `./utils/check_news.sh .`

GitHub Actions restarted for the current head after the branch update.


robe2 (changes requested):
This looks to be more than sfcgal.  This needs to be broken down into 3 separate pull requests

1) address_standardizer
2) tiger_geocoder
3) sfcgal regression tests

robe2 (changes requested):
Based on the discussion we had about these, I'm okay with this going in as a single pull request, but please change the subject line of it so it's clear it's more than just SFCGAL being pushed in.

something like 
Backpatch SFCGAL 2.3, tiger_geocoder, and address standardizer fixes to Stable 3.6

robe2 (approved):

PR #1131, opened 2026-06-27
Backpatch stable-3.2 hardening, tiger_geocoder, and address_standardizer fixes

Komzpa:
## Summary

Roll up the stable-3.2 backpatch work for PostGIS 3.2.11.

This branch carries:

- the stable-3.2 hardening rollup from #1126, including TWKB malformed-input hardening, the accepted geohash Inf-loop guard, ST_Split large-ordinate hardening, address_standardizer hardening, the DocBook XSL guard, and GEOS 3.9 MIC output stabilization
- split-repository `tiger_geocoder` loader/schema hardening that still applies to pre-split stable branches
- stable-3.2-applicable fixes from #893, #894, #895, #897, and #898

## Scope cleanup

The earlier branch revision listed and carried #896 refresh leftovers. Those have been removed from this head after re-checking Paul Ramsey's #896 split. This PR still keeps the explicit accepted geohash Inf-loop guard that belongs in the stable-3.2 hardening rollup, but no longer carries the later #896 exact-geohash/closed-TIN follow-up chain.

#1132 is now only the master fuzzer smoke-link fix and is not a common #896 follow-up for this stable branch.

Current head: `430983a56147dfc71a18caf9be6b773fb30893d0`.

## Validation

- `git diff --check`
- `./utils/check_news.sh .`

GitHub Actions restarted for the current head after the branch update.


robe2 (changes requested):
Can we break this up

1) geocode one should be in a single patch
2) Address standardizer one in another patch
3) The winnie one I have to check on, I think that may cause some problems, probably better to leave this out and I'll manually patch if I feel it's needed
4) liblwgeom twkb
5) libwgeom geos split
6) geos39 tests

Komzpa:
@robe2 they are purposefully on different commits and I plan to commit this keeping this structure, will it be fine that way?

robe2:
> @robe2 they are purposefully on different commits and I plan to commit this keeping this structure, will it be fine that way?

Okay fine then.

robe2 (approved):

robe2:
@Komzpa  the only concern is the 9.6 failure.  I know 9.6 is EOL, but we did say we'd support it for this branch, so that would be considered a breaking change.  Can you fix that one.

robe2:
Looks all good.

PR #1134, opened 2026-06-30
Avoid GEOSSplit for ST_Split

Komzpa:
## Summary

The `CI (latest, garden)` job on `master` is crashing in `ST_Split` with the GEOS 3.15 build image:

https://github.com/postgis/postgis/actions/runs/28475013328/job/84397280248

The crash comes from the new `GEOSSplit` path added for `ST_Split`. This change keeps the existing `GEOSDifference` line split path and the polygonize-based polygon split path active for GEOS 3.15 too, instead of routing those cases through `GEOSSplit`.

I also added focused regression coverage for the garden-shaped cases that use a `MULTILINESTRING Z` blade against line and polygon inputs.

## Testing

Ran the focused split regression in a `postgis/postgis-build-env:latest` container with GEOS `3.15.0dev` and PostgreSQL `20devel`:

```sh
make -C regress check TESTS="$(pwd)/regress/core/split"
```

Both the normal run and the automatic upgrade rerun passed:

```text
regress/core/split .. ok
Run tests: 3
Failed: 0
```

I also ran a local garden-style probe without the `logbt` wrapper. The local container's PostgreSQL build lacks XML support, so it is not a clean full `make garden` equivalent, but the generated garden SQL advanced past the previous `ST_Split` crash region into later `ST_Angle` cases without a server-disconnect or SIGSEGV signature.


Komzpa:
Closing this one. The garden crash has been diagnosed as a GEOS 3.15 / build-image issue, not something PostGIS should work around by backing out `GEOSSplit`.

The right fix is on the GEOS/build-env side, so keeping PostGIS on the `GEOSSplit` path is still the intended direction.

PR #1138, opened 2026-07-06
sfcgal: pass roof height before slope

Komzpa:
## Summary
- Addresses security scanner finding: CG_GenerateRoof swaps height and slope arguments.
- Pass the SQL height argument into SFCGAL before the slope angle so roof generation keeps the advertised API order.

## Backpatch notes
- Backpatch is a one-line argument order fix in the SFCGAL wrapper; older branches need the same call-site check if the wrapper is present.

## Validation
- make -C sfcgal -j32; git diff --check

### DIFF
--- a/sfcgal/lwgeom_sfcgal.c
+++ b/sfcgal/lwgeom_sfcgal.c
@@ -2148,7 +2148,7 @@ sfcgal_generate_roof(PG_FUNCTION_ARGS)
 	geom = POSTGIS2SFCGALGeometry(input);
 	PG_FREE_IF_COPY(input, 0);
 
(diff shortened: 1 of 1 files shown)


Komzpa:
Closing this one as a false positive from the scanner finding. The SFCGAL C API signature is , and the existing regression already checks that generic  matches the specialized gable/skillion wrappers. Swapping the call made those checks fail in CI.

Komzpa:
Correction to the close note, with shell-sensitive formatting avoided: the SFCGAL C API signature is sfcgal_geometry_generate_roof(geom, roof_type, slope_angle, height, primary_edge_index). Current master already passes slope_angle before height and has regression coverage checking generic CG_GenerateRoof against the specialized gable/skillion wrappers. The proposed swap was therefore wrong, and CI caught it.

PR #1143, opened 2026-07-06
topology: clean up TopoRingIsCCW early returns

Komzpa:
## Summary
- Addresses security scanner finding: Memory leak on empty TopoRingIsCCW input.
- Frees deserialized geometry and copied varlena input before empty and unsupported-type exits.

## Backpatch notes
- Small cleanup in one topology C function; should cherry-pick directly to branches with the same early returns.

## Validation


Komzpa:
Landed on canonical Gitea master as 9000ef67ed38616b10e06d8da7bd012f9d866aa0 and mirrored to GitHub master.

PR #1148, opened 2026-07-06
utils: create upgrade-check tempdir safely

Komzpa:
## Summary
- Addresses security scanner finding: Predictable temporary directory in new check script.
- Uses mktemp -d for the upgrade-check script temporary directory instead of a predictable PID-based path.

## Backpatch notes
- Script-only fix; backpatch by applying the tempdir creation change and keeping the existing trap cleanup.

## Validation
- sh -n utils/check_create_upgrade_replaced_function_errors.sh; git diff --check

### DIFF
--- a/utils/check_create_upgrade_replaced_function_errors.sh
+++ b/utils/check_create_upgrade_replaced_function_errors.sh
@@ -2,10 +2,8 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


Komzpa:
Landed on canonical Gitea master as 8a03193ebc452f335c0f51c26dbdbb53b1a15904 and mirrored to GitHub master.

PR #1149, opened 2026-07-06
topology: fix corrupt column fallback variables

Komzpa:
## Summary
- Addresses security scanner finding: New topology repair function fails at runtime.
- Uses the function arguments in the fallback branch instead of nonexistent local variable names.

## Backpatch notes
- Small runtime fix in FixCorruptTopoGeometryColumn; should cherry-pick directly where that helper exists.

## Validation


Komzpa:
Landed on canonical Gitea master as 6ea8421662ed55c8b895e1a6e4d4ef3edefde4c6 and mirrored to GitHub master.

PR #1153, opened 2026-07-06
topology: preserve large ids from edge wrappers

Komzpa:
## Summary
- Addresses security scanner finding: Bigint topology support has truncation and metadata bugs.
- Uses LWT_ELEMID for topology wrapper return values so large edge ids are not narrowed through int.

## Backpatch notes
- C-only type widening in four wrappers; backpatch to branches carrying bigint topology support.

## Validation


Komzpa:
Landed on canonical Gitea master as 6fff4b58bbf6b671fdb34a0bbaa70d1f8ab11b2c and mirrored to GitHub master.

PR #1137, opened 2026-07-01
Run manual SQL examples in garden tests

Komzpa:
## Summary

This wires runnable manual SQL examples into the garden regression workflow with a minimally invasive DocBook convention.

The new Python helper reads `doc/postgis-out.xml`, finds SQL `programlisting` examples with adjacent expected-output `screen` blocks, and runs the parseable examples against a fresh `postgis_exampletest` database. Examples are tested by default when they are self-contained and parseable. Listings that need tables, files, installed driver/catalog state, prior setup, or other state outside the standard example-test database are marked with `role="requires-external-state"` and are reported but not run.

The PR also normalizes example markup so the tester does not need display-oriented special cases:

- SQL input uses `programlisting language="sql"`.
- Expected psql/manual output uses the following `screen` element.
- Terminal sessions and commands such as `psql`, `\d`, `shp2pgsql`, `pgsql2shp`, `raster2pgsql`, and Java command-line invocations use `screen`.
- SQL comments that were acting as prose were moved into nearby `para` text where practical.
- One-column output headers are dropped for one-line scalar outputs where the rendered result is clearer without them.
- Volatile version-reporting and extension-catalog examples are still executed, but comparison avoids exact library version strings and optional extension row counts.

The regression integration adds `make -C regress exampletest-run` as a focused developer target and runs it from the existing top-level `make garden` path after the generated garden SQL tests. The helper can also report coverage-style stats without running the examples.

Current report from `utils/postgis_exampletest.py --report doc/postgis-out.xml`:

```text
programlisting_total=1437
example_tests=610
parseable_example_tests=610
requires_external_state=23
volatile_version_tests=15
```

## Validation

- `git diff --check`
- `make -C doc postgis-out.xml`
- `make -C doc check`
- `utils/postgis_exampletest.py --report doc/postgis-out.xml`
- `make -C regress exampletest-run` (`manual example tests passed: 610`)
- `make -n garden` includes `exampletest-run`
- `make -n check` does not include `exampletest-run`
- `codex review --uncommitted`

PR #1161, opened 2026-07-08
Improve DocBook examples and generated geometry figures

Komzpa:
## Summary

This fixes and hardens the manual rendering regression exposed by the recent SQL/output split in postgis/postgis#1137 and https://gitea.osgeo.org/postgis/postgis/pulls/360.

The root cause is that DocBook `programlisting` blocks and `screen` blocks were both rendered by the upstream HTML/PDF stylesheets as visually similar bare listings. After runnable SQL examples started moving input to `programlisting` and output to `screen`, current manual pages could distinguish the XML source while still making SQL input and query output look the same to readers.

This PR keeps XML as the source of truth and adds small, centralized rendering fixes:

- HTML wraps `programlisting` and `screen` output with visible, real-text `Code` / `Output` labels.
- HTML wrappers expose `role="group"`, `aria-labelledby`, stable `postgis-example-code` / `postgis-example-output` classes, and `data-postgis-block="code"` / `data-postgis-block="output"`.
- CSS makes adjacent Code + Output blocks read as one pair while preserving a text label, a straight 1px border, balanced header/listing padding, print/no-color distinction, a reasonable max width on wide displays, and wrapped code/output with no horizontal scrolling.
- HTML adds a small dependency-free copy button for code blocks only. Legacy mixed code/output blocks are deliberately not copyable until their XML is split.
- HTML preserves or conservatively infers SQL language metadata for safe input-only blocks (`language-sql`, `data-postgis-language="sql"`) and adds a small dependency-free SQL highlighter that runs only on those code blocks. Highlighted PostGIS function/operator tokens link to their manual pages when present and expose the refpurpose as a native hover title; link styling uses a quiet solid underline instead of noisy dotted underlines. `screen` output deliberately gets no SQL/highlight metadata.
- PDF/dblatex output now emits real `Code` / `Output` text before listings, using the same gettext-backed label strings from `doc/xsl-config.xml`.

The labels are generated as real text, not CSS pseudo-content. English source labels live in `doc/xsl-config.xml` and translations are carried by the existing `.po`/gettext manual pipeline, not by hardcoded language tables in XSL.

## Docs markup hardening

This adds/extends repo-side checks for the generated manual:

- `utils/check_docbook_block_labels.py` verifies all generated chunked HTML pages have labeled code/output wrappers, SQL language/highlighter hooks on SQL-looking code input, copy buttons only where safe, and no SQL/highlight classes or data attributes on `screen` output.
- `utils/build_docbook_ref_index.py` generates the small HTML-side function/operator index used by the highlighter links; when Python is unavailable the docs build writes an empty fallback index rather than failing.
- `utils/check_docbook_markup.py` now reports structured findings with severities and explicit `--fail-on error|warning|none` behavior.
- `make -C doc check` runs these Python-based checks when Python 3 is available, skips them with an explicit message in minimal CI images without Python, and fails on definite structural errors when the checker runs. Legacy mixed examples remain report-only warnings so unrelated docs work is not blocked by the remaining old schema universe.

Current warning categories are intentionally still visible for follow-up cleanup.

## Targeted XML cleanups

Kept deliberately small and reviewable:

- split the mixed SQL/output `ST_AsKML` example and moved its availability/enhanced/changed metadata out of note blocks
- split `ST_AsGML` SQL input from GML/XML output examples
- split `ST_GeometryN` table output from SQL input and made the example order deterministic for manual exampletest
- split the obvious `box2df` / `gidx` / geometry operator examples, including `overlaps_geometry_box2df`, so those pages now render as SQL Code plus Output
- renamed the second `ST_MakeValid` `Examples` section to `Collapsed Geometry Examples`

The remaining mixed examples are reported by the checker, not mass-rewritten in this PR.

## Validation

Ran locally:

```sh
./config.status doc/Makefile
make -C doc chunked-html html check
make -C doc exampletest-report
make -C regress exampletest-unit
make postgis_revision.h
make -C doc pdf
make -C doc/po/fr local-html
rm -f doc/html/postgis-ref-index.json && make -C doc html PYTHON3=
rm -f doc/html/postgis-ref-index.json && make -C doc html
for f in doc/po/templates/postgis-manual.pot doc/po/*/postgis-manual.po; do msgfmt -c -o /dev/null "$f"; done
node --check doc/html/copy-code.js
python3 -m py_compile utils/check_docbook_block_labels.py utils/check_docbook_markup.py utils/build_docbook_ref_index.py
xmllint --noout   doc/xsl-config.xml   doc/xsl/postgis-block-labels-common.xsl   doc/xsl/postgis-html-block-labels.xsl   doc/xsl/postgis-dblatex-block-labels.xsl
git diff --check
make -C doc check PYTHON3=
make -C doc check
make -C regress exampletest-run
make -C doc garden
make -C regress garden
# In postgis/postgis-build-env:latest (GEOS 3.15.0beta2):
make -C doc garden && make -C regress exampletest
```

The `postgis/postgis-build-env:latest` Docker run was used specifically to reproduce the GitHub `latest` garden environment; it reported `manual example tests passed: 637`. After rebasing onto current `master` (`b83955d00`), the focused local revalidation also passed: `make -C doc check`, `make -C doc garden`, `make -C regress exampletest-run`, and `make -C regress garden`.

Also inspected generated HTML/PDF for `is_contained_geometry_box2df`, `ST_AsKML`, `ST_AsGML`, `ST_GeometryN`, `ST_MakeValid`, and `geometry_overlaps`. Browser inspection confirmed the known bad `is_contained_geometry_box2df` example now shows a Code block with a copy button followed by a labeled Output block; at a 1920px viewport, example block width is capped at 72rem / 1152px instead of spanning the whole page. The Code block has `language-sql`, `data-postgis-language="sql"`, wrapped `pre-wrap` rendering with no horizontal scroll, straight borders, balanced padding, and SQL highlight/link spans after JS runs. In that example `ST_Buffer`, `ST_GeomFromText`, `ST_MakeBox2D`, `ST_Point`, and the `@` operator are links with hover titles; the `@` operator resolves to the current operator page. Also inspected `overlaps_geometry_box2df`: it now has a highlighted SQL Code block for the `&&` example and a separate plain Output block. The Output block stays plain `pre.screen` with no SQL/highlight metadata. The copy button copies only SQL input.

The current DocBook markup checker summary is:

```text
44 finding(s): warning=44; metadata-role-in-admonition=29, mixed-programlisting-output=13, screen-contains-sql=2
```

Those are report-only legacy/content warnings unless run with `--fail-on warning`.

---

## Update: legacy conversion finished, checkers consolidated

All 44 legacy checker warnings are now fixed, so no special-casing remains:

- Split the remaining 13 mixed SQL/output `programlisting` blocks and moved SQL out of the 2 `screen` blocks that carried input (ST_IsValidDetail, ST_IsValidReason, ST_Geometry_EQ, ST_Centroid, ST_GDALDrivers, ST_Value, ST_SetValue, ST_SetValues, ST_MapAlgebraFct, plus two chapter-level examples). Content was re-homed, not rewritten: per-refentry paragraph multisets and code/output text were verified unchanged modulo whitespace against the previous head.
- Unwrapped the 29 `para role="changed|enhanced|availability"` from admonitions in place, same integrity verification.
- Copy buttons now appear on all code blocks; the legacy mixed-block gating is removed.
- `check-docbook-markup` runs with `--fail-on warning --max-warnings 0`: any new mixed block, SQL-in-screen, or metadata-in-admonition fails `make -C doc check`.

Checker integration rework:

- The three scripts are consolidated into `utils/docbook_qa.py` (`lint-source`, `lint-html`, `ref-index` subcommands) with a shared findings/severity model and unit tests wired into the existing check plumbing. Findings output is byte-equivalent to the old scripts on the manual corpus.
- The hardcoded `DOCBOOK_BLOCK_LABEL_CHECK_PAGES` page whitelist is gone: `lint-html --expect-from postgis-out.xml` derives required Code/Output wrappers for every refentry page from the source XML.
- New `docs-check` job in GitHub Actions CI lints the DocBook sources and runs the tool's unit tests; needs only libxml2-utils + python3, no configure.

Example-test fallout from making old examples runnable:

- `utils/postgis_exampletest.py` gained `--keep-going` (report all failures plus summary; default fail-fast unchanged), with unit tests.
- The two quintuplet validation examples are marked `role="requires-external-state"`: their self-intersection coordinates vary across GEOS versions.
- Three screen blocks showed stylized chrome psql never prints (`--pt_intersect --` header, bare separators); they now show the genuine psql output for the same results.

Validation: `make -C doc check` (strict), `docbook_qa.py` unit tests, `make -C regress exampletest-unit`, local `exampletest-run` (only pre-existing `--without-protobuf` env failures remain, matching the base-branch baseline), and a full configure/build/staged-install/exampletest run inside `postgis/postgis-build-env:latest`: **manual example tests passed: 645** (was 637).

## Update: build-time geometry figures and standard-library XML

The manual examples can now mark geometry output with `role="visual-primary"`. The example test runs the documented query against the staged PostGIS build, asks that same PostGIS build for `ST_AsSVG`, and writes a deterministic SVG asset. HTML and PDF consume that one build-time asset, so the figure does not depend on browser JavaScript and is not lost in PDF output. In HTML, Figure is part of the same visual example card as Code and Output; the textual Output remains available and can be collapsed when the picture is more useful.

`ST_AsSVG` now preserves curve primitives where SVG has a direct equivalent:

- circular strings and curved polygons use SVG elliptical arc (`A`/`a`) commands
- one-span polynomial NURBS use exact quadratic or cubic Bezier (`Q`/`C`) commands
- general rational or multi-span NURBS retain the existing sampled fallback because SVG has no rational B-spline path primitive

The SVG tests cover circular strings, compound curves, curve polygons, multisurfaces, polynomial and weighted NURBS, and curve members nested in collections.

The DocBook QA and example-test XML readers no longer require `lxml`. A small shared reader built from `xml.sax` and `xml.etree.ElementTree` preserves source line numbers plus parent/sibling traversal using only the Python standard library. `xmllint` remains the build step that expands and validates DocBook DTD/XInclude input before Python sees it.

Additional validation for the stdlib conversion:

```text
python3 -S utils/test_docbook_qa.py          21 tests passed
python3 -S utils/test_postgis_exampletest.py 17 tests passed
make -C doc check                            passed
```

On the generated manual corpus, the stdlib implementation produces output identical to the previous implementation for source lint, all-page HTML lint, the manual example report, and both the JSON and JavaScript reference indexes.

## Update: canonical SFCGAL names and generated Voronoi/Delaunay figures

The SFCGAL reference now treats the `CG_*` names as canonical throughout the current manual:

- all 19 deprecated SFCGAL `ST_*` aliases live in a separate compatibility chapter with only their legacy signatures and one migration link
- normal cross-references, function lists, support matrices, cheatsheets, day cards, aggregate indexes, and garden generation use the canonical `CG_*` pages
- searching or highlighting a deprecated spelling opens its canonical `CG_*` page
- DocBook QA rejects new links or examples that use a deprecated alias outside its compatibility entry
- the `ST_MinkowskiSum` runtime deprecation notice now spells `CG_MinkowskiSum` correctly

The remaining authored Voronoi and constrained-Delaunay raster examples were also moved onto the common build-time visual path:

- `ST_VoronoiPolygons` keeps its two existing point sets and outputs, but no longer duplicates them with small legacy PNGs
- `CG_ConstrainedDelaunayTriangles` and its `ST_DelaunayTriangles` comparison now use runnable Code + Output pairs for the original polygon and buffered point geometry
- HTML embeds the generated SVG figures; PDF embeds rasterizations of those same build-time assets
- portable SVG presentation attributes keep strokes and fills visible in fallback PDF rasterizers, and point-only layers render last so input points remain visible above area outputs
- the four obsolete WKT raster-generator fixtures and their Makefile entries are removed

Additional validation on head `27b9a44e2b`:

```text
make -C regress visual-examples-run
manual example tests passed: 413
visual examples rendered: 413

make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.

make -C doc html
make -C doc chunked-html
make -C doc pdf
```

## Update: remaining processing figures and curve edge cases

The common build-time renderer now replaces the legacy raster illustrations for `ST_Buffer`, `ST_BuildArea`, `ST_ConcaveHull`, `ST_LineMerge`, `ST_LargestEmptyCircle`, `ST_MinimumBoundingCircle`, `ST_Transform`, and the tapered-buffer example. The documented SQL is the figure source, so HTML and PDF show the same input and output geometries without parallel WKT image fixtures.

The generated figures also gained a few presentation and geometry fixes found during the manual-wide visual audit:

- geometry layers use reader-facing `Code` and `Output` legend labels rather than SQL column aliases such as `st_astext`
- point layers render above polygon fills, with a contrasting halo so boundary and center points remain visible
- line arrows distinguish direction without adding arrows to closed rings
- layers with different known SRIDs are normalized before comparison, avoiding misleading transformed examples
- adjacent Code, Output, and Figure sections remain one example card in HTML

`ST_AsSVG` now emits collinear `CIRCULARSTRING` triples as straight SVG line segments. This avoids invalid negative arc radii while preserving the middle control point. CUnit covers both absolute and relative path output; non-collinear circular strings continue to use SVG arc commands.

Local validation on head `70e85fbe77`:

```text
make -C regress visual-examples-run
manual example tests passed: 418
visual examples rendered: 418

make -C liblwgeom/cunit check
358 tests passed, 0 failed

make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.

make -B -C doc html chunked-html
make -C doc pdf
```

The resulting 1,143-page PDF was raster-inspected at the affected function pages; generated figures retained colors, transparency, points, curves, and `Code` / `Output` labels.

## Update: accessor/editor figures and text-primary output

The same conversion now covers the legacy WKT rasters on `ST_Boundary`, `ST_Envelope`, `ST_Segmentize`, and `ST_Snap`. Their Code, Output, and generated figures are full-width instead of being constrained by old two-column image tables; nine more WKT image fixtures were removed. A single geometry result is now always labelled `Output`, while meaningful column names remain in figures that compare multiple result geometries.

Examples may mark a `screen` as `role="text-primary"` when the exact type or coordinates carry information that the picture cannot show. HTML then leaves Output expanded while retaining the figure. `ST_ForceCurve` uses this for its `CURVEPOLYGON Z` result, so the essential upcast is visible without opening a collapsed coordinate block. The role is documented in both contributor guides.

Local validation on head `ce7d038bb1`:

```text
python3 utils/test_postgis_exampletest.py
33 tests passed

make -C regress visual-examples-run
manual example tests passed: 418
visual examples rendered: 418

make -B -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: example introductions and distance-line figures

Plain prose immediately before a Code block now joins the same visual card as its Code, Output, and Figure sections. This is progressive CSS: examples without an introduction retain the existing card shape, and browsers without `:has()` retain the previous standalone paragraph layout. The `ST_SnapToGrid` mixed `UPDATE`/`SELECT` listing is split into a standalone update example and a runnable query with its own Output and generated Figure.

The generated figures also replace five legacy WKT rasters for `ST_LongestLine` and `ST_ShortestLine`. Authored `ST_Azimuth` and `ST_HausdorffDistance` diagrams remain because their angle and multi-metric annotations carry information the generic geometry renderer does not yet express.

Local validation on head `774c0d58d1`:

```text
make -C regress visual-examples-run
manual example tests passed: 419
visual examples rendered: 419

make -B -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: overlay and polygonization figures

Generated figures now replace nine more WKT rasters across `ST_Difference`, `ST_SymDifference`, `ST_Polygonize`, and `ST_Split`. The first `ST_Split` example no longer documents an abbreviated `POLYGON(..)` placeholder: it contains the real precision-reduced PostGIS result, and its blade is expressed as the same `LINESTRING(10 10,190 190)` geometry so the common extractor renders the complete cutting line instead of two disconnected constructor points.

Local validation on head `eb27f6636`:

```text
make -C regress visual-examples-run
manual example tests passed: 420
visual examples rendered: 420

make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: derived processing figures

Four more WKT rasters are replaced for `ST_GeometricMedian`, `ST_MaximumInscribedCircle`, and `ST_OrientedEnvelope`. The point-only visual skip is explicitly overridden for the median/centroid comparison. Both maximum-inscribed-circle examples now derive a circle and radius line from the function's returned center, nearest point, and radius, so the common figure preserves every meaningful layer from the authored illustrations for polygonal and lineal inputs.

Local validation on head `c95f8c326`:

```text
make -C regress visual-examples-run
manual example tests passed: 422
visual examples rendered: 422

make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: centroid and point-on-surface figures

The four authored source geometries from the old `ST_Centroid` and `ST_PointOnSurface` illustrations are now runnable examples rather than parallel WKT fixtures: MultiPoint, LineString, Polygon, and mixed GeometryCollection. Point-only output is explicitly visualized when it is meaningful in context, and output points render above the source geometry. The separate point-on-surface versus centroid comparison also uses the common renderer. Nine WKT raster fixtures were removed without replacing the source shapes with synthetic squares.

Local validation on head `7a799e244`:

```text
make -C regress visual-examples-run
manual example tests passed: 428
visual examples rendered: 428

make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: coverage figures

Eight legacy WKT rasters are replaced across `ST_CoverageInvalidEdges`, `ST_CoverageClean`, `ST_CoverageSimplify`, and `ST_CoverageUnion`. The clean example is now one self-contained query: its generated figure overlays the four original polygons, cleaned coverage, and invalid edges, preserving the three old static stages as inspectable geometry layers. The simplify and union examples keep their original authored geometries and use their existing runnable outputs.

The freshly rebuilt `ST_SnapToGrid` page also confirms that the standalone column `UPDATE` and the runnable LineString `SELECT` are separate example cards rather than one mixed listing.

Local validation on head `5d6ed1a06`:

```text
make -C regress visual-examples-run
manual example tests passed: 429
visual examples rendered: 429

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: linear referencing figures

Four more legacy rasters are removed from `ST_LineInterpolatePoint`, `ST_LineInterpolatePoints`, `ST_LineSubstring`, and `ST_LocateBetween`. The generated figures use the same authored LineStrings and exact documented results, retain the old captions as example introductions, and add direction arrows and interactive input/output layers.

Local validation on head `d035c6568`:

```text
make -C regress visual-examples-run
manual example tests passed: 429
visual examples rendered: 429

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: hull, dumped vertices, and generated points

The old `ST_ConvexHull`, `ST_DumpPoints`, and `ST_GeneratePoints` rasters are now generated from runnable examples. `ST_DumpPoints` keeps its path table as textual output and adds a compact `ST_Collect` example for the figure, avoiding a useless 29-item legend while preserving every dumped vertex. `ST_GeneratePoints` documents the current deterministic seed result rather than freezing coordinates produced by an older point generator; its figure shows the source line, buffered area, and points as separate layers.

Local validation on head `c585a8f65`:

```text
python3 utils/test_postgis_exampletest.py
33 tests passed

make -C regress visual-examples-run
manual example tests passed: 431
visual examples rendered: 431

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: simplification figures

Six more WKT rasters are replaced for `ST_Simplify`, `ST_SimplifyPreserveTopology`, and `ST_SimplifyVW`. Each runnable example now emits its actual geometry output, so the generated figure overlays the original authored geometry and simplified result. The second `ST_SimplifyVW` raster had drifted away from the SQL beneath it; the query now uses the polygon from that illustration and records the result produced by the current implementation.

Local validation on head `cdd75936c`:

```text
make -C regress visual-examples-run
manual example tests passed: 437
visual examples rendered: 437

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: shared-path figures

The two `ST_SharedPaths` rasters and their one-column layout table are replaced by the common example card. The generated figure keeps both authored input line sets, highlights the shared paths, and retains line direction; the flipped-orientation query remains a separate textual example.

Local validation on head `a37a1b178`:

```text
make -C regress visual-examples-run
manual example tests passed: 437
visual examples rendered: 437

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: closest-point and Voronoi figures

Three more WKT rasters are removed from `ST_ClosestPoint` and `ST_VoronoiLines`. The polygon-to-buffer closest-point query now returns the buffered polygon as a named geometry layer, so the generated figure preserves both old polygons, the buffer center, and the closest point on top of the boundary.

The SQL geometry extractor also no longer mistakes type substrings inside function names such as `ST_ClosestPoint` or `ST_MakePolygon` for a WKT token and then skips their real arguments. A focused parser test covers this. Fixing the generic lexer exposes 20 additional valid build-time figures, including point editors/accessors and NURBS inspection examples.

Local validation on head `12917d22f`:

```text
python3 utils/test_postgis_exampletest.py
34 tests passed

make -C regress visual-examples-run
manual example tests passed: 457
visual examples rendered: 457

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: triangulation comparison

The old side-by-side original/triangulated rasters in the second `ST_TriangulatePolygon` example are removed; its existing runnable query already generates one interactive figure with both authored layers. The separate letter-P raster remains until its large triangulation result is promoted into a checked Output, so this batch does not trade coverage for a missing figure.

Local validation on head `e079cc5f1`:

```text
make -C regress visual-examples-run
manual example tests passed: 457
visual examples rendered: 457

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: subdivision figures

Both `ST_Subdivide` WKT rasters are removed. Their existing checked examples already generate the five polygon subdivisions and six densified geography sublines; the common SVG also retains the original polygon or line as a comparison layer and preserves direction markers.

Local validation on head `7a25ae38e`:

```text
make -C regress visual-examples-run
manual example tests passed: 457
visual examples rendered: 457

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: line-crossing direction figures

The three `ST_LineCrossingDirection` rasters and one orphaned WKT fixture are removed. Their generated figures preserve the exact authored LineStrings and use direction markers. The generic SQL extractor now recognizes aliases on quoted `::geometry` and `::geography` expressions, so the legends retain `lineA` and `lineB` instead of falling back to numbered Code layers.

Local validation on head `592c46b5b`:

```text
python3 utils/test_postgis_exampletest.py
35 tests passed

make -C regress visual-examples-run
manual example tests passed: 457
visual examples rendered: 457

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: overlap relationship figures

The three WKT rasters attached to runnable `ST_Overlaps` examples are removed. Their generated figures preserve the Point/LineString, Polygon/LineString, and Polygon/Polygon geometry pairs, keep the `a`/`b` aliases, and render boundary points above linework. The separate three-image table of true overlap cases remains until equivalent checked SQL examples are added.

Local validation on head `8c3c3ad33`:

```text
make -C regress visual-examples-run
manual example tests passed: 457
visual examples rendered: 457

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: touches relationship figures

All six standalone `ST_Touches` rasters are now checked SQL examples using their original authored geometries, producing six build-time SVGs with `a`/`b` layers. Running the examples exposed an old data error in the first claimed-true case: its polygons had a one-unit gap (`x=80` versus `x=81`). The second boundary is corrected to `x=80`, preserving the intended drawing while making `ST_Touches(a,b)` actually return true.

Local validation on head `7e495d592`:

```text
make -C regress visual-examples-run
manual example tests passed: 463
visual examples rendered: 463

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: overlap truth-table figures

The remaining MultiPoint, LineString, and Polygon `ST_Overlaps` rasters are converted into checked SQL examples using their original geometries. The point-only comparison uses the existing `visual-primary` author override; all three generated figures retain `a`/`b` layer names and visibly show the shared subset or area.

Local validation on head `7c3705e39`:

```text
make -C regress visual-examples-run
manual example tests passed: 466
visual examples rendered: 466

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: contains relationship figures

All eight standalone `ST_Contains` rasters are now checked SQL examples using the original authored geometries. The generated figures preserve the true, false, and boundary-only groups, including the polygon hole and the line extending outside its polygon. The two boundary-only examples execute both `ST_Contains` and `ST_Covers`, verifying the intended `f | t` distinction instead of leaving it as an untested caption.

Local validation on head `9e103041f`:

```text
make -C regress visual-examples-run
manual example tests passed: 474
visual examples rendered: 474

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: crosses relationship figures

All four standalone `ST_Crosses` rasters are now checked SQL examples using the original authored geometry pairs. The first legacy fixture claimed a true Point/Line crossing while its nearest point was one unit off the line (`130 119` versus the line vertex `130 120`); the point is corrected to the intended vertex, and all four predicates now execute as true.

Local validation on head `da2a6c27b`:

```text
make -C regress visual-examples-run
manual example tests passed: 478
visual examples rendered: 478

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: offset curve figures

The six `ST_OffsetCurve` raster/code table cells are replaced by full-width checked SQL examples and generated figures. This exposed outputs and prose frozen to pre-GEOS-3.11 direction behavior: negative offsets now retain input direction, and the double negative offset combines to -45 rather than -15. The expected WKT and explanation now match the executed current behavior while retaining the original input line and all six join/offset comparisons.

Local validation on head `b227a869e`:

```text
make -C regress visual-examples-run
manual example tests passed: 484
visual examples rendered: 484

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: smoothing figures

The six `ST_ChaikinSmoothing` rasters and four `ST_CatmullRomSmoothing` rasters are replaced by full-width checked comparisons. Each function now uses one Polygon example and one LineString example that overlays the original geometry with all documented smoothing levels. Chaikin labels its three outputs by iteration; Catmull-Rom uses named `segments_5` and `segments_10` layers and rounds text output to three decimal places to avoid exposing floating-point noise. The existing Chaikin triangle example is also converted from in-band sample output to a checked Code/Output/Figure block.

Local validation on head `daf9f4137`:

```text
make -C regress visual-examples-run
manual example tests passed: 489
visual examples rendered: 489

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: polygon hull figures

All three `ST_SimplifyPolygonHull` rasters are replaced by checked generated comparisons. The outer and inner Polygon examples retain the original authored Polygon. The `ST_Letters('xt')` example now renders the current function output rather than the stale stored fixture, and returns both named `input` and `outer_hull` layers; its visible input remains the unsegmentized equivalent linework while the hull is computed from the documented segmentized geometry.

Local validation on head `a17618512`:

```text
make -C regress visual-examples-run
manual example tests passed: 492
visual examples rendered: 492

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: make-valid figures

The six `ST_MakeValid` rasters are replaced by two checked, full-width comparisons using the exact original MultiPolygon inputs. Each generated figure names the input `before_geom` and overlays the default linework result with `after_geom_structure`, preserving the old before/after/method distinction without three detached images. Text WKT is rounded to three decimal places for stable readable output.

Local validation on head `f60d96bd1`:

```text
make -C regress visual-examples-run
manual example tests passed: 494
visual examples rendered: 494

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: measurement figures and comparison labels

The two `ST_Azimuth` rasters are replaced by checked figures that retain the original origin/target points and north rays, with a direction line and arrow generated from the query. The `ST_HausdorffDistance` raster is replaced by a checked comparison of both source lines, the original Hausdorff witness segment, and `ST_ShortestLine`; the witness length is verified against the computed Hausdorff distance. The common renderer now keeps a meaningful single geometry alias when it participates in a comparison, while standalone one-output figures still use the concise `Output` label and generic `st_astext` headers remain filtered.

Local validation on head `372e4b8bd`:

```text
python3 utils/test_postgis_exampletest.py
36 tests passed

make -C regress visual-examples-run
manual example tests passed: 496
visual examples rendered: 496

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: constructor and triangulation figures

The standalone `ST_Letters('Yo')` raster now comes from the current function output, rounded to three decimal places, rather than the stale stored font geometry. The `ST_TriangulatePolygon` letter-P raster is replaced by a checked generated figure using the exact original Polygon and its full GeometryCollection of triangles; individual triangle parts remain visibly distinguished by the common multipart palette.

Local validation on head `d076a1d3c`:

```text
make -C regress visual-examples-run
manual example tests passed: 498
visual examples rendered: 498

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: within figure

The `ST_Within` circle raster is now produced by the same checked query that verifies all six documented boolean relationships. A single CTE defines the center and both buffered circles; the generated figure names `center`, `small_circle`, and `big_circle`, while the complete boolean and WKT output remains available as text.

Local validation on head `573900ccd`:

```text
make -C regress visual-examples-run
manual example tests passed: 499
visual examples rendered: 499

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: documented SFCGAL outputs and PDF figures

The five `CG_AlphaShape` and `CG_OptimalAlphaShape` rasters are replaced by generated SVG figures built from their complete documented WKT outputs. These SFCGAL functions intentionally remain outside exact-output execution because component ordering may vary across supported library versions. An explicit `screen role="visual-primary"` now opts such an example into build-time WKT parsing and rendering without weakening the default skip rules or executing the unstable function.

For visual-preferred examples, HTML retains the text output in its collapsible block. The non-interactive PDF uses the equivalent generated figure instead of typesetting potentially enormous WKT collections; `role="text-primary"` continues to retain text where the literal representation is the important result.

Local validation on head `2457f0fa0`:

```text
python3 utils/test_postgis_exampletest.py
37 tests passed

make -C regress visual-examples-run
manual example tests passed: 499
visual examples rendered: 504

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.

make -C doc pdf
'postgis-3.7.0dev-en.pdf' successfully built
```

## Update: Minkowski sum figures

The four detached `CG_MinkowskiSum` before/after rasters are replaced by two checked generated comparisons. The examples retain the original LineString/circle and MultiPoint/polygon inputs, fix the malformed mixed SQL/WKT blocks, and return the complete input and result layers from one query. Geometry columns named `input_*` provide a generic authoring convention for blue input layers without repeating automatically extracted SQL literals; the displayed legend removes that prefix.

Local validation on head `adf9eb607`:

```text
python3 utils/test_postgis_exampletest.py
38 tests passed

make -C regress visual-examples-run
manual example tests passed: 501
visual examples rendered: 506

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: medial-axis and skeleton figures

The `CG_ApproximateMedialAxis` raster and the two detached `CG_StraightSkeletonPartition` input/result rasters are replaced by two checked generated comparisons. Both examples retain the original stepped Polygon; one overlays its medial axis and the other renders every partition face from the current PolyhedralSurface output.

Local validation on head `11b49d058`:

```text
make -C regress visual-examples-run
manual example tests passed: 503
visual examples rendered: 508

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: polygon tessellation figure

The `CG_Tesselate` Polygon before/after row is replaced by one checked generated comparison using the same authored Polygon. This also removes a clean-build-only defect: the referenced input `st_tesselate03.png` had no tracked source or image-build rule. Exact execution found that the stored output fixture used an older valid diagonal choice, so the documented TIN now matches the current function result.

Local validation on head `2a99b36f0`:

```text
make -C regress visual-examples-run
manual example tests passed: 504
visual examples rendered: 509

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: visibility figures

The three `CG_Visibility` rasters are replaced by two checked generated comparisons using the same polygon-with-holes and the original point/segment observers. The renderer now paints area fills before linework and points, so an observer remains visible and hoverable even where the visibility polygon covers the same coordinates.

Local validation on head `f0056c9d9`:

```text
python3 utils/test_postgis_exampletest.py
39 tests passed

make -C regress visual-examples-run
manual example tests passed: 506
visual examples rendered: 511

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: convex partition figures

Five convex-partition rasters and four orphaned deprecated-alias fixtures are removed. `CG_ApproxConvexPartition`, `CG_GreeneApproxConvexPartition`, `CG_OptimalConvexPartition`, and `CG_YMonotonePartition` now render the same authored Polygon and their complete documented partition geometries through the common build-time SVG path. Their version-sensitive component ordering remains documented-only rather than exact-output executed.

Local validation on head `d8016952f`:

```text
make -C regress visual-examples-run
manual example tests passed: 506
visual examples rendered: 515

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.

make -C doc pdf
'postgis-3.7.0dev-en.pdf' successfully built
```

## Update: CG simplify figures

Three `CG_Simplify` rasters are replaced by one checked comparison of the original GeometryCollection, topology-preserving simplification, and non-topology-preserving simplification. The old SQL claimed a threshold of `0.5`, but that value does not simplify the documented geometry and did not match either stored fixture. A threshold of `50` reproduces both authored result geometries and makes the example internally consistent.

Local validation on head `eda2478f1`:

```text
make -C regress visual-examples-run
manual example tests passed: 507
visual examples rendered: 516

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```

## Update: cluster figures

The `ST_ClusterRelateWin` and `ST_ClusterKMeans` rasters are replaced by generated figures whose geometry columns correspond to the actual cluster ids. The relate example is now a self-contained checked query and keeps the member-id lists in text output. K-means remains documented-current because cluster numbering and subdivision details vary across supported library versions. Documented-only psql headers now flow into the renderer, so these figures retain `cluster_0`, `cluster_1`, and `cluster_2` instead of generic numbered output labels.

Local validation on head `59f05e901`:

```text
python3 utils/test_postgis_exampletest.py
39 tests passed

make -C regress visual-examples-run
manual example tests passed: 508
visual examples rendered: 518

make -C doc chunked-html
make -C doc check
postgis-out.xml validates
DocBook block label checker summary: no generated HTML block label issues found.
```


Komzpa:
Un-spaghettied per review:

- **XML**: removed the output-narration chrome stranded by the splits (`-- you'll get an output something like --` and friends), and a sweep found **12 more hidden mixed blocks** — bare scalar/WKT result lines inside code blocks with no psql chrome for the checker to see (ST_NPoints, ST_NumGeometries, ST_SRID, ST_Square, DropRasterConstraints, ST_EstimatedExtent, …). All split properly with genuine psql output; the two ST_EstimatedExtent examples are marked `requires-external-state` (they query tables the test db doesn't have).
- **Python**: SQL keyword/starter vocabulary now lives once in `utils/docbook_qa.py`; the lint regexes are derived from it and `ref-index` embeds it into `postgis-ref-index.json` (functions/operators output unchanged).
- **JS**: `copy-code.js` → `example-blocks.js` — honest name, no more hand-maintained keyword table (reads the ref index it already loads; highlighting degrades to off on python-less builds, copy button unaffected), internally structured into data-loading / tokenizer / linker / copy-button sections.

Validation: strict `make -C doc check` gates, unit tests, DB exampletest (only the known `--without-protobuf` local failures remain), and the postgis-build-env docker run: **manual example tests passed: 659** (637 before this PR started testing the newly split examples).

Komzpa:
Follow-up per review: the mixed-block heuristic now treats **any run of 3+ dashes after the first SQL statement** as an output marker (dash runs before the first statement — comment banners — are ignored). That caught the ST_LineToCurve side-by-side example and fake `---st_astext--` chrome on 8 more pages (ST_ConvexHull, ST_ConcaveHull, ST_Polygonize, ST_PointOnSurface, ST_MemSize, ST_GeogFromText, ST_Value, ST_BandMetaData) — all split with genuine psql output. Bonus: the newly runnable ST_MemSize example exposed a stale documented value (73 bytes from the GSERIALIZED v1 era → 80 now).

Validation: with protobuf enabled locally the manual example tests pass **669/669** with zero exclusions beyond the documented `requires-external-state` opt-outs; the postgis-build-env docker run also reports 669 passed.

Komzpa:
More review follow-ups:

- **Dash-run heuristic simplified**: 3+ consecutive dashes anywhere inside a `programlisting` is now an output marker — no after-first-statement scoping.
- **New `redundant-output-caption` lint rule**: a bare `Output:`/`Result:` paragraph right before a screen block duplicates the rendered Output label; the two instances (ST_GeometryN) are removed.
- **file:// highlighting fixed**: Chrome blocks `fetch()` on `file://`, which silently killed SQL highlighting for locally browsed manuals once keywords moved into the JSON index. The ref index is now also emitted as `postgis-ref-index.js` (`window.POSTGIS_REF_INDEX = …`) loaded via a classic script tag before `example-blocks.js`, with fetch as fallback.
- **Doc bug uncovered by the caption removal**: the ST_GeometryN "Polyhedral Surfaces, TIN" examples promised decomposition into faces, but PostGIS intentionally treats these as single geometries (`lwgeom_ogc.c`: 'TIN and PolyhedralSurface count as 1 geometry') — `ST_GeometryN` returns NULL there and `ST_PatchN`/`ST_NumPatches` are the face accessors. The bare `Output:` caption had kept these examples out of exampletest coverage all along; they now document the real NULL behavior with an xref to ST_PatchN.

Validation: local (protobuf-enabled) and postgis-build-env docker runs both pass **671/671** manual example tests.

Komzpa:
Next batch of review follow-ups:

- **Info-severity lint tier**: new rules land as report-only until their backlog is cleaned, without weakening the strict warning gate. Four rules shipped: `tab-in-verbatim`, `stacked-admonitions` (116), `ancient-version-note` (36), `example-missing-output` (287).
- **Whitespace normalization across 28 files**: tabs expanded (psql tables aligned to their separator line) and XML author indentation stripped from verbatim blocks — fixes the misaligned tables (ST_NDims, ST_Reverse) and the deep-indented code blocks (ST_RemoveIrrelevantPointsForView). Whitespace-only by construction, verified by a per-refentry content-integrity check.
- **GT-252 metadata**: ST_GeometryN/ST_NumGeometries now carry `Changed: 3.6.0` notes for the unitary TIN/PolyhedralSurface semantics; ST_GeometryN's 1-based-index note folded into prose (0.8.0 archaeology dropped).
- **Example UX (all copy-safe pseudo-elements — selection/copy stays byte-identical)**: bracket-pair highlight on hover; line-number gutter with wrap markers on lines that actually wrap.
- `<=` now tokenizes as one operator (longest-first matching), and the stray blank first line in highlighted blocks is gone.

Validation: 671/671 manual example tests locally (protobuf build) and in postgis-build-env docker. Inline WKT figure rendering (JTS-TestBuilder-style grids) is in progress as the next batch.

Komzpa:
The build-time figure work is now in a reviewable state on [`1b64f19a2`](https://github.com/postgis/postgis/commit/1b64f19a2c604c142528e61d7b92c575a29af2a0):

- Documented geometry examples now produce build-time SVG figures from the same queries and outputs used by exampletest. HTML keeps interactive text/figure controls, while PDF consumes the generated figures directly.
- The renderer covers geometry collections, curves/NURBS, raster image output, named multi-column context, comparable/shared extents, direction markers, grids, colors, and hoverable components. Static PNG illustrations were migrated where the documented query can express the same geometry.
- Example prose is attached to its Code/Output/Figure block without putting the block inside list bullets, and shell/WKT/EWKT examples now use the same labeled presentation and syntax/linking machinery.
- Geography-input-only queries no longer get an automatic planar figure. This removes the misleading straight-line illustration from the Seattle-London great-circle example while preserving planar geometry-input figures and figures derived from actual geometry/WKT output.
- Remaining prose-style SQL comments in the geography introduction were moved out of Copy blocks, including the NAD83/NAD27 and 1000 km examples.

Browser verification covered the synchronized old/new review page and the affected manual pages. In particular, the geography great-circle result has no planar figure, its geometry comparison retains one, and the introductory captions render as prose while Copy contains SQL only.

Validation on this head:

- `python3 utils/test_postgis_exampletest.py`
- `make -C doc/html/images visual-examples-check` (548 generated examples, 0 failures)
- `make -C doc check`

Komzpa:
@coderabbitai review

Komzpa:
CI follow-up on [`ca768c499`](https://github.com/postgis/postgis/commit/ca768c49977853a98e02e80cf97b9a540905b1be): the mingw job failed twice on the existing `mic-box` regression because `ST_MaximumInscribedCircle` selected a different equally-near polygon edge (`POINT(0 49.5)` instead of `POINT(50.5 99)`). Both points satisfy the function contract; the regression was asserting an unspecified tie-break.

The test now checks the stable semantics instead: `nearest` must lie on the input boundary and its distance from `center` must equal `radius`. The center and radius expectations remain unchanged.

Focused local validation passes both normal and upgrade paths:

- `make -C regress/core check-regress TESTS=../../regress/core/geos39`

Komzpa:
@coderabbitai review

PR #1168, opened 2026-07-21
mvt: Fix potential NULL dereference

Gonna-Give-You-Up:
Add ctx check against NULL before using.

### DIFF
--- a/postgis/mvt.c
+++ b/postgis/mvt.c
@@ -1100,17 +1100,19 @@ static bytea *mvt_ctx_to_bytea(mvt_agg_context *ctx)
 	size_t len;
 	bytea *ba;
 
+	if (!ctx)
+    {
+        goto error;
+    }
 	if (!ctx->tile)
 	{
 		ctx->tile = mvt_ctx_to_tile(ctx);
 	}
 
 	/* Zero features => empty bytea output */
-	if (ctx && ctx->layer && ctx->layer->n_features == 0)
+	if (ctx->layer && ctx->layer->n_features == 0)
 	{
-		bytea* ba_empty = palloc(VARHDRSZ);
-		SET_VARSIZE(ba_empty, VARHDRSZ);
-		return ba_empty;
+		goto error;
 	}
 
 	/* Serialize the Tile */
@@ -1119,6 +1121,11 @@ static bytea *mvt_ctx_to_bytea(mvt_agg_context *ctx)
 	vector_tile__tile__pack(ctx->tile, (uint8_t*)VARDATA(ba));
 	SET_VARSIZE(ba, len);
 	return ba;
+
+	error:
+	    bytea* ba_empty = palloc(VARHDRSZ);
+		SET_VARSIZE(ba_empty, VARHDRSZ);
+		return ba_empty;
 }
 
 


postgis/mvt.c
```diff
@@ -1119,6 +1121,11 @@ static bytea *mvt_ctx_to_bytea(mvt_agg_context *ctx)
 	vector_tile__tile__pack(ctx->tile, (uint8_t*)VARDATA(ba));
 	SET_VARSIZE(ba, len);
 	return ba;
+
+	error:
+	    bytea* ba_empty = palloc(VARHDRSZ);
```
ProjectMutilation, line 1126: remove declaration ba_empty and use ba
```
/* return empty */
ba = palloc(VARHDRSZ);
SET_VARSIZE(ba, VARHDRSZ);
return ba;
```

ProjectMutilation (changes requested):

Komzpa:
Hello @Gonna-Give-You-Up - thank you for the contributions. Would you be willing to be credited by a real name and/or a company in the NEWS file, or just the github handle?

Komzpa:
Thanks for the patch. I checked the failing GitHub Actions logs; this is not inherited from `master` (the exact base SHA `1f5885aa8bae29688509a3ca19e17c0cb6351449` had the matching GitHub checks green).

There are two actionable blockers here.

First, the C build failure is from the new `error:` label being followed directly by a declaration:

```text
FreeBSD 14.3:
mvt.c:1126:6: error: label followed by a declaration is a C23 extension [-Werror,-Wc23-extensions]
 1126 |             bytea* ba_empty = palloc(VARHDRSZ);

CI (pg15-clang-geos311-gdal35-proj90, usan_clang):
mvt.c:1126:6: error: expected expression
            bytea* ba_empty = palloc(VARHDRSZ);
```

The simplest fix is to avoid the `goto`/label here and return the empty `bytea` directly before dereferencing `ctx`:

```c
if (!ctx || (ctx->layer && ctx->layer->n_features == 0))
{
	ba = palloc(VARHDRSZ);
	SET_VARSIZE(ba, VARHDRSZ);
	return ba;
}

if (!ctx->tile)
{
	ctx->tile = mvt_ctx_to_tile(ctx);
}
```

That also addresses the earlier review comment: no separate `ba_empty` declaration is needed; the existing `ba` variable is enough.

Second, the contributor-credits job is failing independently:

```text
FAIL: doc/credits.xml is missing 1 contributor(s):
  - Bogdanov Svyatoslav (Git author b992ced13b95)
Add each canonical display name to the manual credits, or normalize an alias in .mailmap.
```

So CI needs a contributor-credit update as well. Since your two commits currently use different author display names (`Bogdanov Svyatoslav` here, `Gonna-Give-You-Up` in the other PR), please pick the display name you want credited under. If `Bogdanov Svyatoslav` is the intended canonical name, add that to the manual credits and normalize the handle/name variant in `.mailmap`; otherwise add the preferred name/handle consistently.

The NULL check itself is worth keeping: the current `mvt_ctx_to_bytea()` dereferences `ctx` before the later `ctx && ...` check, so the guard needs to be placed before `ctx->tile`.

Komzpa:
Thanks for the MVT patch. I opened a combined pull request at https://gitea.osgeo.org/postgis/postgis/pulls/560 that carries this change with Svyatoslav Bogdanov credited as author.

The combined branch fixes the C label/declaration syntax error reported here as a separate maintainer commit and adds a zero-row ST_AsMVT regression. I am leaving this pull request open for the maintainer decision; the combined pull request supersedes it for review.

Komzpa:
merged, thanks

PR #1170, opened 2026-08-09
topo: add allocation handler

ProjectMutilation:
default_allocator can silently return NULL
Signed-off-by: Maksim Korotkov <m.korotkov@postgrespro.ru>


Komzpa:
merged. thank you!

strk:
Thank you 👍

PR #1172, opened 2026-08-22
Guard against NULL in lwgeom_wrapx point clone

ProjectMutilation:
lwgeom_clone_deep can return NULL when the underlying allocator (default_allocator -> malloc) fails, which propagates through lwgeom_as_lwpoint and is dereferenced at pt->point. Add a NULL check before the dereference.
Found by PostgresPro


Komzpa:
Thank you! Merged.

PR #1173, opened 2026-08-22
Fix NULL deref in raster colormap arg cleanup

ProjectMutilation:
__rti_colormap_arg_destroy dereferenced arg->expr[i] without first checking that the array itself was allocated. When rtalloc fails (arg->expr == NULL) the destroy path reads from a NULL array and crashes. Guard the loop on arg->expr != NULL and stop at the first NULL element, which also avoids freeing uninitialized entries when a partial allocation (rtalloc of individual expressions) fails.

Fixes: 46a76d3b39bb36c2b896e2a67dc88dac13be1fce
Found by PostgresPro.


Komzpa:
Thank you! Merged.

PR #1174, opened 2026-08-22
Prevent NULL deref when shrinking raster file list fails

ProjectMutilation:
When rtrealloc fails on the in-memory raster file list, config->rt_file becomes NULL while config->rt_file_count is left non-zero. The subsequent rtdealloc_config() then iterates count..0 dereferencing config->rt_file[i] on a NULL array, crashing in the OOM cleanup path.

Reset rt_file_count to 0 before running the error cleanup so the partial state is not dereferenced. Applies to both the append path (grow) and the schema.table-parse path (shrink).

Fixes: 8b8ce38f5dd2b30f8d8bac3a09bc60213688f48d


Komzpa:
Thank you! Merged.

PR #1178, opened 2026-09-07
loader: check column map allocations and fix cleanup

ProjectMutilation:
Check the return of malloc/calloc in colmap_read and bail out with an error instead of dereferencing NULL on the pgfieldnames/dbffieldnames array and per-field string allocations. The column map arrays are now calloc'd so unused slots are NULL-initialized.

Rework colmap_clean to be safe on partial state: early-return on NULL, free the field-name arrays unconditionally (free(NULL) is valid), and reset the map via colmap_init so the map is reusable and a subsequent ShpLoaderDestroy colmap_clean call stays safe. The calloc-failure path resets map->size first so the cleanup loop doesn't index a half-allocated array.

### DIFF
--- a/loader/shpcommon.c
+++ b/loader/shpcommon.c
@@ -176,18 +176,21 @@ void
 colmap_clean(colmap *map)
 {
 	int i;
-	if (map != NULL){
-		if (map->size)
+	if (map == NULL)
+	{
+		return;
+	}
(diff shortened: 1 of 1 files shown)


ProjectMutilation:
@robe2 @Komzpa hi, could you please review the changes?

